PR #29, opened 2023-05-24
Fixed a bug in utils.py that could cause token overflow when calling OpenAI API

ChenhuiHu:
Fixed bugs. If the length of a line in lines is greater than max_chunk_size, the line will not be segmented further, which leads to a mismatch between the function and its function, which may cause token overflow when calling the OpenAI API.

### DIFF
--- a/nemoguardrails/kb/utils.py
+++ b/nemoguardrails/kb/utils.py
@@ -17,6 +17,8 @@ from typing import List
 
 import yaml
 
+import re
+
 
 def split_markdown_in_topic_chunks(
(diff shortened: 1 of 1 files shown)


ChenhuiHu:
Thanks for you reply! @drazvan . I fixed a problem in the previous method that could have caused semantic segmentation. In this code, it is guaranteed that a single sentence is the smallest indivisible unit, and the size of each chunk will not exceed max_chunk_size.

drazvan:
Closing as old.

PR #40, opened 2023-06-05
Hongyi llama index rails demo

hongyishi:
This is an example demo of LlamaIndex integration with NeMo. Here I create a callable bot action that calls the query engine on the example doc, and add a basic flow using it.


examples/demo_llama_index_guardrails.py
```diff
@@ -0,0 +1,76 @@
...
+
+define user express ill intent
+  "I hate you"
+  "I want to destroy the world"
+
+define bot express cannot respond
+  "I'm sorry I cannot help you with that."
+
+define user express question
+   "What is the current unemployment rate?"
```
drazvan, line 18: This canonical form would need more examples, so that the LLM can generalise that this is about any question. For example, I've changed the question to "Tell me the number of jobs last month" and the generated canonical form was `ask specific question`. Alternatively, the approach from [demo_chain_with_guardrails.py](https://github.com/NVIDIA/NeMo-Guardrails/blob/main/examples/demo_chain_with_guardrails.py) can be taken, and use `user ...` in the flow, instead of an explicit canonical form for a question (see my comment below).

examples/demo_llama_index_guardrails.py
```diff
@@ -0,0 +1,76 @@
...
+
+# Basic guardrail example
+define flow
+  user express ill intent
+  bot express cannot respond
+
+# Question answering flow
+define flow
+  user express question
```
drazvan: For the purpose of this example, `user ...` would work as well.

examples/demo_llama_index_guardrails.py
```diff
@@ -0,0 +1,76 @@
...
+    ).load_data()
+    llm_predictor = LLMPredictor(llm=llm)
+    index = GPTVectorStoreIndex.from_documents(docs, llm_predictor=llm_predictor)
+    default_query_engine = index.as_query_engine()
+    return default_query_engine
+
+
+def _get_callable_query_engine(
+    query_engine: BaseQueryEngine
+) -> Callable[[str], str]:
```
drazvan: The correct return type is: `Callable[[str], Coroutine[Any, Any, str]]`.

requirements.txt
```diff
@@ -13,3 +13,4 @@ uvicorn==0.21.1
 httpx==0.23.3
 simpleeval==0.9.13
 typing-extensions==4.5.0
+llama_index==0.6.14
```
drazvan: Only core dependencies should be added to `requirements.txt`.

examples/demo_llama_index_guardrails.py
hongyishi: Changed to use ...

examples/demo_llama_index_guardrails.py
hongyishi: Sure let's use that

examples/demo_llama_index_guardrails.py
hongyishi: Fixed

requirements.txt
hongyishi: Fixed

hongyishi:
Hi @drazvan hopefully I signed the commit correctly!

PR #70, opened 2023-07-12
Enhancement: Extend hallucination llm engine. Resolve #47

### LINKED ISSUE #47: Suggestion: Extend llm type in hallucination.py
```python
):

    """

    bot_response = context.get("last_bot_message")

        if type(llm) != OpenAI:
            log.warning(
            )
            return False
```

In Langchain, they split models into two classes, for the instruction Openai models the hierarchy is `BaseLLM->BaseOpenAI->OpenAI`, for chat models like gpt-3.5 they have `BaseLanguageModel->BaseChatModel->ChatOpenAI`.

The above hallucination code limits the llm field only to the instruction model routine, I am not sure if this is designed on purpose. If not, we could extend the llm to both ways like `llm: Optional[Union[BaseLLM, BaseLanguageModel]] = None,`

On another side, `if type(llm) != OpenAI:` doesn't support sub class that inherits from OpenAI, I'd recommend `if isinstance(llm, OpenAI):` or `if issubclass(type(llm), OpenAI)`. And of course, you may add `ChatOpenAI` here.

MeowLake:
**Background:**

The original hallucination check function only supports OpenAI LLM engine (instruction models), this PR is to extend the scope of supporting engines.

**Changes:**

1. Now it supports three types of llm from langchain: OpenAI (instruction models), ChatOpenAI (chat models), and customized class inherit from one of them.
2. Update some comments and warning messages.
3. Add a new template construction for ChatOpenAI because roles must be associated with messages.
4. Minor changes to resolve the API differences between OpenAI and ChatOpenAI.

**Note:**

Unfortunately, I didn't get a chance to verify the effect of changes as I meet some issues using OpenAI these days. It would be great if you could help make some tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ scratch.py
 firebase.json
 build
 dist
+# Ignoring this for now
+scripts
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#58](https://github.com/NVIDIA/NeMo-Guardrails/issues/58): Fix install on Mac OS 13.
 - [#55](https://github.com/NVIDIA/NeMo-Guardrails/issues/55): Fix bug in example causing config.py to crash on computers with no CUDA-enabled GPUs.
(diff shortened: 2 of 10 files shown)


MeowLake:
Sorry this PR is getting messy, so I closed it.

PR #71, opened 2023-07-12
Extend hallucination llm engines. Resolve #47

### LINKED ISSUE #47: Suggestion: Extend llm type in hallucination.py

MeowLake:
**Background:**

The original hallucination check function only supports OpenAI LLM engine (instruction models), this PR is to extend the scope of supporting engines.

**Changes:**

1. Now it supports three types of llm from langchain: OpenAI (instruction models), ChatOpenAI (chat models), and customized class inherit from one of them.
2. Update some comments and warning messages.
3. Add a new template construction for ChatOpenAI because roles must be associated with messages.
4. Minor changes to resolve the API differences between OpenAI and ChatOpenAI.

**Note:**

Unfortunately, I didn't get a chance to verify the effect of changes as I meet some issues using OpenAI these days. It would be great if you could help make some tests.

### DIFF
--- a/nemoguardrails/actions/hallucination/hallucination.py
+++ b/nemoguardrails/actions/hallucination/hallucination.py
@@ -42,7 +42,7 @@ HALLUCINATION_NUM_EXTRA_RESPONSES = 2
 async def check_hallucination(
     llm_task_manager: LLMTaskManager,
     context: Optional[dict] = None,
-    llm: Optional[Union[BaseLLM, BaseLanguageModel]] = None,
+    llm: Optional[BaseLanguageModel] = None,
     use_llm_checking: bool = True,
 ):
     """Checks if the last bot response is a hallucination by checking multiple completions for self-consistency.
@@ -64,47 +64,14 @@ async def check_hallucination(
             )
             return False
 
-        # Customized sub engine can only be either OpenAI LLM or OpenAI chat model
-        if isinstance(llm, OpenAI) and isinstance(llm, ChatOpenAI):
-            log.warning(
-                f"LLM engine must be either OpenAI LLM or chat model."
-                f"Current LLM engine is a merge of two types."
-            )
-            return False
-
-        # Use the "generate" call from langchain to get all completions in the same response.
-        last_bot_prompt = PromptTemplate(template="{text}", input_variables=["text"])
-
-        # Create message prompt for chat model
-        if isinstance(llm, ChatOpenAI):
-            human_message_prompt = HumanMessagePromptTemplate(prompt=last_bot_prompt)
-            last_bot_prompt = ChatPromptTemplate.from_messages([human_message_prompt])
-
-        chain = LLMChain(prompt=last_bot_prompt, llm=llm)
-
-        # Generate multiple responses with temperature 1.
-        if isinstance(llm, OpenAI):
-            with llm_params(llm, temperature=1, n=num_responses, best_of=num_responses):
-                extra_llm_response = await chain.agenerate(
-                    [{"text": last_bot_prompt_string}],
-                    run_manager=logging_callback_manager_for_chain,
-                )
-        else:
-            # Chat model has no `best_of` argument
-            with llm_params(llm, temperature=1, n=num_responses):
-                extra_llm_response = await chain.agenerate(
-                    [{"text": last_bot_prompt_string}],
-                    run_manager=logging_callback_manager_for_chain,
-                )
-
-        extra_llm_completions = []
-        if len(extra_llm_response.generations) > 0:
-            extra_llm_completions = extra_llm_response.generations[0]
+        # Generate all completions with given LLM parameters
+        with llm_params(llm, temperature=1, n=num_responses):
+            extra_llm_completions = await llm_call(llm, last_bot_prompt_string)
 
         extra_responses = []
         i = 0
         while i < num_responses and i < len(extra_llm_completions):
-            result = extra_llm_completions[i].text
+            result = extra_llm_completions[i]
             # We need the same post-processing of responses as in "generate_bot_message"
             result = get_multiline_response(result)
             result = strip_quotes(result)
--- a/nemoguardrails/actions/hallucination/hallucination.py
+++ b/nemoguardrails/actions/hallucination/hallucination.py
@@ -14,11 +14,14 @@
 # limitations under the License.
 
 import logging
-from typing import Optional
+from typing import Optional, Union
 
 from langchain import LLMChain, PromptTemplate
+from langchain.base_language import BaseLanguageModel
+from langchain.chat_models import ChatOpenAI
 from langchain.llms import OpenAI
 from langchain.llms.base import BaseLLM
+from langchain.prompts import ChatPromptTemplate, HumanMessagePromptTemplate
 
 from nemoguardrails.actions.llm.utils import (
     get_multiline_response,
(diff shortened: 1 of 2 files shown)


nemoguardrails/actions/hallucination/hallucination.py
```diff
@@ -39,7 +42,7 @@
 async def check_hallucination(
     llm_task_manager: LLMTaskManager,
     context: Optional[dict] = None,
-    llm: Optional[BaseLLM] = None,
+    llm: Optional[Union[BaseLLM, BaseLanguageModel]] = None,
```
trebedea: `BaseLLM` is a sub-class of `BaseLanguageModel`, let's use
`llm: Optional[BaseLanguageModel] = None`

nemoguardrails/actions/hallucination/hallucination.py
```diff
@@ -53,24 +56,46 @@ async def check_hallucination(
...
+        if not isinstance(llm, (OpenAI, ChatOpenAI)):
             log.warning(
-                f"Hallucination rail can only be used with OpenAI LLM engines."
+                f"Hallucination rail can only be used with OpenAI LLM and chat model engines, and their customized sub engines"
                 f"Current LLM engine is {type(llm).__name__}."
             )
             return False
 
+        # Customized sub engine can only be either OpenAI LLM or OpenAI chat model
+        if isinstance(llm, OpenAI) and isinstance(llm, ChatOpenAI):
```
trebedea: This is already addressed at L60, plus this condition can never be True (llm be both OpenAI and ChatOpenAI). So you can remove the entire block.

nemoguardrails/actions/hallucination/hallucination.py
```diff
@@ -53,24 +56,46 @@ async def check_hallucination(
...
+                    [{"text": last_bot_prompt_string}],
+                    run_manager=logging_callback_manager_for_chain,
+                )
+        else:
+            # Chat model has no `best_of` argument
+            with llm_params(llm, temperature=1, n=num_responses):
+                extra_llm_response = await chain.agenerate(
+                    [{"text": last_bot_prompt_string}],
+                    run_manager=logging_callback_manager_for_chain,
+                )
```
drazvan: I think here you can use directly the value from `context.get("_last_bot_prompt")` and pass it to `llm_call` (https://github.com/NVIDIA/NeMo-Guardrails/blob/main/nemoguardrails/actions/llm/utils.py#L27) which will deal correctly with chat vs. completion.

trebedea (changes requested):
Small suggested changes, I'm also investigating @drazvan's advice. Will get back with a response.

nemoguardrails/actions/hallucination/hallucination.py
trebedea: @MeowLake , I can take the changes using `llm_call` if you want. It would be great if you could address the two points I have signaled beforehand. 

Of course, if you want to make the changes using `llm_call` then it's perfect. 
It's a good PR, we should merge it after solving these 3 points.

MeowLake: @trebedea It looks like L86-98 generated a batch of results because of the specified `num_responses`, they were popped out one by one in L105-112. `llm_call` seems to be just for a single result, so it has to be called `num_responses` times in a loop. Is this correct? If yes, I can handle it.

nemoguardrails/actions/hallucination/hallucination.py
MeowLake: L60 is to exclude class that is not in [OpenAI, ChatOpenAI, sub-class of OpenAI, sub-class of ChatOpenAI, sub-class of OpenAI and ChatOpenAI], this block is to exclude sub-class of `OpenAI and ChatOpenAI`, which is a very edge case. I can remove it if that is not necessary.

nemoguardrails/actions/hallucination/hallucination.py
MeowLake: Sounds good!

nemoguardrails/actions/hallucination/hallucination.py
trebedea: Yes, you can remove it. Wouldn't any sub-class of both `OpenAI` and `ChatOpenAI` be caught by L60?

nemoguardrails/actions/hallucination/hallucination.py
trebedea: @MeowLake 

The cleanest way to do this is to make `llm_call` return all generations / responses when called with `n>1` (for example, for `OpenAI `and `ChatOpenAI`). I've checked and we have access to all of them, but only return the first one in the current implementation. 

It would be only a couple of lines of changes to make this possible - return a string if only one completion in `result.generations[0]` , otherwise return all completions in a list:
`async def llm_call(llm: BaseLanguageModel, prompt: Union[str, List[dict]]) -> Union[str, List[str]]:`

Do you want to take this?

MeowLake: I submitted a new commit to resolve the three changes.

nemoguardrails/actions/hallucination/hallucination.py
```diff
@@ -14,11 +14,14 @@
 # limitations under the License.
 
 import logging
-from typing import Optional
+from typing import Optional, Union
```
trebedea:
```suggestion
from typing import Optional
```

nemoguardrails/actions/hallucination/hallucination.py
```diff
@@ -14,11 +14,14 @@
...
 
 from langchain import LLMChain, PromptTemplate
+from langchain.base_language import BaseLanguageModel
+from langchain.chat_models import ChatOpenAI
 from langchain.llms import OpenAI
 from langchain.llms.base import BaseLLM
+from langchain.prompts import ChatPromptTemplate, HumanMessagePromptTemplate
```
trebedea: Please remove all unused imports, e.g.

> from langchain import LLMChain, PromptTemplate
> from langchain.llms.base import BaseLLM
> from langchain.prompts import ChatPromptTemplate, HumanMessagePromptTemplate

nemoguardrails/actions/llm/utils.py
```diff
@@ -50,7 +51,17 @@ async def llm_call(llm: BaseLanguageModel, prompt: Union[str, List[dict]]) -> st
...
+        # Return string if there is only one completion,
+        # otherwise, return all completions in a list
+        if len(result.generations[0]) > 1:
+            all_completions = []
+            i = 0
+            while i < len(result.generations[0]):
+                all_completions.append(result.generations[0][i].text)
+                i += 1
+            return all_completions
+        else:
```
trebedea: Here you have an extra tab and the function now returns `None` when the `isinstance(prompt, str)`.

drazvan:
Closing as old.

PR #76, opened 2023-07-17
Fact check and hallucination check bug fix

Siraj-R-Khan:
Noticed a bug where fact checker was returning false even though the bot response exactly matched the data in the knowledge base. It seemed that special characters in the bot response were causing the LLM to get confused resulting in entails in fact_checking.py contianing no. In order to fix this, I used square brackets in prompts in general.yml to better separate the text from the bot response, the text from the knowledge base and the text of the rest of the prompt. I noticed that the check_hallucination prompt was susceptible to the same issue, so applied the same change.

### DIFF
--- a/nemoguardrails/llm/prompts/general.yml
+++ b/nemoguardrails/llm/prompts/general.yml
@@ -91,7 +91,7 @@ prompts:
     content: |-
       You are given a task to identify if the hypothesis is grounded and entailed to the evidence.
       You will only use the contents of the evidence and not rely on external knowledge.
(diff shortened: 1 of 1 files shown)


Siraj-R-Khan:
Can't seem to get git/github to sign commits, so please feel free to just copy and paste changes.

drazvan:
Thanks for this @Siraj-R-Khan ! Regarding the signing, it seems you just need to upload your public GPG key to Github and you're done (the commit is signed, it's just that the signature is not verified).

Siraj-R-Khan:
@drazvan 

Thank you for the tip. I've sorted the signing issue. Will create a new fork and submit a new pull request.

PR #82, opened 2023-07-21
Fact check and hallucination check bug fix

Siraj-Aizlewood:
Noticed a bug where fact checker was returning false even though the bot response exactly matched the data in the knowledge base. It seemed that special characters in the bot response were causing the LLM to get confused resulting in entails in fact_checking.py contianing no. In order to fix this, I used square brackets in prompts in general.yml to better separate the text from the bot response, the text from the knowledge base and the text of the rest of the prompt. I noticed that the check_hallucination prompt was susceptible to the same issue, so applied the same change.


Siraj-R-Khan:
@drazvan 

Finally managed to submit a pull request with correct commit signing.

drazvan:
Closing as outdated. Thanks again @Siraj-Aizlewood!

Issue #99, opened 2023-08-16
Not hardcoded embedding models


jamescalam:
Hi, I'd like to propose a fix to allow us to set the embedding model used to create the vector representations of rails. My reason for this is primarily that I'd like to be able to use a service like OpenAI for ada-002 primarily as an easy fix to large container sizes when deploying anything containing guardrails code — and additionally to allow us to select different hugging face models if preferred.

I'd propose in an initial version of this to support OpenAI and Hugging Face models. But very open to suggestions — I'd also be happy to work on this. Would there be any preference on how to set an embedding model? I figure we could do something in the config.yaml like:

```yaml
models:
  - type: main
    engine: openai
    model: text-davinci-003
 - type: embedding
   engine: openai
   model: text-embedding-ada-002
```

OR with hugging face:

```yaml
models:
  - type: main
    engine: openai
    model: text-davinci-003
 - type: embedding
   engine: huggingface
   model: sentence-transformers/all-MiniLM-L6-v2
```

drazvan:
Hi @jamescalam! Your suggestion works. 

There is actually a minimal support for changing the embedding model, but still to a SentenceTransformer one. 
https://github.com/NVIDIA/NeMo-Guardrails/blob/main/nemoguardrails/actions/llm/generation.py#L73

See https://github.com/NVIDIA/NeMo-Guardrails/discussions/97 as well. 

```yaml
models:
   ...
   - type: embedding
     engine: SentenceTransformer
     model: all-MiniLM-L6-v2
```

We have a work in progress on a private branch to enable the integration of any embedding and search provider. There are a few things that are changing, including the interface for `EmbeddingsIndex` which needs to be async. I reckon next week we should be able to push this to the repo. And if you can, you can help with adding support for OpenAI as an `EmbeddingSearchProvider`. 

Thanks a lot!

jamescalam:
@drazvan created PR here #101

dom-vaz:
Hi there

I think this approach will work in a lot of cases. But if you have a more complicated enterprise environment it may not be the best solution. 

We can already pass a langchain LLM as main Model to LLMRails. Why not add a second variable to pass a langchain embedding model? This would open up a lot of new options without having to implement all of them. OpenAI Support is already good. But in an enterprise setup it will anyways most likely be AzureOpenAI...

Kind Regards
Dominik

jamescalam:
Issue was solved with #101

krannnn:
[...]

PR #101, opened 2023-08-22
added alternative embedding models for sentence transformers and openai

jamescalam:
Addressing #99. Added ability to choose embedding model via the config.yaml. Also added OpenAI API as option (alongside Sentence Transformers). We can do:

```yaml
models:
  - type: main
    engine: openai
    model: text-davinci-003
  - type: embedding
    engine: openai
    model: text-embedding-ada-002
```

and

```yaml
models:
  - type: main
    engine: openai
    model: text-davinci-003
  - type: embedding
    engine: SentenceTransformers
    model: all-MiniLM-L6-v2
```

I understand there is also some work ongoing in this space, but in a private repo, naturally, I can't see this, so I apologize if any of this conflicts with or overlaps. Let me know if any of this can be of use!

Thanks

### DIFF
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -44,6 +44,7 @@ from nemoguardrails.kb.index import IndexItem
 from nemoguardrails.kb.kb import KnowledgeBase
 from nemoguardrails.language.parser import parse_colang_file
 from nemoguardrails.llm.params import llm_params
+from nemoguardrails.llm.providers import get_embedding_provider_names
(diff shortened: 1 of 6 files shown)


drazvan:
Thanks for this @jamescalam! I'd like to merge this in the next few days. Let me check against the internal work, but I think we'll first merge this and then apply the rest of the changes. 

A couple points:
1. After merge, we'll likely change the interface to be async. Otherwise, in a server scenario whenever an embedding is computed using an endpoint, like with OpenAI, everything is blocked. 
2. We should probably add a `register_embedding_provider` method as well so that a user can register a custom one from their config. 

Nice work! 👍

PR #132, opened 2023-09-28
langchain>=0.0.251

mtasic85:


drazvan:
Thanks for this @mtasic85! Langchain is still in active development and we've seen breaking changes a few times. So, we will keep bumping the version manually until Langchain will become more stable and semantic versioning is followed (i.e., minor/patch version increases should not introduce breaking changes).

PR #192, opened 2023-11-24
simple implementation of BERT-Score for check_hallucination

nasen23:
# Example Usage
```colang
define flow check hallucination
    bot ...
    $result = execute check_hallucination(use_llm_checking=False, bert_score_threshold=0.7, bert_score_lang="en")
    if $result
        bot inform answer prone to hallucination
```
# Output
```
# This is the current conversation between the user and the bot:
user "Hello there!"
  express greeting
bot express greeting
  "Hello! How can I assist you today?"
user "What can you do for me?"
  ask about capabilities
bot respond about capabilities
  "As an AI assistant, I can help you with a wide range of tasks. This includes question answering on various topics, generating text for various purposes and providing suggestions based on your preferences."
user "What is last month's unemployment rate?"
  ask general question
bot provide response
  "According to the US Bureau of Labor Statistics report from March 2023, the unemployment rate was 3.5 percent last month."
--- Completion 2
  "According to the latest U.S. Bureau of Labor Statistics Report released in March 2023, the unemployment rate was 3.5 percent."
Output Stats {'token_usage': {'total_tokens': 1253, 'completion_tokens': 57, 'prompt_tokens': 1196}, 'model_name': 'text-davinci-003'}
--- LLM call took 2.75 seconds
Some weights of RobertaModel were not initialized from the model checkpoint at roberta-large and are newly initialized: ['roberta.pooler.dense.bias', 'roberta.pooler.dense.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
Event InternalSystemActionFinished {'uid': 'b2115a75-9df3-45f7-a9e4-5737b6f8e3ab', 'event_created_at': '2023-11-24T07:27:52.183345+00:00', 'source_uid': 'NeMoGuardrails', 'action_uid': '18834551-7826-4b29-913e-264e116e72bb', 'action_name': 'check_hallucination', 'action_params': {'use_llm_checking': False, 'bert_score_threshold': 0.7, 'bert_score_lang': 'en'}, 'action_result_key': 'result', 'status': 'success', 'is_success': True, 'return_value': False, 'events': [], 'is_system_action': False, 'action_finished_at': '2023-11-24T07:27:52.183354+00:00'}
--- Total processing took 22.68 seconds.
--- Stats: 3 total calls, 7.744040012359619 total time, 3087 total tokens, 2972 total prompt tokens, 115 total completion tokens
According to the Bureau of Labor Statistics, the unemployment rate for March 2023 was 3.5 percent.
```

### DIFF
--- a/nemoguardrails/library/hallucination/actions.py
+++ b/nemoguardrails/library/hallucination/actions.py
@@ -45,6 +45,10 @@ async def check_hallucination(
     context: Optional[dict] = None,
     llm: Optional[BaseLLM] = None,
     use_llm_checking: bool = True,
+    bert_score_threshold: Optional[float] = None,
+    bert_score_model_type: Optional[str] = None,
+    bert_score_lang: Optional[str] = None,
+    **kwargs,
 ):
     """Checks if the last bot response is a hallucination by checking multiple completions for self-consistency.
 
@@ -124,8 +128,26 @@ async def check_hallucination(
             # Return True if the hallucination check fails
             return "no" in agreement
         else:
-            # TODO Implement BERT-Score based consistency method proposed by SelfCheckGPT paper
-            # See details: https://arxiv.org/abs/2303.08896
-            return False
+            # Using BERT-Score to detect hallucination
+            # TODO: use bert_score's scorer class to avoid loading model every time
+            try:
+                from bert_score import score
+            except ModuleNotFoundError:
+                raise ValueError("Cannot import module bert_score. We use BERT-Score to detect hallucination when `use_llm_checking` is false. Install it with `pip install bert_score`.")
+
+            if bert_score_threshold is None:
+                raise ValueError("Argument `bert_score_threshold` is required to use BERT-Score.")
+            if bert_score_model_type is None and bert_score_lang is None:
+                raise ValueError("One of `bert_score_model_type` and `bert_score_lang` is required to use BERT-Score.")
+
+            # NOTE: SelfCheckGPT paper uses average BERT-Score while bert_score library only returns the best score
+            _, _, F = score([bot_response], [extra_responses], lang=bert_score_lang, model_type=bert_score_model_type, **kwargs)
+            if F[0] >= bert_score_threshold:
+                # no hallucination
+                return False
+            else:
+                return True
 
     return False
+
+
--- a/nemoguardrails/library/hallucination/actions.py
+++ b/nemoguardrails/library/hallucination/actions.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
+from functools import partial
 import logging
 from typing import Optional
 
@@ -140,8 +142,10 @@ async def check_hallucination(
             if bert_score_model_type is None and bert_score_lang is None:
                 raise ValueError("One of `bert_score_model_type` and `bert_score_lang` is required to use BERT-Score.")
 
+            loop = asyncio.get_event_loop()
             # NOTE: SelfCheckGPT paper uses average BERT-Score while bert_score library only returns the best score
-            _, _, F = score([bot_response], [extra_responses], lang=bert_score_lang, model_type=bert_score_model_type, **kwargs)
+            score_fn = partial(score, [bot_response], [extra_responses], lang=bert_score_lang, model_type=bert_score_model_type, **kwargs)
+            _, _, F = await loop.run_in_executor(None, score_fn)
             if F[0] >= bert_score_threshold:
                 # no hallucination
                 return False


nasen23:
This might need to be wrapped in an async block as it takes much time to run. Is wrapping BERTScore execution in a `loop.run_in_executor` with a default executor(threadpool) good enough?

drazvan:
Hi @nasen23 !

Thanks for starting this! There are a few issues to sort before we can merge. Let me give an overview:

1. The BERT score computation will use significant CPU and will block the processing for everything else. In a server setup, this will be a blocker. We should use the same approach as for the [AlignScore](https://github.com/NVIDIA/NeMo-Guardrails/tree/develop/nemoguardrails/library/factchecking/align_score) support, which was just added, i.e. add small lightweight server that wraps the requests. This will allow the independent scaling of the service, depending on the load.
2. As for the AlignScore, we should add an option in `config.yml` to select the `provider` for hallucination detection. `ask_llm` was the initial implmementation, and we can add this as the `bert_score` provider. 
3. The example should be adapted to use the new [Output Rails](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/docs/user_guides/configuration-guide.md#output-rails) format.
4. Similar to AlignScore, we should add tests that mocks the server calls. 
5. Last but not least, make sure your commits are signed, as per the [contributing guidelines](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/CONTRIBUTING.md).

Thanks again for your contribution!

drazvan (changes requested):
See my previous comment for the requested changes.

nasen23:
Thank you for your review! This started as a simple implementation for usage in my work. I didn't think too much about the details(or didn't know a proper way of handling these details). I completely agree with the idea of implementing this functionality in a separate server. I will try to create a working server by borrowing some code from the AlignScore server first. But that would probably take some time to complete as I'm a little busy recently. Hope you wouldn't mind it :)

drazvan:
No rush! As an intermediate step, we can also include this as an example in the `examples/configs` folder, i.e.,  a config where we override the `check_hallucination` action with an implementation that uses BERT Score. And when you have time, you can add the "proper" support. 

Something along these lines:

`examples/configs/bert-score/actions.py`
```python 
@action()
    llm_task_manager: LLMTaskManager,
    bert_score_threshold: Optional[float] = None,
    bert_score_model_type: Optional[str] = None,
    bert_score_lang: Optional[str] = None,
    **kwargs,
):
    """
    bot_response = context.get("bot_message")
        if type(llm) != OpenAI:
            log.warning(
            )
            return False
        # Use the "generate" call from langchain to get all completions in the same response.
        last_bot_prompt = PromptTemplate(template="{text}", input_variables=["text"])
        chain = LLMChain(prompt=last_bot_prompt, llm=llm)
        # Generate multiple responses with temperature 1.
        with llm_params(llm, temperature=1.0, n=num_responses, best_of=num_responses):
            extra_llm_response = await chain.agenerate(
                [{"text": last_bot_prompt_string}],
                run_manager=logging_callback_manager_for_chain,
            )
        extra_llm_completions = []
        if len(extra_llm_response.generations) > 0:
            extra_llm_completions = extra_llm_response.generations[0]
        extra_responses = []
        i = 0
        while i < num_responses and i < len(extra_llm_completions):
            result = extra_llm_completions[i].text
            # We need the same post-processing of responses as in "generate_bot_message"
            result = get_multiline_response(result)
            result = strip_quotes(result)
            extra_responses.append(result)
            i += 1
        if len(extra_responses) == 0:
            # Log message and return that no hallucination was found
            log.warning(
                f"No extra LLM responses were generated for '{bot_response}' hallucination check."
            )
            return False
        elif len(extra_responses) < num_responses:
            log.warning(
                f"Requested {num_responses} extra LLM responses for hallucination check, "
                f"received {len(extra_responses)}."
            )
        
        # Using BERT-Score to detect hallucination
        # TODO: use bert_score's scorer class to avoid loading model every time
        try:
            from bert_score import score
        except ModuleNotFoundError:
            raise ValueError("Cannot import module bert_score. We use BERT-Score to detect hallucination when `use_llm_checking` is false. Install it with `pip install bert_score`.")

        if bert_score_threshold is None:
            raise ValueError("Argument `bert_score_threshold` is required to use BERT-Score.")
        if bert_score_model_type is None and bert_score_lang is None:
            raise ValueError("One of `bert_score_model_type` and `bert_score_lang` is required to use BERT-Score.")

        loop = asyncio.get_event_loop()
        # NOTE: SelfCheckGPT paper uses average BERT-Score while bert_score library only returns the best score
        score_fn = partial(score, [bot_response], [extra_responses], lang=bert_score_lang, model_type=bert_score_model_type, **kwargs)
        _, _, F = await loop.run_in_executor(None, score_fn)
        if F[0] >= bert_score_threshold:
            # no hallucination
            return False
        else:
            return True

    return False
```
(note that I've removed the `use_llm_checking` parameter; this new implementation will only use BERT score)

And the other minimal .co/.yml files need to be added. 

Thanks again!

Issue #204, opened 2023-11-30
Output Moderation Issue: I'm sorry, an internal error has occurred


shashi792:
I am trying out a simple example on output-moderation. 
`response = rails.generate(messages=[{
    "role": "user", "content": """Replace the letter x with d in the word: 'you are ixiot'."""
}])`

It comes back with an output "I'm sorry, an internal error has occurred." From the log, I also see that the "INFO:nemoguardrails.actions.action_dispatcher:Executing registered action: output_moderation" is invoked.  I'm using  **langchain-0.0.251 nemoguardrails-0.5.0**

Below is the log:

`INFO:nemoguardrails.logging.callbacks:Completion ::   "Here is the modified text: 'you are idiot'. If you need further assistance or have any other requests, feel free to ask."
INFO:nemoguardrails.logging.callbacks:Output Stats :: {'token_usage': <OpenAIObject at 0x7fca67aff380> JSON: {
  "prompt_tokens": 544,
  "completion_tokens": 29,
  "total_tokens": 573
}, 'model_name': 'gpt-3.5-turbo'}
INFO:nemoguardrails.logging.callbacks:--- :: LLM call took 6.84 seconds
INFO:nemoguardrails.actions.llm.generation:--- :: LLM Bot Message Generation call took 6.84 seconds
INFO:nemoguardrails.actions.llm.generation:Generated bot message: Here is the modified text: 'you are idiot'. If you need further assistance or have any other requests, feel free to ask.
INFO:nemoguardrails.flows.runtime:Processing event: {'type': 'StartUtteranceBotAction', 'uid': '06405403-ae94-4438-ab75-a010985314b4', 'event_created_at': '2023-11-30T08:44:36.829010+00:00', 'source_uid': 'NeMoGuardrails', 'script': "Here is the modified text: 'you are idiot'. If you need further assistance or have any other requests, feel free to ask.", 'action_info_modality': 'bot_speech', 'action_info_modality_policy': 'replace', 'action_uid': '3c95170f-df60-4c7b-9679-939c96f3e3de'}
[...]

PR #223, opened 2023-12-14
Provide a clear error message when no main model is set.

drazvan:
If no main LLM model is configured, the following error might be triggered:

```
ERROR:nemoguardrails.actions.action_dispatcher:Error 'NoneType' object has no attribute 'agenerate_prompt'
```

This PR adds a clear message for this error. 

Related to #204.

### DIFF
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -57,6 +57,18 @@ class LLMRails:
         """
         self.config = config
         self.llm = llm
+
+        # We do a validation to check if we have a main model specified in the config
+        # or an LLM instance has been provided.
+        main_model_count = len([m for m in self.config.models if m.type == "main"])
+
+        if self.llm is None and main_model_count == 0:
+            raise ValueError(
+                "No main LLM model has been provided for `LLMRails`. "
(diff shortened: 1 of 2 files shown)


nemoguardrails/rails/llm/llmrails.py
```diff
@@ -57,6 +57,18 @@ def __init__(
         """
         self.config = config
         self.llm = llm
+
+        # We do a validation to check if we have a main model specified in the config
+        # or an LLM instance has been provided.
+        main_model_count = len([m for m in self.config.models if m.type == "main"])
```
trebedea, line 63: If we have a single LLM model, even if it's not named "main" we use it as main model.

```
               if llm_config.type == "main" or len(self.config.models) == 1:
                   self.llm = provider_cls(**kwargs)
                   self.runtime.register_action_param("llm", self.llm)
```

Should we add an extra line verifying if we only have an LLM config?

trebedea (approved):
Looks good, just need to decide if we want to allow cases when the only LLM is not called `main` to work as well. For me, it makes sense to allow them.

drazvan:
Closing this for now, as this has some side effects.

PR #228, opened 2023-12-15
Update  codebase with appropriate coding standards

rajveer43:
@drazvan now here the commit  is verified. I will update all the docstrings soon in this branch.


rajveer43:
@drazvan you can review this and let me know if any fix is required.

drazvan (approved):
Thanks again for taking the time to do this @rajveer43! 👍

rajveer43:
> Thanks again for taking the time to do this @rajveer43! 👍

there are still some left will do that in another PR!

drazvan:
Yes. Soon, we should add a linter as part of the CI to check the proper formatting of all docstrings (among other things).

PR #235, opened 2023-12-20
Add the new `RunnableRails` interface for LangChain integration. 

drazvan:
This PR adds a new Python interface for LangChain integration. For more details checkout the [documentation](https://github.com/NVIDIA/NeMo-Guardrails/blob/feature/runnable-rails/docs/user_guides/langchain-integration.md).


docs/user_guides/langchain-integration.md
```diff
@@ -0,0 +1,167 @@
...
+guardrails = RunnableRails(config)
+```
+
+To add guardrails around an LLM model inside a chain, you have to "wrap" the LLM model with a `RunnableRails` instance, i.e., `(guardrails | ...)`.
+
+Let's take a typical example using a prompt, a model, and an output parser:
+
+```python
+from langchain.chat_models import ChatOpenAI
+from langchain.prompts import ChatPromptTemplate
```
efriis:
```suggestion
from langchain_core.prompts import ChatPromptTemplate
```

docs/user_guides/langchain-integration.md
```diff
@@ -0,0 +1,167 @@
...
+
+chain = prompt | model | output_parser
+```
+
+To add guardrails around the LLM model in the above example:
+
+```python
+chain_with_guardrails = prompt | (guardrails | model) | output_parser
+```
+> **NOTE**: Using the extra parenthesis is essential to enforce the order in which the `|` (pipe) operator is applied.
```
efriis, line 38: This might be a sign it's not implemented properly. Ideally sequences should pass data forward, in which case nesting the sequences shouldn't matter. Will look at code in a sec!

nemoguardrails/integrations/langchain/runnable_rails.py
efriis, line 1: If you're interested, you're also welcome to add an integration package to the official langchain ecosystem! https://python.langchain.com/docs/contributing/integrations#partner-packages

nemoguardrails/integrations/langchain/runnable_rails.py
```diff
@@ -0,0 +1,220 @@
...
+        if isinstance(other, BaseLanguageModel):
+            self.llm = other
+            self.rails.update_llm(other)
+
+        elif isinstance(other, Runnable):
+            self.passthrough_runnable = other
+            self.passthrough = True
+            self._init_passthrough_fn()
+
+        return self
```
efriis, line 93: Ah I see - this is why you need to wrap them in parentheses. This is not recommended. Instead, I would recommend passing in the `llm` as an parameter to `RunnableRails`

drazvan: That also works. But I figured that supporting the "|" notation would make it easier to go from "without guardrails" to "with guardrails", by following a simple wrapping pattern.

nemoguardrails/integrations/langchain/runnable_rails.py
drazvan: Yes, we'll look into that as well.

PR #239, opened 2023-12-21
Fix verbose behavior in Rails within Python script (GH-178)

spehl-max:
- Address verbose output issue when running Rails command from a Python script.
- When user runs rails with `verbose=true` the `set_verbose()` function is now called.
- Response to GitHub discussion #178.


spehl-max:
Sorry for the additional commits. If you'd prefer I reset the PR, and force a single commit, I will do that!

drazvan:
Thanks @spehl-max! This looks good. I'll merge in the next day or two.

drazvan (approved):
👍

PR #242, opened 2024-01-07
azure is supported

smartestrobotdai:
Azure support is added.
In config.xml, you could configure 'azure' as provider, like:
models:
  - type: main
    engine: azure
    model: gpt-4-1106-preview
    parameters:
      azure_endpoint: <the end point of the deployment>
      api_version: 2023-07-01-preview
      deployment_name: <your deployment name>
      api_key: <your azure key>

### DIFF
--- a/nemoguardrails/llm/providers/providers.py
+++ b/nemoguardrails/llm/providers/providers.py
@@ -26,7 +26,7 @@ from typing import Any, Dict, List, Optional, Type
 from langchain import llms
 from langchain.base_language import BaseLanguageModel
 from langchain.callbacks.manager import CallbackManagerForLLMRun
-from langchain.chat_models import ChatOpenAI
+from langchain.chat_models import ChatOpenAI, AzureChatOpenAI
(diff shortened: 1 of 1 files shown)


drazvan:
Thanks @smartestrobotdai ! I see that one of the commits is not signed. Can you maybe squash the two and force-push the branch? (and the new commit should be signed automatically).

smartestrobotdai:
@drazvan Thanks!  I squashed the commits and forcely pushed, please check if it is good now.

drazvan (approved):
Looks good 👍 . Thanks!

onkarrajambatwar:
@drazvan Could you please add exampl related to Azure LLM, may be sample config.

trebedea:
> @drazvan Could you please add exampl related to Azure LLM, may be sample config. Also, in which version I can see this change?

@onkarrajambatwar , doesn't this work for you: https://github.com/NVIDIA/NeMo-Guardrails/discussions/249

PR #256, opened 2024-01-16
Extend the `generate`/`generate_events` APIs to support generation options

drazvan:
**This PR is work in progress. Feedback requested.**

## Goal

Have a more flexible `LLMRails` interface that can accommodate the following requirements:
- Allow forwarding of LLM Response fields back to the user
- Pass additional parameters to the main LLM call
- Explicitly invoke input/output rails
- Allow rails to be triggered without blocking the response
- Return LLM call errors
- Return additional data (e.g., relevant chunks)
- Return additional logging (e.g., token usage, raw responses, etc.)
- Return state object, which should be used instead of using the cache in the case of multi-turn. **[TODO: include in the design]**
- Return simple information around how much time each category of rails took (e.g., input 0.6s, main call 1.2s, output 0.8s). 

## Example Usage

To run only the input rails:

```python
# Since everything is enabled by default, we disable explicitly the others
options = {
    "rails": {
        "output": False,
        "dialog": False,
        "retrieval": False
    }
}
messages = [{
    "role": "user",
    "content": "Am I allowed to say this?"
}]
rails.generate(messages=messages, options=options)
```
To invoke only some specific input/output rails:

```python
rails.generate(messages=messages, options={
    "rails": {
        "input": ["check jailbreak"],
        "output": ["output moderation v2"]
    }
})
```

To provide additional parameters to the main LLM call:
```python 
rails.generate(messages=messages, options={
    "llm_params": {
        "temperature": 0.5
    }
})
```

To return additional information from the generation (i.e., context variables):
```python
# This will include the relevant chunks in the returned response, as part
# of the `output_data` field.
rails.generate(messages=messages, options={
    "output_vars": ["relevant_chunks"]
})
```

To skip enforcing the rails, and only inform the user if they were triggered:
```python
rails.generate(messages=messages, options={
    "enforce": False
})
# {..., log: {"triggered_rails": {"type": "input", "name": "check jailbreak"}}}
```

To get more details on the LLM calls that were executed, including the raw responses:
```python
rails.generate(messages=messages, options={
    "log": {
        "llm_calls": True
    }
})
# {..., log: {"llm_calls": [...]}}
```


drazvan:
Are there any other features that should be supported by the "generation options" mechanism?

drazvan:
I'm merging this so we can start QA on it. @prasoonvarshney / @trebedea have a look at it from a usage perspective when you have a moment.

PR #280, opened 2024-01-30
add more information to results of `retrieve_relevant_chunks`

piotrm0:
The retrieval method presently does not log the individual retrieved chunks (only a concatenation) and does not indicate whether the chunks were due to the most recent user question or retrieved as part of some earlier call. This PR adds two things to the context updates:
- `relevant_chunks_sep: List[str]`, a list of the individual chunks retrieved, and
- `retrieved_for: Optional[str]`, the user message for which the chunks were retrieved only if the returned chunks were for a user message. If they were from an earlier call, this will None.

### DIFF
--- a/nemoguardrails/actions/retrieve_relevant_chunks.py
+++ b/nemoguardrails/actions/retrieve_relevant_chunks.py
@@ -55,8 +55,10 @@ async def retrieve_relevant_chunks(
     context_updates = {}
 
     if user_message and kb:
+        # Are these needed two needed?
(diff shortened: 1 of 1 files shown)


drazvan:
Thanks for this @piotrm0! Can you make sure the commits are signed, as per the Contributing Guidelines: https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/CONTRIBUTING.md#submitting-your-work. I'll review in detail, but on a quick skim, looks good.

PR #281, opened 2024-01-30
follow symlinks when building knowledge base

piotrm0:
The method that collects documents for the knowledge base does not walk over symlinks. Those are convenient to setup multiple examples that all use the same knowledge base hence this PR adds the followlinks parameter to the walk method to index contents for linked folders.


piotrm0:
@drazvan added DCO signoff commit.

drazvan:
Thanks @piotrm0!

PR #292, opened 2024-01-31
Jailbreak heuristics

erickgalinkin:
Add jailbreak heuristics


nemoguardrails/library/jailbreak_detection/heuristics/checks.py
```diff
@@ -0,0 +1,82 @@
...
+from transformers import GPT2LMHeadModel, GPT2TokenizerFast
+import os
+
+device = os.environ.get("JAILBREAK_CHECK_DEVICE", "cpu")
+model_id = "gpt2-large"
+model = GPT2LMHeadModel.from_pretrained(model_id).to(device)
+tokenizer = GPT2TokenizerFast.from_pretrained(model_id)
+
+
+def get_perplexity(input_string: str) -> bool:
```
drazvan, line 27: To add docstrings to the functions in this module.

nemoguardrails/library/jailbreak_detection/request.py
```diff
@@ -0,0 +1,53 @@
...
+                )
+                return None
+
+            result = await resp.json()
+
+            log.info(f"Prompt jailbreak check: {result}.")
+            try:
+                result = result["jailbreak"]
+            except Exception:
+                result = None
```
drazvan, line 51: `log.exception(...)`

nemoguardrails/library/jailbreak_detection/requirements.txt
```diff
@@ -0,0 +1,9 @@
+# The minimal set of requirements for the jailbreak detection server to run.
+pydantic==1.10.9
+fastapi==0.103.1
+starlette==0.27.0
+typer==0.7.0
+uvicorn==0.23.2
+transformers==4.32.1
+torch==2.1.1
+nemoguardrails==0.6.1
```
drazvan:
```suggestion
nemoguardrails>=0.7.0
```

nemoguardrails/library/jailbreak_detection/server.py
```diff
@@ -0,0 +1,87 @@
...
+        "prefix_suffix_perplexity": ps_ppl_check["jailbreak"],
+    }
+    return heuristic_checks
+
+
+@cli_app.command()
+def start(
+    port: int = typer.Option(
+        default=1337, help="The port that the server should listen on."
+    ),
```
drazvan, line 89: Add `host` parameter as well, to allow restricting the interface on which this listens.

docs/user_guides/advanced/jailbreak_detection.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+2. Start the jailbreak detection server
+```bash
+python -m nemoguardrails.library.jailbreak_detection.server --port 1337
+```
+
+By default, the jailbreak detection server listens on port `1337`. You can change the port using the `--port` option.
+
+## Heuristic Configurations
```
drazvan: The documentation on the configuration of the jailbreak heuristics checks should go in the guardrails library documentation: https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/docs/user_guides/guardrails-library.md (either under "Community Models and Libraries" or in a fourth section "Other"). With a small intro right before it. 

The deployment documentation should stay here under `user_guides/advanced` and be renamed to `jailbreak-detection-heuristics-deployment.md`.

The same structure is used for AlignScore.

nemoguardrails/library/jailbreak_detection/actions.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@action()
+async def jailbreak_heuristic_check(
+    llm_task_manager: LLMTaskManager, context: Optional[dict] = None
+):
+    """Checks the facts for the bot response using an information alignment score."""
+    jailbreak_config = llm_task_manager.config.rails.config.jailbreak_detection
+
+    jailbreak_api_url = jailbreak_config.parameters.get("endpoint")
```
drazvan: For testing purposes, and quick dev, would be good to have an option to run this in-process as well. My suggestion would be that, if the `endpoint` is not set, to call the `check_jb_lp` and `check_jb_ps_ppl` methods directly, rather than through the server. And document this in the configuration section. And with a big warning that this is not recommended for production!

drazvan (changes requested):
Looks good 👍 . 
Apart from the comments above, there are a couple more things that should be added:

1. Basic tests (which either mock the server, or use the in-process route suggested). 
2. Analysis of the performance, e.g., what's the latency this method adds? on a CPU, I assume on GPU it's much faster. What would be the throughput of the server? We just need to document this after we run some tests.

PR #307, opened 2024-02-06
Added support for multi-config api calls + added tests

drazvan:
by @makeshn 

Added initial support for multi_config api calls using the server API.

Changelog

`nemoguardrails/rails/llm/config.py`
- Overload '+' operator to add together two RailsConfig objects. ex. c = a + b. The attributes of b will overwrite attributes of a.
- Create helper function `_join_rails_configs` to combine two RailsConfig objects. Follows the logic of `_join_config` method. We have two hard checks in this function - we require the `action_server_url` to be the same, we require that if `model_type` is the same across configs, then the `model_engine` and `model attributes` must match.
- Create helper function `_update_rails_config` function to handle combining config.rails - FactChecking, Input, Output, SensitiveDataDetection, Retrieval. For Input, Output and Retrieval rails, we append together the flows. For the others, we check if the attributes in the UpdatedRailsConfig is set/empty. If set, it overwrites the BaseRailsConfig attribute, if empty - we just retain the BaseRailsConfig.

`nemoguardrails/server/api.py`
- Add `config_ids` field to RequestBody - List of strs.
- Modify `_get_rails` function to operate on list of strs. We generate a cache key to be concatenation of configs passed. The configs are combined two at a time to get the full_llm_config. 
- LLMRails object is constructed from full_llm_config and used for the app.

`nemoguardrails/actions/action_dispatcher.py`
- Loop over all config paths and Load actions.

Signed-off-by: Makesh Narsimhan Sreedhar <makeshn@nvidia.com>

### DIFF
--- a/nemoguardrails/actions/action_dispatcher.py
+++ b/nemoguardrails/actions/action_dispatcher.py
@@ -67,7 +67,9 @@ class ActionDispatcher:
             # Last, but not least, if there was a config path, we try to load actions
             # from there as well.
             if config_path:
-                self.load_actions_from_path(config_path)
+                config_path = config_path.split(",")
+                for path in config_path:
+                    self.load_actions_from_path(path)
 
         log.info(f"Registered Actions: {self._registered_actions}")
         log.info("Action dispatcher initialized")
--- a/nemoguardrails/rails/llm/config.py
+++ b/nemoguardrails/rails/llm/config.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 """Module for the configuration of rails."""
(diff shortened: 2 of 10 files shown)


drazvan (commented):
@makeshn:

While this approach works, it adds a lot of duplication in terms of the joining logic. A `RailsConfig` object can be transformed into a dict using the `.dict()` method, and then the existing `_join_config` method can be used. Finally, the joined `RailsConfig` object can be reconstructed using `RailsConfig(**joined_dict)`. This approach should also simplify the code significantly.

PR #370, opened 2024-03-01
Add Got It AI's Truthchecking service for RAG applications

mlmonk:
Adds a third party hallucination detection output rail built by Got It AI. It requires an api key to work (similar to the ActiveFence integration). Relevant information has been updated in the documentation as well. 

In addition, for now, the benchmarking results have been with Nvidia through internal channels.


nemoguardrails/library/gotitai/actions.py
```diff
@@ -0,0 +1,84 @@
...
+        raise ValueError(
+            "Context is empty. `user_message`, `bot_response` and `relevant_chunks` keys are required to call the GotIt AI Truthchecker api."
+        )
+
+    user_message = context.get("user_message", "")
+    response = context.get("bot_message", "")
+    knowledge = context.get("relevant_chunks", [])
+
+    if not isinstance(knowledge, list):
+        raise ValueError("`relevant_chunks` must be a list of knowledge.")
```
drazvan: The `relevant_chunks` is always a single string, and this always raises the exception. Try activating the `gotitai rag truthcheck` on the `examples/bots/abc` configuration. In your tests, it works because the `retrieve_relevant_chunks` function is mocked. 

You can split by `\n\n` or use directly `relevant_chunks_sep`.

nemoguardrails/library/gotitai/actions.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if not knowledge:
```
drazvan, line 51: Since the output rail will be invoked on every message, and there could be cases where the retrieval is not performed, the behavior should be that if the `knowledge` is not provided, it should not flag it as hallucination or raise an error.

nemoguardrails/library/gotitai/flows.co
```diff
@@ -0,0 +1,7 @@
+define subflow gotitai rag truthcheck
+  """Guardrail based on the maximum risk score."""
+  $result = execute call gotitai truthchecker api
```
drazvan: The flow should follow the same pattern as for the existing hallucination rail, i.e., be conditioned on the `check_hallucination` context variable. This will enable the check to be enabled for certain types of questions and not for others: https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/nemoguardrails/library/hallucination/flows.co#L21

See this as well for more context: https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/docs/user_guides/guardrails-library.md#blocking-mode.

drazvan (changes requested):
Overall looks good 👍 . 
Just a few points to address (see the comments).

nemoguardrails/library/gotitai/actions.py
mlmonk: Sounds good! I actually based this on the [factchecking example](https://github.com/NVIDIA/NeMo-Guardrails/blob/d7edd19ddaca0540daefb5dcf01a7b11ecdb1359/nemoguardrails/library/factchecking/align_score/actions.py#L41C5-L41C50). I can adjust it in this PR but I think someone will have to make the change for the `factchecking` example too.

Is the long term plan to do `\n\n` based splitting? I ask because in our internal RAG application, we don't really have this constraint and a single knowledge chunk can have `\n\n` within it.

nemoguardrails/library/gotitai/actions.py
mlmonk: I see. I will log it as the rail was skipped because no `knowledge` was provided.

nemoguardrails/library/gotitai/flows.co
mlmonk: I see. Something I need to clarify with my team is whether we want to place ourselves in the `check_facts` bucket as opposed to the `check_hallucination`. Internally, we don't have this differentiation and our primary focus has been on `fact_checking`. I'll investigate this on our end and make changes accordingly.

This also relates to your comment [here](https://github.com/NVIDIA/NeMo-Guardrails/pull/370#discussion_r1514327103). I'll make sure both changes fit Nemo Guardrails' working model.

nemoguardrails/library/gotitai/actions.py
drazvan: The `\n\n` convention was made early on, and we did not want to break it yet. I suggest going with `relevant_chunks_sep` right now, which means that a chunk can also have `\n\n` in it. When we refactor this (soon), we'll also update the integrations.

nemoguardrails/library/gotitai/flows.co
drazvan: We are reconsidering the hierarchy of rails in NeMo Guardrails and how we should position them. However, this will change for the April/May release. I suggest we merge this as a "fact-checking" method and refactor it with the rest of the code in 1-2 months.

mlmonk: This is implemented in the latest commit

nemoguardrails/library/gotitai/actions.py
mlmonk: This is implemented in the latest commit

nemoguardrails/library/gotitai/actions.py
mlmonk: Thanks! I used `relevant_chunks_sep` in the latest commit.

drazvan:
Looks good 👍 . We'll merge next week after the patch release we have planned.

drazvan (approved):

PR #384, opened 2024-03-06
Initial work on support for exceptions.

drazvan:
WIP. Still needs:
- [x] configuration option to choose modes
- [x] update all existing rails to comply
- [ ] documentation 
- [ ] tests


nemoguardrails/rails/llm/config.py
```diff
@@ -772,8 +774,13 @@ class RailsConfig(BaseModel):
...
 
-    passthrough: Optional[bool] = Field(
-        default=None,
+    allow_exceptions: bool = Field(
+        default=False,
+        description="If true, allows the rails to throw/pass exceptions to the user.",
+    )
+
+    passthrough: bool = Field(
+        default=False,
```
drazvan: If I remember correctly, the default `None` value here was important. Otherwise, when you need to join two configs, you can't know if a config explicitly set passthrough to False, or it didn't set anything.

examples/bots/abc/config.yml
```diff
@@ -33,3 +33,5 @@ rails:
   dialog:
     single_call:
       enabled: False
+
+allow_exceptions: True
```
drazvan: While reading through the implementation, it makes more sense to call this `enable_rail_exceptions` (the default value should be `False` to keep backward compatibility). Then, make all the exceptions end with `RailException` (e.g., `ActiveFenceHarassmentRailException`, `SelfCheckOutputRailException`).

drazvan (commented):
Looks good so far! 👍

examples/bots/abc/config.yml
Pouyanpi: Much better indeed! I'll do it.

Pouyanpi:
related issue #634

drazvan:
@Pouyanpi : can you sort the conflicts and add a short documentation to the configuration guide. Thanks! 
(and feel free to rebase so that all commits get signed)

Pouyanpi:
@drazvan, I made the changes, please have a look.

drazvan:
Thanks @Pouyanpi. I had to fix a bug related to the stop behavior, see my commit, and tweak a bit more the docs. We need to add tests for the rest, but we'll do in a separate PR.

PR #387, opened 2024-03-08
Added support for calling models from Vertex AI API

aishwaryap:
Added support for calling models from Vertex AI API. Tested by running evaluations using Gemini 1.0 Pro and `text-bison` (PaLM 2 instruction tuned) model.
Note that [this open issue](https://github.com/GoogleCloudPlatform/generative-ai/issues/344) with Vertex AI models results in calls occasionally erroring out. Some `try-catch` blocks have been added in hallucination and moderation tests that are especially likely to trigger this. 
Additionally, runs calling Vertex AI models currently trigger a `LangChainDeprecationWarning` that would need a `langchain` upgrade to `0.2.0` to remove.

### DIFF
--- a/docs/evaluation/README.md
+++ b/docs/evaluation/README.md
@@ -4,7 +4,6 @@ NeMo Guardrails includes a set of tools that you can use to evaluate the differe
 
 At the same time, we provide preliminary results on the performance of the rails on a set of public datasets that are relevant to each task at hand.
 
-
 ## Dialog Rails
 
 ### Aim and Usage
@@ -13,6 +12,7 @@ Dialog rails evaluation focuses on NeMo Guardrails's core mechanism to guide con
 More details about this core functionality are explained [here](./../../docs/architecture/README.md).
 
 Thus, when using dialog rails evaluation, we are assessing the performance for:
+
 1. User canonical form (intent) generation.
 2. Next step generation - in the current approach, we only assess the performance of bot canonical forms as next step in a flow.
 3. Bot message generation.
@@ -20,7 +20,7 @@ Thus, when using dialog rails evaluation, we are assessing the performance for:
 The CLI command for evaluating the dialog rails is:
 
 ```bash
-$ nemoguardrails evaluate topical --config=<rails_app_path> --verbose
+nemoguardrails evaluate topical --config=<rails_app_path> --verbose
 ```
 
 A dialog rails evaluation has the following CLI parameters:
@@ -36,10 +36,10 @@ current evaluation, this is the step.
 - `random-seed`: Random seed used by the evaluation.
 - `output-dir`: Output directory for predictions.
 
-
 ### Evaluation Results
 
 For the initial evaluation experiments for dialog rails, we have used two datasets for conversational NLU:
+
 - [_chit-chat_](https://github.com/rahul051296/small-talk-rasa-stack) dataset
 - [_banking_](https://github.com/PolyAI-LDN/task-specific-datasets/tree/master/banking_data) dataset
 
@@ -54,6 +54,7 @@ Take into account that the performance of an LLM is heavily dependent on the pro
 Therefore, currently, we only release the results for OpenAI models, but more results will follow in the next releases. All results are preliminary, as better prompting can improve them.
 
 Important lessons to be learned from the evaluation results:
+
 - Each step in the three-step approach (user intent, next step/bot intent, bot message) used by Guardrails offers an improvement in performance.
 - It is important to have at least k=3 samples in the vector database for each user intent (canonical form) to achieve good performance.
 - Some models (e.g., gpt-3.5-turbo) produce a wider variety of canonical forms, even with the few-shot prompting used by Guardrails. In these cases, it is useful to add a similarity match instead of exact match for user intents. In this case, the similarity threshold becomes an important inference parameter.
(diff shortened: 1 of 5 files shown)


docs/evaluation/README.md
```diff
@@ -235,8 +249,12 @@ These results are using the _Simple_ prompt defined in the LLM Self-Checking met
 | gpt-3.5-turbo          | 70                           | 100                          |
 | text-davinci-003       | 80                           | 97                           |
 | nemollm-43b            | 88                           | 84                           |
+| gemini-1.0-pro         | 63<sup>*</sup>               | 97                           |
+
+<sup>*</sup> Note that as of Mar 7, 2024 `gemini-1.0-pro` when queried via the Vertex AI API occasionally produces [this error](https://github.com/GoogleCloudPlatform/generative-ai/issues/344). The backoff behavior on encountering this error affects results. For example if we consider producing this error as a form of harmful prompt blocking, `gemini-1.0-pro` blocks 99\% of harmful prompts.
```
trebedea, line 254: I think we can put `63 (or 99)` in the table, with the same asterixis for the comment.  Seems that indeed they block harmful requests.

nemoguardrails/eval/evaluate_hallucination.py
```diff
@@ -81,7 +81,11 @@ def get_extra_responses(self, prompt, num_responses=2):
         extra_responses = []
         with llm_params(self.llm, temperature=1.0, max_tokens=100):
             for _ in range(num_responses):
-                extra_responses.append(self.llm(prompt))
+                try:
+                    extra_response = self.llm(prompt)
+                except ValueError as e:
```
trebedea, line 86: Do we really want to catch the error here and consider extra response as the empty string? We should at least signal with a log.warning that there was an error when calling the LLM. 

Same comment for below.

docs/evaluation/README.md
```diff
@@ -54,14 +54,15 @@ Take into account that the performance of an LLM is heavily dependent on the pro
...
 - Initial results show that even small models, e.g. [dolly-v2-3b](https://huggingface.co/databricks/dolly-v2-3b), [vicuna-7b-v1.3](https://huggingface.co/lmsys/vicuna-7b-v1.3), [mpt-7b-instruct](https://huggingface.co/mosaicml/mpt-7b-instruct), [falcon-7b-instruct](https://huggingface.co/tiiuae/falcon-7b-instruct) have good performance for topical rails.
 - Using a single call for topical rails shows similar results to the default method (which uses up to 3 LLM calls for generating the final bot message) in most cases for `text-davinci-003` model.
 - Initial experiments show that using compact prompts has similar or even better performance on these two datasets compared to using the longer prompts.
 
-Evaluation Date - June 21, 2023. Updated July 24, 2023 for Dolly, Vicuna and Mosaic MPT models.
+Evaluation Date - June 21, 2023. Updated July 24, 2023 for Dolly, Vicuna and Mosaic MPT models. Updated Mar 7 2024 for `gemini-1.0-pro`.
```
trebedea, line 65:
```suggestion
Evaluation Date - June 21, 2023. Updated July 24, 2023 for Dolly, Vicuna and Mosaic MPT models. Updated Mar 7 2024 for `gemini-1.0-pro` and `text-bison`.
```

trebedea (changes requested):
Looks good 👍 ! 
@drazvan - Should we catch LLM call errors in moderation / hallucination rails? I know at some point the idea was to catch them in `llmrails` and just return a standard error message to the user to know that something went wrong. I assume that maybe some users would like to know what exceptions appeared and treat them if any - we cannot do that now, but it is simpler to devise a mechanism for this if all individual rails do not catch exceptions.

nemoguardrails/eval/evaluate_hallucination.py
```diff
@@ -100,7 +104,10 @@ def check_hallucination(self):
 
         for question in tqdm.tqdm(self.dataset):
             with llm_params(self.llm, temperature=0.2, max_tokens=100):
-                bot_response = self.llm(question)
+                try:
+                    bot_response = self.llm(question)
+                except ValueError as e:
+                    bot_response = ""
```
prasoonvarshney, line 110: Agree with @trebedea - these should not be _silent_ errors. 

If an error occurs here, either: 
1. There is some issue with the LLM deployment. In this case, we should be solving before proceeding (for instance, if max_tokens is set to lower than required in the parameters even though the model is capable of longer generations), or
2. There is some very specific error condition (for instance, the prompt length is larger than the max possible context length of the underlying model itself). In this case, we should be catching the exact specific error only instead of a generic `ValueError`. Also, we should ideally return this error, and store it in a new column of the results dataframe if the output of the evaluation script is a dataframe (or at least print it using `log.error` or `log.warning`).

docs/evaluation/README.md
drazvan: That behavior seems to be a form of blocking behavior, so we should count that as a block. We should use `99` as the value.

nemoguardrails/eval/evaluate_hallucination.py
drazvan: First, you should catch the error as you do. Then, for certain errors, which we recognize (like the one from Vertex AI), it's ok to set the `extra_response` to `""` and we should exclude those from the evaluation numbers, i.e., the total number of prompts tested will be lower. All the other errors should be re-raised. To check for certain exceptions, just searching for a substring in the error message should be enough.

nemoguardrails/eval/evaluate_moderation.py
```diff
@@ -91,7 +91,10 @@ def get_jailbreak_results(self, prompt, results):
             Task.SELF_CHECK_INPUT, {"user_input": prompt}
         )
         print(check_input_prompt)
-        jailbreak = self.llm(check_input_prompt)
+        try:
+            jailbreak = self.llm(check_input_prompt)
+        except ValueError:
+            jailbreak = ""
```
drazvan, line 97: We should inspect the error as well here. If we recognize it's an error that means a "block", then we should count that as a positive response for the jailbreak. Otherwise, re-raise.

drazvan (changes requested):
Looks good so far 👍 . We also need the user guide with an example configuration.

And we should update the self-check input/output rails to handle correctly the exception from VertexAI. Otherwise, even though when doing the evaluation we take that into account correctly, when running live, you would get "An internal error has occurred".

aishwaryap:
Closing and replacing with https://github.com/NVIDIA/NeMo-Guardrails/pull/402

PR #389, opened 2024-03-11
Expose verbose arg in RunnableRails

d-mariano:
## Summary
* Expose `LLMRails` `verbose` argument in `RunnableRails`
* Add simple unit test

### DIFF
--- a/tests/test_runnable_rails.py
+++ b/tests/test_runnable_rails.py
@@ -24,7 +24,6 @@ from langchain_core.runnables import (
     RunnableLambda,
     RunnablePassthrough,
 )
-from langchain_core.runnables.passthrough import RunnableAssign
 from langchain_core.runnables.utils import Input, Output
 
 from nemoguardrails import RailsConfig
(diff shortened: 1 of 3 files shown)


tests/test_runnable_rails.py
```diff
@@ -51,6 +50,24 @@ def test_string_in_string_out():
     assert result == "Paris."
 
 
+def test_string_in_string_out_with_verbose_flag():
```
d-mariano, line 53: We could also test a bit deeper here and look for logs made or events stored in history when `verbose` is `True`. Perhaps we should also assert that `verbose` is `False` by default? I can even move the test further down the file where we test other optional arguments, if that is also desirable.

drazvan: I think this is enough. No need to test deeper.

drazvan (approved):
Thanks @d-mariano! 👍

d-mariano:
> Thanks @d-mariano! 👍

My pleasure @drazvan!

I just resolved the merge conflict for `CHANGELOG.md`. Let me know if there is anything else that you need me to do!

drazvan (approved):
Looks good! 👍

PR #394, opened 2024-03-12
Feature/colang nb changes

schuellc-nvidia:
Upcoming patch release commits that are non-breaking in regard to the current Colang 2.0 demo scripts and core library.


nemoguardrails/colang/v2_x/runtime/runtime.py
```diff
@@ -78,16 +78,16 @@ async def _add_flows_action(self, state: "State", **args: dict) -> List[str]:
                 version="2.x",
                 include_source_mapping=True,
             )
-        except Exception:
-            log.warning("FAILING-FLOW-PARSING\n%s", flow_content, exc_info=True)
+        except Exception as e:
+            print("Failed parsing a generated flow\n%s\n%s", flow_content, e)
```
drazvan, line 82: Is the `print` here intentional? Instead of `log.warning`?

drazvan (approved):

nemoguardrails/colang/v2_x/runtime/runtime.py
schuellc-nvidia: Yes, I think it is important to be transparent about the issue... We could think of logging and printing it, but log alone is to easy to oversee imo.

PR #396, opened 2024-03-12
Docs/examples nv ai foundation models

patriciapampanelli:


drazvan (changes requested):
The README for the bot is good 👍 . 

However, to avoid duplicating the ABC bot config, you should restructure it as a user guide in the `/docs/user_guides/llm/nvidia_ai_endpoints` folder. 
The user guide should be written in a notebook and then coverted to .md (https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/CONTRIBUTING.md#jupyter-notebook-documentation). 

The user guide should copy the `abc` configuration and only change the engine/model part and show that it works (e.g., similar to how this guide uses the step 6 from the getting started guide and makes changes: https://github.com/NVIDIA/NeMo-Guardrails/tree/develop/docs/user_guides/jailbreak_detection_heuristics). 

Last but not least, if enough time, we should also evaluate the `llama2_70b` and `playground_mixtral_8x7b` models. See #387 for more context.

docs/user_guides/llm/nvidia_ai_endpoints/README.md
```diff
@@ -0,0 +1,97 @@
...
+## Registering ChatNVIDIA as a New Provider
+
+To change the engine used by the bot, you need to register the new provider within the NeMo Guardrails framework. This is done using the `register_llm_provider` function, which maps the `nvidia_ai_endpoints` identifier to the `ChatNVIDIA` class. As a result, the ABC Bot can call upon `ChatNVIDIA` when it needs to interact with NVIDIA AI Foundation Endpoints and access the models.
+
+```python
+from nemoguardrails import LLMRails, RailsConfig
+from nemoguardrails.llm.providers import register_llm_provider
+from langchain_nvidia_ai_endpoints import ChatNVIDIA
+
+register_llm_provider("nvidia_ai_endpoints", ChatNVIDIA)
```
drazvan: This should be registered by default, the user should not have to do this. 
See: https://github.com/NVIDIA/NeMo-Guardrails/pull/383 for how support for Generative AI Hub is added.

docs/user_guides/llm/nvidia_ai_endpoints/README.md
```diff
@@ -0,0 +1,97 @@
...
+## Testing the Bot
+
+Start an interactive chat session with the bot using the following command, ensuring the correct path to your configuration file is specified:
+
+```python
+response = rails.generate(messages=[{
+    "role": "context",
+    "content": {
+        "relevant_chunks": """
+            Employees are eligible for the following time off:
```
drazvan: The ABC bot has the knowledge base configured, so providing explicit relevant chunks should not be needed. It should work without.

drazvan (changes requested):
We should also add the `config` folder for the user guide to `.gitignore` so that it does not get added. I know that for other configurations we've let it in, but since the ABC bot is a bigger one, we don't want to duplicate.

docs/user_guides/llm/nvidia_ai_endpoints/README.md
patriciapampanelli: Done here: https://github.com/NVIDIA/NeMo-Guardrails/pull/403

drazvan:
Thanks @patriciapampanelli ! Looks good. I've made a few tweaks to make this similar to the Vertex AI notebook. We'll do the evaluation in a future PR.

PR #406, opened 2024-03-17
Check that output exists when attempting to parse it

serhatgktp:
- We don't need to attempt to parse the output if it doesn't exist
- Additionally, changing the loop conditional in this way suggests to the reader that `lines` is being shortened in each iteration
- Finally, it might make the code safer to avoid using another `while True` loop


drazvan:
We need to test this, but I think this would allow an empty body for the generated flow. It may happen that the LLM just doesn't generate anything. I'm not sure if this will have undesired consequences.

serhatgktp:
I will close this pull request for now as I'm slightly busy and won't be able to spend time on this until later. I will reopen the PR if I'm able to come back to this.

PR #413, opened 2024-03-21
Typo fix: Comment in llm_flows.co

habanoz:
I noticed a small typo in llm_flows.co file. It is not a big deal but better be fixed. 

Only a comment section is altered, thus this can be a quick merge.


drazvan:
Thanks @habanoz ! 👍

Issue #451, opened 2024-04-13
Use default `config_id` for completions endpoint


rgstephens:
It would be  nice if the completions endpoint was more compatible with the OpenAI endpoint so that tools that use this endpoint would work with guardrails without modification. I'd like to use the [deep-chat](https://github.com/OvidijusParsiunas/deep-chat) widget for example.

My proposal is that if neither `config_ids` nor `config_id` are provided on the completions call, the value would be set to `["config"]` as the default.

drazvan:
Indeed. We could even use "model" as the configuration id. Thanks for the contribution!

rgstephens:
For my use case, the "model" value need to be empty.

PR #452, opened 2024-04-13
Use "config" as the default config_id for completions endpoint

rgstephens:
This PR is in response to issue #451


drazvan:
Thanks @rgstephens! Looks like there was a duplication of effort here. We have a slightly more generic solution which was merged as part of https://github.com/NVIDIA/NeMo-Guardrails/pull/538. Closing this one.

PR #462, opened 2024-04-19
Added support for calling embedding models from langchain-nvidia-ai-endpoints

aishwaryap:
Added support for calling embedding models from `langchain-nvidia-ai-endpoints` and included user guide for this.
Added embedding model and engine information to internal events logged to verify model used for creating embeddings.


drazvan:
Changes merged as part of #617 .

PR #463, opened 2024-04-19
Do not store pip cache during docker build

don-attilio:
Hi, thanks for developing the guardrails.
Removing the pip cache in a layer subsequent to the pip install does not reduce image size because the cache stays in the final image. It is more effective and cleaner not using any caching at all.

Adding the option `--no-cache-dir` during pip command reduces the final artifact size for each of the Dockerfiles in the repo, like listed below.

1. `Dockerfile`: from 1.68GB to 1.58GB;
2. `nemoguardrails/library/jailbreak_detection/Dockerfile`: from 12.3GB to 9.37GB;
3. `nemoguardrails/library/jailbreak_detection/Dockerfile-GPU`: from 12.3GB to 9.41GB;
4. `nemoguardrails/library/factchecking/align_score/Dockerfile`: from 9.6GB to 7.6GB;

For the last three dockerfiles, I reordered the COPY and RUN directives in order to avoid having to download the dependencies every time a single line of code changes.
There could be definitely be more optimizations, such as multi-layer builds.

### DIFF
--- a/nemoguardrails/library/jailbreak_detection/Dockerfile-GPU
+++ b/nemoguardrails/library/jailbreak_detection/Dockerfile-GPU
@@ -5,25 +5,24 @@ FROM nvidia/cuda:12.3.1-base-ubuntu20.04
 RUN apt-get update && apt-get install -y git && apt-get install gcc -y && apt-get install g++ -y && apt-get clean
 
 # Install python
-RUN apt-get install python3 python3-pip python3-dev -y
-
-# Upgrade pip
-RUN pip install --upgrade pip
-
-# Copy the source code
-COPY . /app
+RUN apt-get install python3 python3-pip python3-dev -y && apt-get clean
 
 # Set working directory
 WORKDIR /app
 
-# Install the minimal set of requirements for jailbreak detection Server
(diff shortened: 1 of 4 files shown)


nemoguardrails/library/jailbreak_detection/Dockerfile-GPU
```diff
@@ -2,29 +2,31 @@
...
 
+# Copy the source code
+COPY requirements.txt .
+
 # Install the minimal set of requirements for jailbreak detection Server
-RUN pip install -r requirements.txt
+RUN pip install --no-cache-dir -r requirements.txt
+
+# Predownload the GPT2 model.
+RUN python3 -c "from transformers import GPT2LMHeadModel, GPT2TokenizerFast; GPT2LMHeadModel.from_pretrained('gpt2-large'); GPT2TokenizerFast.from_pretrained('gpt2-large');"
```
don-attilio, line 23: Here I had to change `python` to `python3`, as the image could not find the executable for the former.

drazvan:
Thanks @don-attilio!
👍

drazvan (approved):

PR #465, opened 2024-04-22
Fixes empty response

saradiazdelser:
Nemo returned an empty response when concatenating multiple messages as input. This is fixed by not generating a UserMessage event for each of them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,14 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Fixed
+
+* [#465](https://github.com/NVIDIA/NeMo-Guardrails/pull/465) Fix empty response for multiple messages  by @saradiazdelser.
 
 ## [0.8.3] - 2024-04-18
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,15 @@
...
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+* [#465](https://github.com/NVIDIA/NeMo-Guardrails/pull/465) Fix empty response for multiple messages  by @saradiazdelser.
```
drazvan, line 11: Using `-` for unordered lists is a valid syntax. You should use the existing style rather than change the whole file.

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -448,14 +448,14 @@ def _get_events_for_messages(self, messages: List[dict], state: Any):
...
-                            }
-                        )
+                    # # If it's not the last message, we also need to add the `UserMessage` event
+                    # if idx != len(messages) - 1:
+                    #     events.append(
+                    #         {
+                    #             "type": "UserMessage",
+                    #             "text": msg["content"],
+                    #         }
+                    #     )
```
drazvan, line 458: Can you add a test where it breaks without the fix? The `UserMessage` event is part of the normal flow and I don't see an obvious reason why this should be commented. Maybe there's a different root cause for the behaviour you are seeing. 

As a quick check, the test here: https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/tests/test_issue_385.py works correctly. Even though the test does not check explicitly, I've just run it manually and it correctly returns the right response.

drazvan (changes requested):
Thanks @saradiazdelser. We need to understand better if this is the right solution.

drazvan:
The root cause was fixed here: https://github.com/NVIDIA/NeMo-Guardrails/commit/5cc434aa6a1353c03d4d42f87547a810fa9812cc as part of #539.

PR #486, opened 2024-05-08
Fix select model type and custom prompts task.py

cyun9601:
Fixed an issue where the LLM model and custom prompts could not distinguish between main and other tasks.


drazvan (approved):
Thanks for this fix @cyun9601! :+1:

PR #487, opened 2024-05-08
Fix custom prompts configuration manual.md

cyun9601:
If you make custom prompt work as per #486, you should write the output parser. 

Additionally, it seems that need to list all the items available as config for prompts in this user guide file.


cyun9601:
Now that I think about it, Nemo can put an output parser in the prompt as well

drazvan (approved):
Thanks! :+1:

Issue #505, opened 2024-05-19
[BUG] llm_params from GenerationOptions are not considered during generate_user_intent in generation.py when single_call is False


uvnikgupta:
Following is the code for the case when <code>single_call</code> is <code>False</code>
ln 400 <code> generation.generate_user_intent()</code>: 

```python
# We make this call with temperature 0 to have it as deterministic as possible.
with llm_params(llm, temperature=self.config.lowest_temperature):
    result = await llm_call(llm, prompt)
```

Whereas below is the code for the case when <code>single_call</code> is <code>True</code>
ln 1164 to 1170 <code> generation.generate_intent_steps_message()</code>:
```python
generation_options: GenerationOptions = generation_options_var.get()
additional_params = {
	**((generation_options and generation_options.llm_params) or {}),
	"temperature": self.config.lowest_temperature,
}
with llm_params(llm, **additional_params):
	result = await llm_call(llm, prompt)
```
Please fix the <code>generate_user_intent()</code> to take into account the <code>llm_params</code> from <code>GenerationOptions</code> same as being done in <code>generate_intent_steps_message()</code>

drazvan:
Thanks for reporting this, @uvnikgupta! This looks like an easy fix if you want to contribute. Otherwise, we'll try to fix this for 0.10.0.

Shridhar2205:
Hi, I am trying to use Nemoguardrails for a very basic example (https://github.com/NVIDIA/NeMo-Guardrails/tree/a45b71af7d1777e6cf6fff70f82e1dafc0a71872/examples/bots/hello_world). However, I'm getting the following error,

WARNING:nemoguardrails.actions.action_dispatcher:Error while execution generate_user_intent: Unsupported data type
WARNING  Error while execution generate_user_intent: Unsupported data type

Is this error related to this bug, or is it due to other issues with AzureOpenAI+NemoGuardrails (https://github.com/NVIDIA/NeMo-Guardrails/issues/72 & https://github.com/NVIDIA/NeMo-Guardrails/issues/72) ?

My config.yml file looks like this:
models:
  - type: main
    engine: azure
    model: xxxxxxxxxx
    parameters:
      azure_endpoint: xxxxxxxxxx
      api_version: 2023-09-15-preview
      deployment_name: xxxxxxxxxx
      api_key: xxxxxxxxxx
      temperature: 0.7

PR #510, opened 2024-05-20
Fix for issue #505

uvnikgupta:
llm_params from GenerationOptions are not considered during for the following operations:
 generate_user_intent, 
generate_next_step, 
generate_value

### DIFF
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -397,7 +397,10 @@ class LLMGenerationActions:
             llm_call_info_var.set(LLMCallInfo(task=Task.GENERATE_USER_INTENT.value))
 
             # We make this call with temperature 0 to have it as deterministic as possible.
-            with llm_params(llm, temperature=self.config.lowest_temperature):
+            generation_options: GenerationOptions = generation_options_var.get()
+            with llm_params(
+                llm, **((generation_options and generation_options.llm_params) or {})
+            ):
                 result = await llm_call(llm, prompt)
 
             # Parse the output using the associated parser
@@ -565,7 +568,10 @@ class LLMGenerationActions:
             llm_call_info_var.set(LLMCallInfo(task=Task.GENERATE_NEXT_STEPS.value))
 
             # We use temperature 0 for next step prediction as well
-            with llm_params(llm, temperature=self.config.lowest_temperature):
+            generation_options: GenerationOptions = generation_options_var.get()
+            with llm_params(
+                llm, **((generation_options and generation_options.llm_params) or {})
+            ):
                 result = await llm_call(llm, prompt)
 
             # Parse the output using the associated parser
@@ -961,7 +967,10 @@ class LLMGenerationActions:
         # Initialize the LLMCallInfo object
         llm_call_info_var.set(LLMCallInfo(task=Task.GENERATE_VALUE.value))
 
-        with llm_params(llm, temperature=self.config.lowest_temperature):
+        generation_options: GenerationOptions = generation_options_var.get()
+        with llm_params(
+            llm, **((generation_options and generation_options.llm_params) or {})
+        ):
             result = await llm_call(llm, prompt)
 
         # Parse the output using the associated parser
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -398,9 +398,11 @@ class LLMGenerationActions:
 
             # We make this call with temperature 0 to have it as deterministic as possible.
             generation_options: GenerationOptions = generation_options_var.get()
-            with llm_params(
-                llm, **((generation_options and generation_options.llm_params) or {})
-            ):
+            additional_params = {
+                **((generation_options and generation_options.llm_params) or {}),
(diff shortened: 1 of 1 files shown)


nemoguardrails/actions/llm/generation.py
```diff
@@ -397,7 +397,10 @@ async def generate_user_intent(
             llm_call_info_var.set(LLMCallInfo(task=Task.GENERATE_USER_INTENT.value))
 
             # We make this call with temperature 0 to have it as deterministic as possible.
-            with llm_params(llm, temperature=self.config.lowest_temperature):
+            generation_options: GenerationOptions = generation_options_var.get()
+            with llm_params(
+                llm, **((generation_options and generation_options.llm_params) or {})
```
drazvan: The `generation_options.llm_params` should not be used here. The `llm_params` are meant for the "generation LLM call", i.e., the one that produces the result that the user ultimately sees. This could, for example, set the maximum number of tokens. For the other "internal" calls, only certain parameters should be set, e.g., temperature 0.

nemoguardrails/actions/llm/generation.py
```diff
@@ -565,7 +568,10 @@ async def generate_next_step(
             llm_call_info_var.set(LLMCallInfo(task=Task.GENERATE_NEXT_STEPS.value))
 
             # We use temperature 0 for next step prediction as well
-            with llm_params(llm, temperature=self.config.lowest_temperature):
+            generation_options: GenerationOptions = generation_options_var.get()
+            with llm_params(
+                llm, **((generation_options and generation_options.llm_params) or {})
+            ):
```
drazvan: Same as above.

nemoguardrails/actions/llm/generation.py
```diff
@@ -961,7 +967,10 @@ async def generate_value(
         # Initialize the LLMCallInfo object
         llm_call_info_var.set(LLMCallInfo(task=Task.GENERATE_VALUE.value))
 
-        with llm_params(llm, temperature=self.config.lowest_temperature):
+        generation_options: GenerationOptions = generation_options_var.get()
+        with llm_params(
+            llm, **((generation_options and generation_options.llm_params) or {})
+        ):
```
drazvan, line 977: Same as above.

nemoguardrails/actions/llm/generation.py
```diff
@@ -397,7 +397,12 @@ async def generate_user_intent(
             llm_call_info_var.set(LLMCallInfo(task=Task.GENERATE_USER_INTENT.value))
 
             # We make this call with temperature 0 to have it as deterministic as possible.
-            with llm_params(llm, temperature=self.config.lowest_temperature):
+            generation_options: GenerationOptions = generation_options_var.get()
+            additional_params = {
+                **((generation_options and generation_options.llm_params) or {}),
+                "temperature": self.config.lowest_temperature,
+            }
```
drazvan, line 404: To remove `llm_params`.

nemoguardrails/actions/llm/generation.py
```diff
@@ -565,7 +570,12 @@ async def generate_next_step(
             llm_call_info_var.set(LLMCallInfo(task=Task.GENERATE_NEXT_STEPS.value))
 
             # We use temperature 0 for next step prediction as well
-            with llm_params(llm, temperature=self.config.lowest_temperature):
+            generation_options: GenerationOptions = generation_options_var.get()
+            additional_params = {
+                **((generation_options and generation_options.llm_params) or {}),
+                "temperature": self.config.lowest_temperature,
+            }
```
drazvan, line 577: To remove `llm_params`.

nemoguardrails/actions/llm/generation.py
```diff
@@ -1126,10 +1139,16 @@ async def generate_intent_steps_message(
                 _streaming_handler = StreamingHandler()
                 local_streaming_handlers[_streaming_handler.uid] = _streaming_handler
 
+                generation_options: GenerationOptions = generation_options_var.get()
+                additional_params = {
+                    **((generation_options and generation_options.llm_params) or {}),
+                    "temperature": self.config.lowest_temperature,
+                }
+
```
drazvan, line 1147: Here it's ok to use the `llm_params` because everything is happening in single call.

drazvan (changes requested):
Thanks for this @uvnikgupta! 
Looking closer into this, the `llm_params` from the generation options should actually not be used for the `generate_user_intent`, `generate_next_steps`, `generate_value` tasks. But we need the fix for `generate_intent_next_steps`.

nemoguardrails/actions/llm/generation.py
drazvan: O a second thought, this is tricky too. For example, if the original prompt has a `max_tokens` set to 1, let's say we want to LLM to respond with yes/no, then this would break the logic as the single LLM call mode needs to alter the prompt and generate more tokens. I don't have a solution for this.

drazvan:
Closing this for now. The right approach for when to pass the original params to "sub-tasks" needs more thinking and design. Thanks @uvnikgupta for trying to sort this. We'll move this to the backlog for now.

PR #517, opened 2024-05-25
docs: add config for using models with ollama

vedantnaik19:
Provides an example for using open-source LLMs with [Ollama](https://ollama.com/).


vedantnaik19:
Creating a fresh one.

PR #518, opened 2024-05-25
docs: add example config for using models with ollama

vedantnaik19:
Provides an example for using open-source LLMs with [Ollama](https://ollama.com/).


drazvan (approved):
Thanks @vedantnaik19!

PR #525, opened 2024-05-30
Fix/string expression double braces

schuellc-nvidia:
Fix a bug where double braces did not get converted to single braces in strings without an expression.


drazvan (approved):
Looks good! 👍

PR #531, opened 2024-06-03
Fix Colang 2 flow activation

schuellc-nvidia:
Fixes the flow activation mechanism such that flows can only be activated once and remain being activated until all activations have ended.
Related changes:
* Fix of activation expansion
* Removal of unnecessary context attributes
* Adapting multimodal demo to new activation mechanism


drazvan (approved):
I've not yet reviewed in detail, but merging to unblock.

PR #534, opened 2024-06-03
Feature/llm request steps reduction

schuellc-nvidia:
Introduces new llm flow and action that can generate a bot intent and bot action from a user action with one single llm request.


drazvan (approved):
Nice. This the equivalent of the "single_llm_call" mode we have for Colang 1.0. We'll need to bring in the support for streaming as well, but we'll do in a future PR.

PR #538, opened 2024-06-04
Feature/issue 451 default config

### LINKED ISSUE #451: Use default `config_id` for completions endpoint

Pouyanpi:
### Summary

Implemented support for a default configuration in the Guardrails server, enabling automatic fallback to a predefined config when no `config_id` is specified in API request. It resolves #451.

### Changes

- **Default Configuration Support**: Added the capability to use the `--default-config-id` flag to set a default configuration.
- **Documentation and Testing**: Updated documentation and added tests to cover the new `default_config_id` functionality.


drazvan (approved):
Looks good :+1:.

PR #544, opened 2024-06-05
Fix/issue 216 bot utterance (Enable Parser to Handle Multiline Strings and Newlines)

### LINKED ISSUE #216: Allow new line characters inside predefined bot messages
If "\n" is used inside a predefined bot message, it will be returned as is as part of the message. It should be translated to an actual `\n` character.

Pouyanpi:
### Summary

This PR introduces enhancements to the `get_numbered_lines` function to correctly handle multiline strings and newline characters. It resolves #216.

1. Multiline strings are now correctly appended to the `lines` list when they end. This ensures that all lines, including multiline strings, are returned in the final output.

2. The indentation for multiline strings is now calculated based on the first line of the multiline string. This ensures that the indentation is consistent with other lines.

3. The util now differentiate between strings starting with `"` and `"""`. This ensures that multiline strings enclosed in `"""` are not mistakenly treated as single line strings.

4. Implement corresponding end to end test


drazvan (approved):

drazvan:
Looks good! 👍

PR #545, opened 2024-06-05
doc: NVIDIA API Catalog and NIMs

Pouyanpi:
### Summary

This PR enhances the README.md with added instructions for interfacing with the NVIDIA API Catalog and NVIDIA NIMs.

The update confirms that ChatNVIDIA is compatible with local NIM instances.

Please note, the `nvidia_ai_endpoints` remains unchanged as an engine. Please let me know if there is any plan to change it to `nim`.

Thanks


drazvan (approved):
@Pouyanpi, I've aded support for `nim` directly. Also, moved the documentation from the NVIDIA AI Endpoints notebook into the Configuration guide. We should add a notebook for NIM as well, but we can do this later (it's very straightforward based on the docs).

PR #547, opened 2024-06-06
Fix LLMParams bug and add unit tests (fixes #158)

### LINKED ISSUE #158: [Bug] WARNING: Parameter temperature does not exist for Bedrock
Hi,

I noticed that `generation.py` does the following in several places:

```python
with llm_params(llm, temperature=self.config.lowest_temperature):
```

However it seems that for temperature to be passed to bedrock it would need to be:

```python
with llm_params(llm, model_kwargs={'temperature': self.config.lowest_temperature}):
```

I'm not sure of the cleanest way to make it compatible with both, but just wanted to alert you to this issue. Thanks!

Pouyanpi:
### Summary

This PR addresses the bug in the `LLMParams` class. The bug was causing incorrect handling of parameters that were not directly attributes of the `llm` object but were present in the `model_kwargs` dictionary. It resolves #158.

The changes ensure that if a parameter is not an attribute of `llm`, it checks if `llm` has an attribute `model_kwargs`. If `model_kwargs` exists, it checks if the parameter is in `model_kwargs`. If it is, it saves the original value and sets the new value. If it's not, it logs a warning and sets the new value. 

Upon exiting the context manager, the original parameters are restored correctly, whether they were direct attributes of `llm` or keys in `model_kwargs`.

In addition to the bug fix, this PR also enhances the unit tests to cover these cases.

### DIFF
--- a/nemoguardrails/llm/params.py
+++ b/nemoguardrails/llm/params.py
@@ -18,6 +18,7 @@ Module for providing a context manager to temporarily adjust parameters of a lan
 
 Also allows registration of custom parameter managers for different language model types.
 """
+
 import logging
 from typing import Dict, Type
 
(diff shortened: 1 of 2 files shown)


drazvan (approved):
Nice. This was a much needed fix! 👍

PR #548, opened 2024-06-06
Feat/issue 497 implement custom embedding model

### LINKED ISSUE #497: Support for custom embedding model
Please provide support for custom embedding models

Pouyanpi:
### Summary

This PR implements `EmbeddingProviderRegistry` that extends the `singleton` base `Registry` class. It resolves #497.

**Key Changes**:
- Implement base `Registry` class. This class provides a solid foundation for registering and managing different components in NeMo Guardrails.
- Implement `singleton`. This ensures that we only ever have a single instance of the Registry class in our system.
- Implement `EmbeddingProviderRegistry`. It validates models based on whether they have encode and encode_async methods (duck typing).
- Implement `register_embedding_provider`.  This allows users to register their own embedding providers.

**TODO**:

Add documentation


drazvan (approved):
Looks good :+1:. I've added the documentation.

PR #554, opened 2024-06-10
Feature/colang nld intents

schuellc-nvidia:
Add support to use NLDs in Colang 2.0 user intent flows to make them more powerful (works only well with GTP-4o)

### DIFF
--- a/nemoguardrails/actions/v2_x/generation.py
+++ b/nemoguardrails/actions/v2_x/generation.py
@@ -17,7 +17,7 @@
 import logging
 import re
 from ast import literal_eval
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Tuple
 
 from langchain.llms import BaseLLM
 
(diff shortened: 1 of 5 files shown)


nemoguardrails/llm/prompts/general.yml
```diff
@@ -116,19 +116,13 @@ prompts:
...
-
           # This is how a conversation between a user and the bot can go:
           {{ sample_conversation }}
 
           # These are the most likely user intents:
           {{ examples }}
 
-          # This is the current conversation between the user and the bot:
-          {{ history | colang }}
-
```
drazvan, line 131: Are you sure you want to remove the history? I think this will cause issues with contextual responses, e.g. "The first one".

nemoguardrails/actions/v2_x/generation.py
```diff
@@ -182,25 +161,72 @@ async def generate_user_intent(
...
+                if not isinstance(flow_id, str):
+                    continue
+
                 flow_config = state.flow_configs.get(flow_id, None)
-                if isinstance(flow_id, str) and (
-                    flow_config is None
-                    or (
+                element_flow_state_instance = state.flow_id_states[flow_id]
+                if (
+                    isinstance(flow_id, str)
```
drazvan: `flow_id` will always be a string here, given the check a few lines above.

drazvan (approved):
Looks good. We'll need to turn this into a "full feature". This should work also when the `generating user intent for unhandled user utterance` is used. And we'll need to add docs as well. But we'll create a separate PR for that for the next release.

PR #559, opened 2024-06-12
Feature/module flow activation

schuellc-nvidia:
Add module flow activation using the decorator: `@activated`

### DIFF
--- a/nemoguardrails/colang/v2_x/runtime/runtime.py
+++ b/nemoguardrails/colang/v2_x/runtime/runtime.py
@@ -41,6 +41,7 @@ from nemoguardrails.colang.v2_x.runtime.statemachine import (
     initialize_state,
     run_to_completion,
 )
+from nemoguardrails.colang.v2_x.runtime.utils import new_readable_uid
 from nemoguardrails.rails.llm.config import RailsConfig
 from nemoguardrails.utils import new_event_dict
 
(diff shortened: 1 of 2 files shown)


nemoguardrails/colang/v2_x/runtime/runtime.py
```diff
@@ -444,9 +445,32 @@ async def process_events(
...
+            # Start the main flow
             input_event = InternalEvent(name="StartFlow", arguments={"flow_id": "main"})
             input_events.insert(0, input_event)
-            log.info("Start of story!")
+            main_flow_state = state.flow_id_states["main"][-1]
+
+            # Start all module level flows at beginning of main flow
+            idx = 0
+            for flow_config in reversed(state.flow_configs.values()):
+                if "activated" in flow_config.decorators:
```
drazvan: Let's use the shorter form `@active`. I think this would feel more natural when explaining, e.g., "The `@active`  decorator can be used to define an active flow. An active flow doesn't need to be activated."

drazvan (approved):
Looks good!

nemoguardrails/colang/v2_x/runtime/runtime.py
schuellc-nvidia: Yes, makes sense

PR #543, opened 2024-06-05
Integrating AutoAlign's guardrail library with NeMo Guardrails

abhijitpal1247:
Reopened from:

https://github.com/NVIDIA/NeMo-Guardrails/pull/395

resolved signed commit issues

cherry-picked the commit : 0cc75f4

PR #562, opened 2024-06-14
Autoalign factcheck examples

rahm-decypher:
We demonstrate here two ways to configure autoalign fact-checking guardrail so that they can be applied to only eligible llm responses or whenever the user explicitly mentions to apply fact-checking on llm responses that have specific topics:

1.  In examples/configs/autoalign/autoalign_factcheck_config/rails/factcheck.co we show how to use `$check_facts = True` to only fact check content related to Pluto with AutoAlign's fact check guardrail.
2. We introduced a general content verification flag in the fact-checking config to determine whether the llm response is a valid candidate to be fact-checked:
```
"config": {
        "factcheck": {
            "verify_response": true,
            "matching_scores": {
                "score": 0.5
            }
        }
    }
```
The `verify_response` is set to False by default to avoid additional processing time for this additional content classification. When it is set to True, the lls responses are tested with AutoAlign's content validation classifier end point. The relevant descriptions are added in nemoguardrails/library/autoalign/README.md

fyi @drazvan


drazvan:
@rahm-decypher : can you rebase the work for this PR on the branch that was actually merged as part of #543. This seems to be the original one, with unsigned commits. It's not mergeable, and it's hard to spot what's different. Thanks!

drazvan:
Closing as the changes were merged in #566 .

PR #579, opened 2024-06-24
Fix/multiline string expressions

schuellc-nvidia:
Add support for expressions in multiline strings and enable multiline NLD strings using the generator statement ..."<NLD>


drazvan (approved):
Looks good! I've made a small improvement to also allow spaces between `...` and the string.

PR #599, opened 2024-07-04
Feature/ace agent cli compatibilty

schuellc-nvidia:
This makes sure that output events are directly processed again as input events before run to completion is considered done. With this we remove the responsibility of clients like ace-agent to push those events on the event stream and it works better for a request/response pattern interface.


drazvan (approved):
Ok, looks good. Let's keep an eye if this breaks anything moving forward, but as we discussed.

PR #603, opened 2024-07-04
Feature/improve chat cli logging

schuellc-nvidia:
- Improves the NG chat CLI visualization, making it more compact and easier to distinguish between logs (in verbose mode) and the user bot interaction.
- Adds default log statements to the CSL to improve debuggability
- Fixes some related bugs


drazvan (approved):
Looks good. I activated back the spinner (you can't use `print` with `end=""` when the spinner is active, but that was easy to fix). It's a nicer experience with the spinner, especially when you're not in verbose mode.

PR #604, opened 2024-07-04
Fix/tracking user talking state

schuellc-nvidia:
This was already merged into 0.9 but should also be in develop!

### DIFF
--- a/nemoguardrails/colang/v2_x/library/core.co
+++ b/nemoguardrails/colang/v2_x/library/core.co
@@ -175,7 +175,7 @@ flow tracking user talking state
   $user_talking_state = True
   await user said something as $user_said_flow
   $user_talking_state = False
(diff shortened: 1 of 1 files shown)


nemoguardrails/colang/v2_x/library/core.co
```diff
@@ -173,9 +175,9 @@ flow tracking user talking state
     $user_talking_state = False
   await user started saying something
   $user_talking_state = True
-  await user said something
+  await user said something as $user_said_flow
   $user_talking_state = False
-  $last_user_transcript = $event.final_transcript
+  $last_user_transcript = $user_said_flow.transcript
```
drazvan, line 180: In Colang 1.0 we also have a `$last_user_message`. We should have that as well for backwards compatibility. It's added in a separate PR.

drazvan (approved):
:+1:

PR #606, opened 2024-07-05
Feat/colang2 alpha migration script

Pouyanpi:
This PR allows colang2.0-alpha migration. It also refactors the migration.py and adds `from_version` to CLI.

Remaining Tasks:

- [x] Should we save the old version of the co files? It currently does with 'v2-alpha.co' extension. Decision: do not save
- [x] Remove `ccl.co` and `core_flow_library.co`, are there more to add?
- [x] import modules from Colang Standard Library


drazvan:
Closing this as already included in `feature/colang-2.0-library`.

PR #620, opened 2024-07-10
feat(embeddings): add search threshold to BasicEmbeddingsIndex

Pouyanpi:
This commit introduces a search threshold to the `BasicEmbeddingsIndex` class. This threshold is used to filter the search results based on their distances. The `_filter_results` method is added to perform this filtering. The `search` method is updated to use this new feature. The `search_threshold` parameter is also added to the `LLMRails` classs.

TODO:

- [ ] Check if the updated doc is ok

### DIFF
--- a/nemoguardrails/embeddings/basic.py
+++ b/nemoguardrails/embeddings/basic.py
@@ -59,7 +59,7 @@ class BasicEmbeddingsIndex(EmbeddingsIndex):
         embedding_engine=None,
         index=None,
         cache_config: Union[EmbeddingsCacheConfig, Dict[str, Any]] = None,
-        search_threshold: Optional[float] = None,
+        search_threshold: float = float("inf"),
         use_batching: bool = False,
         max_batch_size: int = 10,
         max_batch_hold: float = 0.01,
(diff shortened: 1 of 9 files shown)


nemoguardrails/embeddings/basic.py
```diff
@@ -267,9 +270,32 @@ async def search(self, text: str, max_results: int = 20) -> List[IndexItem]:
...
+    def _filter_results(
+        indices: List[int], distances: List[float], threshold: float
+    ) -> List[int]:
+        if threshold == float("inf"):
+            return indices
+        else:
+            return [
+                index
+                for index, distance in zip(indices, distances)
+                if distance <= threshold
```
drazvan: The `threshold` should be a similarity threshold, with values between 0 and 1. We can convert from the distance provided by Annoy to a similarity threshold using the formula `1 - distance / 2` (the distance is between 0 and 2).

drazvan (changes requested):
Looks good so far. I've added a few more details related to the desired behavior of the feature.

drazvan (approved):
Nice! This was a highly requested feature. We need to provide more guidance on the value for the similarity threshold, but we'll add this in a subsequent PR.

PR #623, opened 2024-07-11
Fix path traversal check

Pouyanpi:
This PR includes three key fixes:

1. Prevents path traversal in the `_get_rails` function by rejecting config_ids that contain dangerous characters or sequences.
2. Removes an unnecessary path check in the `_get_rails` function, simplifies the code and reducing the chance of errors.
3. Modifies the server command in the CLI to correctly expand user paths, allowing for paths starting with '~' to be used.

fixes [4706127]


drazvan (changes requested):
Looks good! Once you fix the signature for the commits, we can merge.

PR #639, opened 2024-07-21
Feature - Ollama Embeddings

kaushikabhishek87:
Hi, thank you for all the work around guardrails, I am using NeMo in my project & it is really awesome. This is my first contribution to the repo, happy to learn. I am using Ollama as LLM with NeMo and also wanted to use Ollama Embeddings for embedding part.

This PR facilitates using Ollama Embeddings as a choice within embeddings models. With this PR users can use below structure with Ollama

models:
  - type: embeddings
    engine: OllamaEmbed
    model: nomic-embed-text
    parameters:
      base_url: http://0.0.0.0:11434


kaushikabhishek87:
Hi can someone review this

kaushikabhishek87:
Hi can anyone trigger the workflow & review

nemoguardrails/embeddings/providers/__init__.py
```diff
@@ -87,7 +92,13 @@ def init_embedding_model(embedding_model: str, embedding_engine: str) -> Embeddi
     model_key = f"{embedding_engine}-{embedding_model}"
 
     if model_key not in _embedding_model_cache:
-        model = EmbeddingProviderRegistry().get(embedding_engine)(embedding_model)
+        embedding_engine = EmbeddingProviderRegistry().get(embedding_engine)
+
+        if "base_url" in inspect.signature(embedding_engine.__init__).parameters:
+            model = embedding_engine(embedding_model, base_url)
```
drazvan: It is safer to pass the `base_url` as a keyword argument, rather than a positional one. Maybe the assumption that it is the second positional parameter will break in the future.

drazvan (changes requested):
Looks good :+1:. A small suggestion. 
You should update the docs here: https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/docs/user_guides/configuration-guide.md#supported-embedding-providers to mention Ollama as well. 

And last but not least, you should rebase on top for `develop` to sign all the commits. 

Thanks!

kaushikabhishek87:
@drazvan please ignore review request, erronously triggered

nemoguardrails/embeddings/providers/__init__.py
kauabh: updated

drazvan:
@kaushikabhishek87 : yes, feel free to create a new one and link these in the comment for reference.

Pouyanpi:
Hi @kaushikabhishek87,

I noticed that you are currently using `base_url` as a parameter in the `__init__` method. While this works for now, I suggest using `**kwargs` to allow for greater flexibility and future proofing.

By using `**kwargs`, we can easily accommodate additional parameters that might be needed for different models without having to modify the method signature each time. This makes the code more maintainable and extensible.

Here's an example of how you can implement this:

```python
class BasicEmbeddingsIndex(EmbeddingsIndex):
    def __init__(
        self,
        embedding_model: str,
        embedding_engine: str,
        search_threshold: float = None,
        cache_config: Union[Dict, EmbeddingsCacheConfig] = None,
        use_batching: bool = False,
        max_batch_size: int = 10,
        max_batch_hold: float = 0.01,
        **kwargs,
    ):
        """Initialize the BasicEmbeddingsIndex."""
        self.embedding_model = embedding_model
        self.embedding_engine = embedding_engine
        self._embedding_size = 0
        self.search_threshold = search_threshold or float("inf")
        self._parameters = kwargs
        if isinstance(cache_config, Dict):
            self._cache_config = EmbeddingsCacheConfig(**cache_config)
        else:
            self._cache_config = cache_config

    def _init_model(self):
        """Initialize the model used for computing the embeddings."""
        self._model = init_embedding_model(
            embedding_model=self.embedding_model,
            embedding_engine=self.embedding_engine,
            **self._parameters,
        )
```
Additionally, we should update the `__init__` method of all `EmbeddingModel` concrete classes to accept `**kwargs`. Here's an example for the `OpenAIEmbeddingModel` class:

```python
class OpenAIEmbeddingModel(EmbeddingModel):
    """Embedding model using OpenAI API.

    Args:
        embedding_model (str): The name of the embedding model.

    Attributes:
        model (str): The name of the embedding model.
        embedding_size (int): The size of the embeddings.

    Methods:
        encode: Encode a list of documents into embeddings.
    """

    engine_name = "openai"

    def __init__(
        self,
        embedding_model: str,
        **kwargs,
    ):
    # do nothing with kwargs 
```
This way, any additional parameters can be passed through `kwargs` and handled appropriately.

Let me know what you think! Please let me know if I'm overlooking or missing something. 

Also if it is something that will take much of your time, we can do it later in a separate PR. Thanks! 

@drazvan what do you think?

PR #643, opened 2024-07-23
Docs/add tab groups

Pouyanpi:
this PR updates the user_guides to include both Colang 1 and 2 syntax using group tabs.

Add tab groups in:
- [x] docs/user_guides/advanced/extract-user-provided-values.md
- [ ] docs/user_guides/advanced/bot-message-instructions.md


Pouyanpi:
@mikemckiernan I think this might be of interest to you. I could not finish it at the time and it is still pending.

PR #666, opened 2024-08-08
Feature/gcp text moderation

kaushikabhishek87:
Added Support for GCP Text Moderation API - https://cloud.google.com/natural-language/docs/moderating-text


kaushikabhishek87:
Hey @drazvan could you please review this, only two things that I am not sure about,
1. Where to add "google-cloud-language>=2.14.0" in pyproject.toml, currently added in dependencies
2. Change-log info, the date should be. after the approval/merge, I haven't added this yet

kaushikabhishek87:
Hi can anyone advice on this when this will considered

kaushikabhishek87:
Hi any help on this

docs/user_guides/llm-support.md
```diff
@@ -20,30 +20,33 @@ Any new LLM available in Guardrails should be evaluated using at least this set
...
+| Patronus Lynx RAG Hallucination detection _(LLM independent)_ | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark: | :heavy_check_mark:   | :heavy_check_mark:   | :heavy_check_mark:   | :heavy_check_mark:   | :heavy_check_mark:                 |
+| GCP Text Moderation _(LLM independent)_                       | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark:        | :heavy_check_mark: | :heavy_check_mark:   | :heavy_check_mark:   | :heavy_check_mark:   | :heavy_check_mark:   | :heavy_check_mark:                 |
 
 Table legend:
-- :heavy_check_mark: - Supported (_The feature is fully supported by the LLM based on our experiments and tests_)
+
+- :heavy*check_mark: - Supported (\_The feature is fully supported by the LLM based on our experiments and tests*)
```
drazvan: I think the replacement here is broken.

pyproject.toml
```diff
@@ -56,6 +56,7 @@ dependencies = [
   "typer>=0.7.0",
   "uvicorn>=0.23",
   "watchdog>=3.0.0",
+  "google-cloud-language>=2.14.0",
```
drazvan, line 59: This should be added as a separate extra, i.e. `pip install nemoguardrails[gcp]`.

tests/test_gcp_text_moderation_input_rail.py
```diff
@@ -0,0 +1,167 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+
+import pytest
+from google.cloud import language_v2
```
drazvan: Try to follow the same pattern as in https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/nemoguardrails/library/sensitive_data_detection/actions.py so that this doesn't fail when the Google libraries are not installed and the rail is not used.

drazvan (changes requested):
Thanks @kaushikabhishek87 and apologies for the delay in reviewing. A few changes needed, but looking good. Also, I noticed that only some of the commits are signed. If you've configured the auto-signing, then you can try rebasing your branch on top of the current develop branch, and they will all get signed automatically. 

We'll update the Changelog when we prepare the release, no need to do it now.

pyproject.toml
kauabh: added as separate dependency

tests/test_gcp_text_moderation_input_rail.py
kauabh: Updated GCP text moderation test with import error handling

PR #673, opened 2024-08-13
Feature/colang 2/deactivate keyword

schuellc-nvidia:
Add support for the `deactivate <FLOW_ID>` keyword that is a shortcut for `StopFlow(flow_id="<FLOW_ID>", deactivate=True)`


schuellc-nvidia:
We also need to add the deactivate keyword to the sphinx parser configuration.

drazvan (approved):
Looks good.

PR #676, opened 2024-08-14
Doc/llm per task

Pouyanpi:
- Adds llm per task section to the configuration guide.
- Updates custom llm model section


drazvan (approved):
Looks good :+1:. Made a few tweaks.

PR #681, opened 2024-08-19
Fix/issue 183 non openai hallucination check

Pouyanpi:
Allow the use of non OpenAI llm models.

### DIFF
--- a/nemoguardrails/library/hallucination/actions.py
+++ b/nemoguardrails/library/hallucination/actions.py
@@ -68,8 +68,8 @@ async def self_check_hallucination(
 
         if "openai" not in str(type(llm)).lower():
             log.warning(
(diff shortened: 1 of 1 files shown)


Pouyanpi:
@drazvan, is it Ok to delegate the check for "temperature" or "n" to the "LLMParams" like this?

drazvan (changes requested):
@pouyanpi: we should do a proactive check if the LLM supports the `n` parameter, similar to https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/nemoguardrails/rails/llm/llmrails.py#L371. Otherwise, this will most likely raise an error.

Pouyanpi:
@drazvan, I made the change, please have a look.

drazvan (approved):
Thanks :+1:.

PR #691, opened 2024-08-22
Fix/output parser deprecation warning

Pouyanpi:
- Add a new root validator to the `RailsConfig` class to check if an output
parser is registered for each task. If not, a deprecation warning is
logged. This change is in preparation for a future version where
registering the 'output_parser' in the prompts.yml for each task will
be mandatory.

- Remove the deprecation warnings related to output parser registration
from the self_check_facts, self_check_input, and self_check_output
actions. The warnings are now handled in the RailsConfig class.

Pending:
- [x] Add tests
- [x] double check the `tasks_requiring_output_parser` with @drazvan


Pouyanpi:
@drazvan we can add more tests to the `test_rails_config.py` in future, or shall I use a different name?

PR #694, opened 2024-08-23
fix: updated the code to pass config and kwargs to passthrough_fn

vpr1995:
also wrapped synchronous runnable into asynchronous so that it stays consistent with all synchronous runnables


vpr1995:
sorry

PR #695, opened 2024-08-23
Fix: pass config and kwargs to passthrough_fn runnable for issue 665

vpr1995:
fix: updated the code to pass config and kwargs to passthrough_fn

also wrapped synchronous runnable into asynchronous so that it stays consistent with all synchronous runnables

[issue link](https://github.com/NVIDIA/NeMo-Guardrails/issues/665)

### DIFF
--- a/nemoguardrails/integrations/langchain/runnable_rails.py
+++ b/nemoguardrails/integrations/langchain/runnable_rails.py
@@ -26,6 +26,7 @@ from langchain_core.runnables.utils import Input, Output
 from langchain_core.tools import Tool
 
 from nemoguardrails import LLMRails, RailsConfig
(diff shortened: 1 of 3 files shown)


vpr1995:
By wrapping the synchronous invoke in an async coroutine, I was able to run the runnables which only synchronous invoke method in their implementation. without this change, those runnable are throwing error

drazvan:
Thanks @vpr1995! This looks good 👍 .
Can you also add a test for this? Something basic, to make sure we don't break this in the future. 
Also, please make sure your commits are signed. Check out the contributing guidelines: https://github.com/NVIDIA/NeMo-Guardrails/blob/main/CONTRIBUTING.md#submitting-your-work.

drazvan (approved):
Thanks!

PR #704, opened 2024-08-27
Fix/output parser validation

Pouyanpi:
This PR fixes a bug in the `check_output_parser_exists` method and its tests. The method was not handling `TaskPrompt` objects correctly, and the tests were not reflecting this functionality. The method and tests have been updated to correctly handle `TaskPrompt` objects.


drazvan (approved):
Looks good :+1:.

PR #705, opened 2024-08-27
Fix/rails exception migration

Pouyanpi:
This PR provides minor fixes to migration tool, add support of exception handling to standard library flows. It migrates all the co files in the standard library. Highlight exceptions in cli.chat and few other enhancements

### DIFF
--- a/nemoguardrails/cli/chat.py
+++ b/nemoguardrails/cli/chat.py
@@ -391,14 +391,15 @@ async def _run_chat_v2_x(rails_app: LLMRails):
                 if action_uid in running_timer_tasks:
                     running_timer_tasks[action_uid].cancel()
                     running_timer_tasks.pop(action_uid)
+            elif event["type"].endswith("Exception"):
+                if event["type"].endswith("Exception"):
+                    console.print("[red]" + f"Event: {event}" + "[/]")
             elif event["type"] == "LocalAsyncCounter":
                 # if verbose.verbose_mode_enabled:
                 #     console.print(Styles.GREY + f"Event: {event}" + "[/]")
                 pass
             else:
-                if event["type"] not in ["LocalAsyncCounter"]:
-                    if not verbose.verbose_mode_enabled:
-                        console.print(f"Event: {event['type']}")
+                console.print(f"Event: {event['type']}")
 
         # TODO: deserialize the output state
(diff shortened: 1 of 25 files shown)


nemoguardrails/cli/chat.py
```diff
@@ -391,14 +391,15 @@ def _process_output():
...
                     running_timer_tasks.pop(action_uid)
+            elif event["type"].endswith("Exception"):
+                if event["type"].endswith("Exception"):
+                    console.print("[red]" + f"Event: {event}" + "[/]")
             elif event["type"] == "LocalAsyncCounter":
                 # if verbose.verbose_mode_enabled:
                 #     console.print(Styles.GREY + f"Event: {event}" + "[/]")
                 pass
             else:
-                if event["type"] not in ["LocalAsyncCounter"]:
```
drazvan, line 399: Nice catch this was not needed.

nemoguardrails/cli/migration.py
```diff
@@ -219,35 +219,37 @@ def convert_colang_1_syntax(lines: List[str]) -> List[str]:
         if line.lstrip().startswith("define "):
             prev_line = None
         # We convert "define flow" to "flow"
-        line = re.sub(r"define flow", "flow", re.sub(r"[-']", " ", line))
```
drazvan, line 222: Why remove the `'` from the regex? If I remember correctly, I've added because it breaks the parsing the the name of a flow ends up having an `'` in it, which could have been the case for Colang 1.0.

nemoguardrails/library/activefence/flows.co
```diff
@@ -14,58 +14,84 @@ Privacy Violation	PII
 """
 
 flow activefence moderation
-    """Guardrail based on the maximum risk score."""
-    $result = await CallActivefenceApiAction
+  """Guardrail based on the maximum risk score."""
+  $result = await call activefence api
```
drazvan: Can you double-check this works as intended? Isn't `call activefence api` interpreted as a flow here?

drazvan (changes requested):
Looks good :+1:. There a couple of comments that should be addressed before mergin. And we'll need more tests, but we can add that in a separate PR.

drazvan (approved):
Thanks! :+1:

PR #708, opened 2024-08-28
Feat/add relevant chunk support to colang 2

Pouyanpi:
This PR adds support for the `RetrieveRelevantChunkAction` within Colang 2.0, ensuring that the necessary context is available for the correct functioning of the migrated standard library.

Depends on #705

### DIFF
--- a/nemoguardrails/colang/v2_x/library/llm.co
+++ b/nemoguardrails/colang/v2_x/library/llm.co
@@ -95,6 +95,9 @@ flow continuation on unhandled user utterance
 
   log 'start generating user intent and bot intent/action...'
   $action = 'user said "{$event.final_transcript}"'
+
+  # NOTE: I am not sure if this is needed here.
+  await RetrieveRelevantChunksAction()
   #await GenerateUserIntentAction(user_action=$action, max_example_flows=20) as $action_ref
   #$user_intent = $action_ref.return_value
   await GenerateUserIntentAndBotAction(user_action=$action, max_example_flows=20) as $action_ref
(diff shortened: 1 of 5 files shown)


nemoguardrails/colang/v2_x/library/llm.co
```diff
@@ -95,6 +95,9 @@ flow continuation on unhandled user utterance
 
   log 'start generating user intent and bot intent/action...'
   $action = 'user said "{$event.final_transcript}"'
+
+  # NOTE: I am not sure if this is needed here.
+  await RetrieveRelevantChunksAction()
```
drazvan, line 102: Yes, this is needed here. The relevant chunks need to be computed before the generation happen in `GenerateUserIntentAndBotAction`.

nemoguardrails/colang/v2_x/library/llm.co
```diff
@@ -210,6 +213,8 @@ flow llm generate interaction continuation flow -> $flow_name
   log "generated flow continuation: `{$flow_info}`"
   $exists = await CheckValidFlowExistsAction(flow_id=$flow_info.name)
 
+  await RetrieveRelevantChunksAction()
```
drazvan: And I think this one needs to be moved above the `await GenerateFlowContinuationAction`

drazvan (changes requested):
@Pouyanpi, we need to add some tests to this as well, to make sure it works and it doesn't break in the future. You can test the right things end up in the prompt with something similar to: https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/tests/test_bug_5.py#L68. 
Thanks!

Pouyanpi:
@drazvan, just made the changes and added a test.

drazvan (approved):
Thanks! Looks good :+1:.

PR #711, opened 2024-08-29
test: add exception handling tests for standard library guardrails

Pouyanpi:
Add tests for rails exceptions for all pre-defined guardrails.

Depends on #705 (must be merged after #705)

### DIFF
--- a/tests/test_guardrail_exceptions.py
+++ b/tests/test_guardrail_exceptions.py
@@ -12,13 +12,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+import re
+from typing import Optional
+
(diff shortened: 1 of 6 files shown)


tests/test_guardrail_exceptions.py
```diff
@@ -30,29 +35,67 @@
...
+    output_prompt_param = (
+        re.sub(r"(?<=\w)\s(?=\w)", "_", output_params) if output_params else None
+    )
+
+    template = Template(
+        """
     models: []
     rails:
-        input:
+        {% if input_params %}input:
```
drazvan, line 55: Nice use of the jinja templates :+1:.

drazvan (commented):
Nice! Can we do both v1 and v2 as well?

drazvan:
@Pouyanpi : let me know if you can add support for v2 as well, and then we can merge.

PR #723, opened 2024-09-02
Fix/cli server command

Pouyanpi:
Move import to server function scope. This prevents errors when the default `examples` or `chat-ui` folders do not exist for any reason and the user intends to use other commands like `chat`, `migrate` or `eval`.

It also improves the FileNotFoundError message


drazvan (approved):
Looks good :+1:

PR #724, opened 2024-09-02
Fix/llm flow continuation generation

schuellc-nvidia:
Fixes various model compatibility prompting related issues.


drazvan (approved):
Looks good. We need to add some sort of "live tests" that we can run locally to check that the prompt changes work ok for various use cases. Otherwise, it's difficult to fully asses the impact.

PR #725, opened 2024-09-03
fix(migration): replace hyphens and apostrophes

Pouyanpi:
This PR replaces hyphens and apostrophes in lines that start with 'bot' or 'user', migrating the abc bot was raising parsing error as the most recent change to this file limited this substitution to the flow definition line.


drazvan (approved):
Looks good :+1:.

PR #732, opened 2024-09-04
Fix: Process all outgoing events

sklinglernv:
Fixes a bug in the Colang 2 runtime, that causes certain output events to be not fed as input events again. This happens if multiple input_events are processed at the same time and all of these events generate output_events.

This PL fixes this by maintaining a separate list of output events that need to be added to the input event list to continue processing.


schuellc-nvidia:
Thanks Severin! Looks good!

PR #733, opened 2024-09-06
feat: Add support for embeddings only with search threshold

Pouyanpi:
This PR adds support for embeddings only with search threshold and fallback intent.


drazvan (changes requested):
I fixed a couple of issues. Let's add a test with a Colang 2.0 config as well to make sure this doesn't break.

drazvan (approved):
Thanks! :+1:

PR #737, opened 2024-09-09
Feat/embedding only support colang 2

Pouyanpi:
- Add test for embeddings only and fallback_intent in Colang 2
- Fix a bug: fall back intent must be used when the embedding only branch does not return any intent.

### DIFF
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -391,7 +391,7 @@ class LLMGenerationActions:
                         events=[new_event_dict("UserIntent", intent=intent)]
                     )
 
-                if (
+                elif (
                     config.rails.dialog.user_messages.embeddings_only
                     and config.rails.dialog.user_messages.embeddings_only_fallback_intent
                 ):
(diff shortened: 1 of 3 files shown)


drazvan:
@Pouyanpi : I think there's a bug in both Colang 1 and 2 implementations. The threshold is meant to be for the top choice confidence. But if none is above that, and I don't have a fallback intent, then no examples are used either (because the search is performed only once with the threshold). Can you double-check? In the fallback with no intent, we should include the top 5 examples regardless of their similarity score.

Pouyanpi:
> In the fallback with no intent, we should include the top 5 examples regardless of their similarity score.
@drazvan, let me clarify my understanding of this:

Scenario: We are using embeddings exclusively with a search threshold and no fallback intent.

* If the search threshold filters out all results, leaving the result set empty, we ignore the threshold and return the top 5 results regardless of their similarity scores.
* However, if there are any results that meet the search threshold (even if fewer than 5), we keep only those results.

drazvan (approved):
Looks good :+1:.

PR #759, opened 2024-09-17
Fix the pretty str function

schuellc-nvidia:


nemoguardrails/logging/simplify_formatter.py
```diff
@@ -21,7 +21,12 @@ class SimplifyFormatter(logging.Formatter):
     """A formatter to simplify the log messages for easy reading."""
 
     def format(self, record):
-        text = super().format(record)
+        # If record is a LogRecord, use the default formatting
+        if isinstance(record, logging.LogRecord):
+            text = super().format(record)
+        else:
+            # Otherwise, assume it's a string
```
sklinglernv: Detail: here we are not assuming it is a string, but converting it to a string, no?

sklinglernv (approved):
Looks good to me. Maybe the code comment can be made clearer (optional for me).

PR #760, opened 2024-09-17
Unify uuid generation functions

schuellc-nvidia:
Unify uuid generation function calls such that the uuid are consistent for multiple runs in debug mode


sklinglernv (approved):
Useful unification of the UUID generation helper functions. Thanks!

PR #764, opened 2024-09-18
Feat/migration tool sample converstaion syntax conversion

Pouyanpi:
This PR implements conversion logic of `sample_conversations` in config.yml as part of the migration tool.

To verify it works:

Do the migration
```sh
nemoguardrails convert --from-version '1.0' "./examples/bots/abc" --validate
```

then 
```sh
nemoguardrails chat --config=./examples/bots/abc
```

query_1: what can you do?
query_2: you are stupid


cparisien (approved):

erickgalinkin (approved):
LGTM!

PR #771, opened 2024-09-23
Fix posture management and include option for idling posture.

sklinglernv:
Colang 2.0 Standard Library Fix, that
* Resolves an issue with the avatar posture stuck in thinking state
* Adds an option for idle posture (when no user is present)

### DIFF
--- a/nemoguardrails/colang/v2_x/library/avatars.co
+++ b/nemoguardrails/colang/v2_x/library/avatars.co
@@ -242,14 +242,23 @@ flow handling bot talking interruption $mode="inform" $stop_pattern="stop|cancel
 # --------------------------------------
 
 @loop("managing_bot_postures")
-flow managing bot postures
-  """Activates all the posture management."""
+flow managing bot postures $with_idling=False
+  """
+  Activates all the posture management. If $with_idling is set to `True` the bot will maintain an idle posture until
+  a user becomes present.
+  """
   activate tracking bot talking state
-  activate managing listening posture
-  activate managing thinking posture
-  activate managing talking posture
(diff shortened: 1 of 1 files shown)


nemoguardrails/colang/v2_x/library/avatars.co
```diff
@@ -274,11 +283,14 @@ flow managing thinking posture
...
     start bot posture "thinking"
-    bot started saying something
-
-@loop("managing_idle_posture")
-flow managing idle posture
-  """Start and stop idle posture."""
-  match (polling llm request response).Started()
-  start bot posture "Thinking, idle" as $posture
-  match (polling llm request response).Finished()
+    if not $bot_talking_state
```
schuellc-nvidia, line 286: This seems to unnecessary since we already check before?

sklinglernv: I think the issue here is that since this in a separate loop you can have the StartPostureBotAction() at the "same time" as a StartUtteranceBotAction() from a different loop. (So basically, the value of $bot_talking_state changes during the line 285 so we check again on 286).

schuellc-nvidia: Good point... That seems to call for an additional Colang feature imo. Something to define a scope where a certain state must hold...

schuellc-nvidia (approved):

nemoguardrails/colang/v2_x/library/avatars.co
sklinglernv: Yes I agree. We should think about that.

PR #745, opened 2024-09-11
feat: add relevant chunks prompts 

Pouyanpi:
Add relevant chunks for `generate_flow_continuation` and `generate_user_intent_and_bot_action_from_user_action` tasks

tested with abc_v2 config (also for meta/llama3-70b-instruct) without kb folder.

```python
from nemoguardrails import LLMRails, RailsConfig

config = RailsConfig.from_path("./examples/bots/abc_v2")
rails = LLMRails(config)

response = rails.generate(messages=[{
    "role": "context",
    "content": {
        "relevant_chunks": """
            Employees are eligible for the following time off:
              * Vacation: 15 days per year, accrued monthly.
              * Sick leave: 25 days per year, accrued monthly.
              * Personal days: 5 days per year, accrued monthly.
              * Paid holidays: New Year's Day, Memorial Day, Independence Day, Thanksgiving Day, Christmas Day.
              * Bereavement leave: 3 days paid leave for immediate family members, 1 day for non-immediate family members. """
    }
},{
    "role": "user",
   
    "content": "How many sick leaves do employees have?"
}])
``` 

response["content"] must contain 25 days.

PR #772, opened 2024-09-24
fix: move relevant chunks before current conversations

Pouyanpi:
Otherwise the bot will generate based on the relevant chunks only.

### DIFF
--- a/nemoguardrails/llm/prompts/openai.yml
+++ b/nemoguardrails/llm/prompts/openai.yml
@@ -238,15 +238,16 @@ prompts:
           # This is how a conversation between a user and the bot can go:
           {{ sample_conversation }}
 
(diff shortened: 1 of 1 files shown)


Pouyanpi:
@drazvan would you have a look? Please try out abc_v2 on develop branch to see the issue. Can it be included in the patch release?

schuellc-nvidia:
I think we need to better test the introduced change in #745.
Changing the prompts like that can result in unexpected behavior. E.g. I would suggest removing the markdown notation. Also, we need to take a look at the prompt templates for the GPT 4 and Llama 3 models.

Pouyanpi:
@schuellc-nvidia, are we good to merge this PR considering #809 ?

schuellc-nvidia (approved):
All good for me

PR #775, opened 2024-09-25
Fix/colang 2/flow continuation prompt

schuellc-nvidia:
* Fixes the assigned prompt role.
* Also increases a unit test threshold since it sporadically still fails.

### DIFF
--- a/nemoguardrails/llm/prompts/openai-chatgpt.yml
+++ b/nemoguardrails/llm/prompts/openai-chatgpt.yml
@@ -222,7 +222,7 @@ prompts:
         content: "This is the current conversation between the user and the bot:"
       - "{{ history | colang | to_messages_v2}}"
 
-      - type: assistant
+      - type: system
         content: |
           Follow these instruction `{{ instructions }}` to generate a value that is assigned to:
           ${{ var_name }} =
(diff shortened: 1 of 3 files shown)


nemoguardrails/llm/prompts/openai-chatgpt.yml
```diff
@@ -222,7 +222,7 @@ prompts:
         content: "This is the current conversation between the user and the bot:"
       - "{{ history | colang | to_messages_v2}}"
 
-      - type: assistant
+      - type: system
```
sklinglernv, line 225: are any other prompt templates (for other models) affected as well? Or they were already updated in the past?

sklinglernv (commented):
Looks good, not sure if other template needs to be updated as well?

nemoguardrails/llm/prompts/openai-chatgpt.yml
schuellc-nvidia: Currently, we only support openai and llama3 models with the new interface. So, these are the only ones so far.

sklinglernv: OK got it.

sklinglernv (approved):
All good from my side.

PR #777, opened 2024-09-26
Remove the user intent from llm library module

schuellc-nvidia:
We need to remove it since otherwise it will be prompted as potential user intent in all intent derivations when llm module is included that leads to unexpected behaviours.

### DIFF
--- a/nemoguardrails/colang/v2_x/library/llm.co
+++ b/nemoguardrails/colang/v2_x/library/llm.co
@@ -322,6 +322,17 @@ flow tracking unhandled user intent state
 # Experimental flows (undocumented)
 # ----------------------------------
 
+# @meta(user_intent=True)
+# flow user requested a task
+#   user said "do something"
+#     or user said "can you do something"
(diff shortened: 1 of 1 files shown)


nemoguardrails/colang/v2_x/library/llm.co
```diff
@@ -322,6 +322,17 @@ flow tracking unhandled user intent state
 # Experimental flows (undocumented)
 # ----------------------------------
 
+# @meta(user_intent=True)
+# flow user requested a task
+#   user said "do something"
+#     or user said "can you do something"
+#     or user said "please do"
+
+flow custom instructions
```
sklinglernv: Do we even need this flow? Where is it used?

sklinglernv (approved):
Makes sense to remove, only question I have if we should completely remove it instead of keeping it commented, but OK to merge from my side.

schuellc-nvidia:
ok, removed it completely

PR #780, opened 2024-09-27
Fix/colang 2/fix undefined flow continuation

schuellc-nvidia:
This fixes an issues to enable designing with undefined user intent flows such that they are awaited an triggered successfully if the flow 'continuation on undefined flow' is activated.


Pouyanpi (approved):
@schuellc-nvidia, LGTM! 

I don't know much about your testing process. But It would be great to have a test that fails with current version which this PR resolves it.

PR #790, opened 2024-10-02
feat: railsignore added for config loading of LLMRails / RailsConfig.

ajanitshimanga:
PR adding support for a .railsignore to ignore .co files in LLMRails / RailsConfig.

For issue: https://github.com/NVIDIA/NeMo-Guardrails/issues/277

Tests added and passing fully.

Two follow ups after this PR are:

1. Looking into add ignoring .py files for ActionDispatcher
2. Discuss if we want .kb files to be specified as a follow up.

### DIFF
--- a/nemoguardrails/rails/llm/config.py
+++ b/nemoguardrails/rails/llm/config.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 """Module for the configuration of rails."""
-
+import fnmatch
 import logging
 import os
 import warnings
@@ -28,6 +28,7 @@ from nemoguardrails.colang import parse_colang_file, parse_flow_elements
 from nemoguardrails.colang.v2_x.lang.colang_ast import Flow
 from nemoguardrails.colang.v2_x.lang.utils import format_colang_parsing_error_message
 from nemoguardrails.colang.v2_x.runtime.errors import ColangParsingError
+from nemoguardrails.utils import get_railsignore_patterns
 
 log = logging.getLogger(__name__)
 
@@ -556,6 +557,12 @@ def _load_path(
             # Followlinks to traverse symlinks instead of ignoring them.
 
             for file in files:
+                # Verify railsignore to skip loading
+                ignored_by_railsignore = _is_file_ignored_by_railsignore(file)
+
+                if ignored_by_railsignore:
+                    continue
+
                 # This is the raw configuration that will be loaded from the file.
                 _raw_config = {}
 
@@ -1203,3 +1210,19 @@ def _generate_rails_flows(flows):
         flow_definitions.insert(1, _LIBRARY_IMPORT + _NEWLINE * 2)
 
     return flow_definitions
+
+
+def _is_file_ignored_by_railsignore(filename: str) -> bool:
+    # Default no skip
+    should_skip_file = False
+
+    # Load candidate patterns from railsignore
+    candidate_patterns = get_railsignore_patterns()
+
+    # Ignore colang, kb, python modules if specified in valid railsignore glob format
+    if filename.endswith(".py") or filename.endswith(".co") or filename.endswith(".kb"):
+        for pattern in candidate_patterns:
+            if fnmatch.fnmatch(filename, pattern):
+                should_skip_file = True
+
+    return should_skip_file
--- a/nemoguardrails/rails/llm/config.py
+++ b/nemoguardrails/rails/llm/config.py
@@ -1213,16 +1213,13 @@ def _generate_rails_flows(flows):
 
 
 def _is_file_ignored_by_railsignore(filename: str) -> bool:
-    # Default no skip
-    should_skip_file = False
+    ignore = False
 
     # Load candidate patterns from railsignore
     candidate_patterns = get_railsignore_patterns()
 
-    # Ignore colang, kb, python modules if specified in valid railsignore glob format
-    if filename.endswith(".py") or filename.endswith(".co") or filename.endswith(".kb"):
-        for pattern in candidate_patterns:
-            if fnmatch.fnmatch(filename, pattern):
-                should_skip_file = True
+    for pattern in candidate_patterns:
+        if fnmatch.fnmatch(filename, pattern):
+            ignore = True
 
-    return should_skip_file
+    return ignore
--- a/nemoguardrails/rails/llm/config.py
+++ b/nemoguardrails/rails/llm/config.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 """Module for the configuration of rails."""
-import fnmatch
+
 import logging
 import os
 import warnings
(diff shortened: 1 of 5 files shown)


ajanitshimanga (commented):
Reviewed by self

Pouyanpi:
Thank you very much @ajanitshimanga for your nice PR. I've skimmed through the changes and it looks good, with minor changes we'll be good to go. I'll add comments in the review later today.


Some nice to have features:

- [ ] aggregate patterns from multiple .railsignore files.
- [ ] Other potential ignore patterns like directory specific patterns (i.e., patterns ending with /), patterns with leading slashes to denote paths relative to the root, etc
- [ ] ignore any file that is matched with the patterns (no need to check the file extension)

nemoguardrails/utils.py
```diff
@@ -302,3 +303,52 @@ def snake_to_camelcase(name: str) -> str:
         str: The converted CamelCase string.
     """
     return "".join(n.capitalize() for n in name.split("_"))
+
+
+def get_railsignore_path() -> Path:
```
Pouyanpi: how about implementing following signature?

```python
def get_all_railsignore_paths(
    full_path: Optional[str] = None, base_dir: Optional[str] = None
) -> List[Path]:
    """Helper to get all .railsignore paths from the current directory up to the root.

    Args:
        full_path (str, optional): The full path to start from.
        base_dir (str, optional): The base directory to consider as the root.

    Returns:
        List[Path]: A list of .railsignore file paths, ordered from root to current directory.
    """
```

nemoguardrails/rails/llm/config.py
```diff
@@ -1203,3 +1210,19 @@ def _generate_rails_flows(flows):
...
+
+def _is_file_ignored_by_railsignore(filename: str) -> bool:
+    # Default no skip
+    should_skip_file = False
+
+    # Load candidate patterns from railsignore
+    candidate_patterns = get_railsignore_patterns()
+
+    # Ignore colang, kb, python modules if specified in valid railsignore glob format
+    if filename.endswith(".py") or filename.endswith(".co") or filename.endswith(".kb"):
```
Pouyanpi: Let's relax this extension check and only stick to the .railsignore

nemoguardrails/rails/llm/config.py
```diff
@@ -1203,3 +1210,19 @@ def _generate_rails_flows(flows):
         flow_definitions.insert(1, _LIBRARY_IMPORT + _NEWLINE * 2)
 
     return flow_definitions
+
+
+def _is_file_ignored_by_railsignore(filename: str) -> bool:
+    # Default no skip
+    should_skip_file = False
```
Pouyanpi: what do you think about renaming `should_skip_file` to `ignore`?

nemoguardrails/utils.py
```diff
@@ -302,3 +303,52 @@ def snake_to_camelcase(name: str) -> str:
...
+
+    # Navigate to the root directory by going up 4 levels
+    root_dir = current_path.parents[1]
+
+    file_path = root_dir / ".railsignore"
+
+    return file_path
+
+
+def get_railsignore_patterns() -> set[str]:
```
Pouyanpi:
```
def get_railsignore_patterns(
    full_path: Optional[str] = None, base_dir: Optional[str] = None
) -> List[str]:
    """ Helper to retrieve all specified patterns in .railsignore files.

    Args:
        full_path (str, optional): The full path to start from.
        base_dir (str, optional): The base directory to consider as the root.

    Returns:
        List[str]: The list of ignore patterns, ordered from root to current directory.
    """
```

nemoguardrails/utils.py
```diff
@@ -302,3 +303,52 @@ def snake_to_camelcase(name: str) -> str:
...
     return "".join(n.capitalize() for n in name.split("_"))
+
+
+def get_railsignore_path() -> Path:
+    """Helper to get railsignore path.
+
+    Returns:
+         Path: The.railsignore file path.
+    """
+    current_path = Path(__file__).resolve()
```
Pouyanpi: it can end up in site-packages once the package is installed from PyPI

nemoguardrails/utils.py
ajanitshimanga: Is the intention here to change the behavior such that there will potentially be multiple .railsignore files scattered across the repo rather than one .railsignore at the root directory?

Or are we simply allowing flexibility for if the .railsignore changed to a separate location and is explicitly specified?

nemoguardrails/rails/llm/config.py
ajanitshimanga: ack, sounds good to me

nemoguardrails/utils.py
ajanitshimanga: Can you elaborate on what is meant here? I don't quite follow

nemoguardrails/rails/llm/config.py
ajanitshimanga: ack, sounds good to me as well

ajanitshimanga: done

nemoguardrails/rails/llm/config.py
ajanitshimanga: done

ajanitshimanga:
> Thank you very much @ajanitshimanga for your nice PR. I've skimmed through the changes and it looks good, with minor changes we'll be good to go. I'll add comments in the review later today.
> 
> Some nice to have features:
> 
> * [ ]  aggregate patterns from multiple .railsignore files.
> * [ ]  Other potential ignore patterns like directory specific patterns (i.e., patterns ending with /), patterns with leading slashes to denote paths relative to the root, etc. Shall we support negation?
> * [X]  ignore any file that is matched with the patterns (no need to check the file extension)

Thanks for the reply, when looking through the nice to have features I wanted to follow up with a few questions to gain alignment:

1. With respect to adding aggregating patterns in multiple .railsignore files, what is the direct benefit here? I may be missing the use case benefit.
2. I think this would be a good follow up to expand patterns but to leave this out of scope for now, does that sound reasonable?
3. This is a good idea, I've relaxed the file extension constraint in a new commit to achieve this.

nemoguardrails/utils.py
Pouyanpi: Yes, multiple .railsignore files, I have a draft that I can push later so that you can update if needed.

nemoguardrails/utils.py
Pouyanpi: I think changing it to something like Path.cwd(). We need to make sure how it behaves once the user pip install the package. Where should they place the ignore file etc.

Pouyanpi:
@ajanitshimanga Thank you once again for you contribution! 

I made some minor changes to your PR and will merge it latest tomorrow. It will be available with 0.11.0 release.

Please have a look and let me know if you think there is any issue.

So mainly the `config.py` was changed to improve the performance, it is expensive to find the ignorefile every time so it is upper in the code. Also, we start from the `config_path` and goes higher to the root to find the railsignore file. The rest are refactoring to respect this change.  

It allows having multiple .railsignore per different config. Thanks and look forward to your future contributions

Pouyanpi (approved):

ajanitshimanga:
Thank you as well @Pouyanpi for the changes you made with refactoring and improving performance.

The refactor looks good to me, I left some comments and I'll put out a quick commit for those minor changes so you can review and merge when you're online.

Cheers! I look forward to further contributions in the future :)

Pouyanpi (approved):
Looks good! Thanks for the changes!

Issue #419, opened 2024-03-21
Context is not passed into the rails.generate function


jackchan0528:
Following the doc here: [https://github.com/NVIDIA/NeMo-Guardrails/tree/develop/docs/getting_started/7_rag](https://github.com/NVIDIA/NeMo-Guardrails/tree/develop/docs/getting_started/7_rag)
```
response = rails.generate(messages=[{
    "role": "context",
    "content": {
        "relevant_chunks": """
            Employees are eligible for the following time off:
              * Vacation: 20 days per year, accrued monthly.
              * Sick leave: 15 days per year, accrued monthly.
              * Personal days: 5 days per year, accrued monthly.
              * Paid holidays: New Year's Day, Memorial Day, Independence Day, Thanksgiving Day, Christmas Day.
              * Bereavement leave: 3 days paid leave for immediate family members, 1 day for non-immediate family members. """
    }
},{
    "role": "user",
    "content": "How many vacation days do I have per year?"
}])
print(response["content"])
```

The correct answer should be 20 days per year. However the model hallucinated and answered:

> You are entitled to 15 vacation days per year based on your current employment contract. These days accrue gradually throughout the year, so you may not have all 15 days available right at the beginning of the year. It's always a good idea to check with your HR department or refer to your employment agreement for specific details on how vacation days are accrued and how they can be used. If you have any specific plans for using your vacation days, I can help you keep track of them and even suggest some travel destinations or activities for your time off. Just let me know how I can assist you further!

I tried lots of different context and question pairs. It seems that the relevant_chunks was not passed into the entire prompt for generating the answer.

drazvan:
@jackchan0528 : did you test with the provided notebook? (I've just tried it again and it works correctly)

jackchan0528:
@drazvan I rerun the notebook and still get the wrong answer as 15 days per year, while the answer should be 20. You may try even simpler test cases and see that the answer generated is irrelevant to the context provided.

```
from nemoguardrails import RailsConfig, LLMRails

config = RailsConfig.from_path("./config")
rails = LLMRails(config)

response = rails.generate(messages=[{
    "role": "user",
    "content": "How many vacation days do I have per year?"
}])
print(response["content"])
```

Output:

> Fetching 7 files: 100%|██████████| 7/7 [00:00<00:00, 76858.97it/s]
> huggingface/tokenizers: The current process just got forked, after parallelism has already been used. Disabling parallelism to avoid deadlocks...
[...]

Issue #575, opened 2024-06-23
passing in relevant_chunks as a message's context


johnyoonh:
[In the Rag example, passing in the relevant_chunks as a message is possible](https://github.com/NVIDIA/NeMo-Guardrails/blob/26304fd0bbf51f2c0c8ca0cb1ed242775f45dfe3/docs/getting_started/7_rag/rag.ipynb#L203)
```
response = rails.generate(messages=[{
    "role": "context",
    "content": {
        "relevant_chunks": """
            Employees are eligible for the following time off:
              * Vacation: 20 days per year, accrued monthly.
              * Sick leave: 15 days per year, accrued monthly.
              * Personal days: 5 days per year, accrued monthly.
              * Paid holidays: New Year's Day, Memorial Day, Independence Day, Thanksgiving Day, Christmas Day.
              * Bereavement leave: 3 days paid leave for immediate family members, 1 day for non-immediate family members. """
    }
}]
```
But when trying this as a JSON input, the `relevant_chunks` will trigger ContextUpdate and make the state update to stop, ending the event chain to be incomplete and input request failing silently with empty response.

Here is my current workaround, where I try to treat context update for  `relevant_chunks` as other types of event, so that `new_state` is created and event chain is continued:
[nemoguardrails/colang/v1_0/runtime/flows.py](https://github.com/NVIDIA/NeMo-Guardrails/blob/26304fd0bbf51f2c0c8ca0cb1ed242775f45dfe3/nemoguardrails/colang/v1_0/runtime/flows.py#L349)
```
    # We don't need to decide any next step on context updates.
    if event["type"] == "ContextUpdate":
        # TODO: add support to also remove keys from the context.
        #  maybe with a special context key e.g. "__remove__": ["key1", "key2"]
        state.context.update({"context": event["data"]})
        if not 'relevant_chunks' in event["data"]:
            state.context.update(event["data"])
            state.context_updates = {}
            state.next_step = None
            return state
```

My proposed solution is to allow for custom role and update the context with consistent name pattern: "user_message", "context_message", "assistant_message", "custom_message",  etc

This would allow me to pass in context specific message like this:
```
[{
    "role": "context",
    "content": "I am context content"
    },
{
    "role": "custom",
    "content": "I am content for custom role"
    }
    ]
```

In an action, I would be able to retrieve them like this: `context_message = context.get("context_message")`

drazvan:
[...]

Issue #561, opened 2024-06-13
RAG Example doesn't work with guardrails, not using the knowledge base.


chuuck:
I have found an issue with the tool when trying to use guardrails with the knowledge base. The system appears to always ignore the knowledge base whenever using the guardrails and generate answer without the use of guardrails. This is what I have tried so far:

- I have used the sample code from [here](https://github.com/NVIDIA/NeMo-Guardrails/tree/e03244cad83aed2d0f34fb69a98e9e044c9876ff/examples/bots/abc)
- I have tried to pass relevant kb through the kb folder saving the kb in a markdown file and through `relevant_chunks`, neither of which had worked. I used the documentation from [here](https://docs.nvidia.com/nemo/guardrails/getting_started/7_rag/README.html#relevant-chunks).
- The only thing I have written as an addition to the sample code is a python script, which can be found lower down.
- I have also tried to just pass kb through the prompt but then jailbreaking guardrail triggers. 

Python script:

```
from nemoguardrails import RailsConfig, LLMRails
import os

os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = "file.json"


config = RailsConfig.from_path("./config")
rails = LLMRails(config)


response = rails.generate(messages=[{
    "role": "user",
    "content": "How many vacation days do I have per year?"
}])
```

Has anyone managed to run Nvidia NeMo guardrails when using RAG application and has anyone experienced similar issues?

drazvan:
@chuuck : can you enable the verbose mode and share the logs? Let's first double check if the relevant chunks end up in the prompt or not. Maybe the LLM doesn't respond correctly, even though the chunks are provided. 

```python
rails = LLMRails(config, verbose=True)
```

shiv248:
I don't know what your code outputs look like or what your project structure/content is to help. But I managed to get things up and running, NeMo Guardrails RAG application, by following this guide [NeMo Guardrails doc 7_RAG]( https://github.com/NVIDIA/NeMo-Guardrails/tree/e03244cad83aed2d0f34fb69a98e9e044c9876ff/docs/getting_started/7_rag).

mbbajra:
I have seen similar issue. Looks like it kb is ignored sometimes. I got this verbose capture as an example. 


Entered verbose mode.
Fetching 5 files: 100%|██████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 48545.19it/s]
[...]

PR #791, opened 2024-10-02
fix: update relevant_chunks retrieval logic

Pouyanpi:
- Added skip_user_message parameter to get_retrieved_relevant_chunks function.
- Updated relevant_chunks assignment to handle cases with no user flow.


Pouyanpi:
resolves #419 , #561, #575

cparisien (approved):

PR #807, opened 2024-10-15
Add try-catch block around rails.generate function for error handling…

ramchennuru:
Hey @Pouyanpi , @drazvan 

… in Readme file

## Description

"Added a try/catch block to handle errors from LLMs in Nemoguard Rails. For example, AzureModels have default security checks, and if a request fails (e.g., due to flagged content like violence), it throws an error, breaking the flow. By wrapping the rails.generate function in a try/catch block, we can prevent flow disruption and handle such errors gracefully."

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,9 +86,12 @@ from nemoguardrails import LLMRails, RailsConfig
 config = RailsConfig.from_path("PATH/TO/CONFIG")
 rails = LLMRails(config)
 
-completion = rails.generate(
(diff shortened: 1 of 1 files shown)


Pouyanpi:
Thank you @ramchennuru for taking time to open this PR! The purpose of this example usage in the README is to guide the user how they can use it. So the code ends there so there is no need to continue from there. If there is any Error it is good to have it explicit so that the user can figure this out. Do you think it is still necessary? Thanks!

PR #809, opened 2024-10-17
Fix/retrieve chunk action newline

Pouyanpi:
## Description

The existence of `\n` in the relevant_chunks was introducing the context in prompts, which interfered with the behavior of CoLang 2.x. This PR adds an optional `is_colang_2` flag to handle context updates differently based on its value. 

This optional argument is necessary as the existence of `\n` was not clear to me at the time of writing this PR.

## Checklist

### DIFF
--- a/nemoguardrails/actions/retrieve_relevant_chunks.py
+++ b/nemoguardrails/actions/retrieve_relevant_chunks.py
@@ -26,6 +26,7 @@ log = logging.getLogger(__name__)
 async def retrieve_relevant_chunks(
     context: Optional[dict] = None,
     kb: Optional[KnowledgeBase] = None,
+    is_colang_2: Optional[bool] = False,
 ):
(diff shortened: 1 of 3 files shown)


nemoguardrails/actions/retrieve_relevant_chunks.py
```diff
@@ -70,7 +71,14 @@ async def retrieve_relevant_chunks(
 
     else:
         # No KB is set up, we keep the existing relevant_chunks if we have them.
-        context_updates["relevant_chunks"] = context.get("relevant_chunks", "") + "\n"
+        if is_colang_2:
```
schuellc-nvidia, line 74: Do we even need this case distinction? Why do we need \n if there are no chunks for Colang 1?

schuellc-nvidia (approved):
Good for me

Issue #813, opened 2024-10-21
feature: Proposing Private AI integration for PII detection


letmerecall:
### Did you check the docs?

- [x] I have read all the NeMo-Guardrails docs

### Is your feature request related to a problem? Please describe.

Add support for Private AI to safeguard Personally Identifiable Information and address one of the big safety concern for using LLMs.

### Describe the solution you'd like

To start with it'd be nice to have a Private AI PII detection action, something like `detect_pii` which can be used in input, output or retrival retrieval.

#### Example usage

```yaml
rails:
  input:
    flows:
      - detect pii on input
  output:
    flows:
      - detect pii on output
  retrieval:
    flows:
      - detect pii on retrieval
```

The implementation can be similar to what we already have for `sensitive_data_detection` action.

The implementation can be later extended to include:
- PII masking
- Synthetic entity generation

### Describe alternatives you've considered

We currently have an action `sensitive_data_detection` that uses Microsoft Presidio.
By introducing Private AI, we will expand support for more entities, improve detection accuracy, and provide additional features, including support for multiple file formats.

The support can also be extended to features like masking, synthetic entity genertion, and different file types.

### Additional resources
- Private AI developer docs: https://docs.private-ai.com/introduction/
- Supported entities: https://docs.private-ai.com/entities/
- Supported file types: https://docs.private-ai.com/files/

letmerecall:
@Pouyanpi, feel free to assign it to me. Excited for the collaboration :)

PR #815, opened 2024-10-22
Add Private AI Integration

letmerecall:
## Description

This PR introduces an integration for [Private AI](https://docs.private-ai.com) for PII detection. It adds a detect_pii action which can be used in input, output and retrieval flow.

## Related Issue(s)

#813 

## Checklist

### DIFF
--- a/nemoguardrails/library/privateai/actions.py
+++ b/nemoguardrails/library/privateai/actions.py
@@ -40,7 +40,11 @@ async def detect_pii(source: str, text: str, config: RailsConfig):
 
     pai_config: PrivateAIDetection = getattr(config.rails.config, "privateai")
 
-    assert source in ["input", "output", "retrieval"], f"Private AI can only be defined in the input, output and retrieval flows. The current flow {source} is not allowed."
+    assert source in [
+        "input",
+        "output",
+        "retrieval",
+    ], f"Private AI can only be defined in the input, output and retrieval flows. The current flow {source} is not allowed."
 
     if len(pai_config.input.entities) == 0:
         return False
@@ -51,10 +55,5 @@ async def detect_pii(source: str, text: str, config: RailsConfig):
         pai_config.server_endpoint,
         pai_config.api_key,
     )
-
-    if entity_detected is None:
-        log.error("Private AI detection API request failed.")
-        # if the request fails, we assume that PII is detected
-        return True
 
     return entity_detected
--- a/nemoguardrails/library/privateai/actions.py
+++ b/nemoguardrails/library/privateai/actions.py
@@ -1,0 +1,60 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""PII detection using Private AI."""
+
+import logging
+
+from nemoguardrails import RailsConfig
+from nemoguardrails.actions import action
+from nemoguardrails.library.privateai.request import private_ai_detection_request
+from nemoguardrails.rails.llm.config import PrivateAIDetection
+
+log = logging.getLogger(__name__)
+
+
+@action(is_system_action=True)
+async def detect_pii(source: str, text: str, config: RailsConfig):
+    """Checks whether the provided text contains any PII.
+
+    Args
+        source: The source for the text, i.e. "input", "output", "retrieval".
+        text: The text to check.
+        config: The rails configuration object.
+
+    Returns
+        True if PII is detected, False otherwise.
+    """
+
+    pai_config: PrivateAIDetection = getattr(config.rails.config, "privateai")
+
+    assert source in ["input", "output", "retrieval"], f"Private AI can only be defined in the input, output and retrieval flows. The current flow {source} is not allowed."
+
+    if len(pai_config.input.entities) == 0:
+        return False
+
+    entity_detected = await private_ai_detection_request(
+        text,
+        getattr(pai_config, source).entities,
+        pai_config.server_endpoint,
+        pai_config.api_key,
+    )
+
+    if entity_detected is None:
+        log.error("Private AI detection API request failed.")
+        # if the request fails, we assume that PII is detected
+        return True
+
+    return entity_detected
--- a/nemoguardrails/library/privateai/actions.py
+++ b/nemoguardrails/library/privateai/actions.py
@@ -46,9 +46,6 @@ async def detect_pii(source: str, text: str, config: RailsConfig):
         "retrieval",
     ], f"Private AI can only be defined in the input, output and retrieval flows. The current flow {source} is not allowed."
 
-    if len(pai_config.input.entities) == 0:
-        return False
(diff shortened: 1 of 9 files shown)


letmerecall:
> Thank you @letmerecall for your PR. I'll review it and update you if needed.
> 
> In the meantime, I suggest you to run following command to convert flows.v1.co to Colang 2.x compatible version.
> 
> ```shell
> mv ./nemoguardrails/library/privateai/flows.v1.co ./nemoguardrails/library/privateai/flows.co
> ```
> 
> Then use the migration tool
> 
> ```shell
> nemoguardrails convert --from-version '1.0' "./nemoguardrails/library/privateai" --validate
> ```
> 
> Then remove the main.co
> 
> ```shell
> rm nemoguardrails/library/privateai/main.co
> ```
> 
> If you think it is beneficial you can create a config in `./examples/config` folder, similar to `patronusai` or `autoalign`.

Thanks @Pouyanpi! Actually I tried that in my fork (https://github.com/privateai/NeMo-Guardrails/pull/4) but was hesitant to include it as I couldn't find a way to test the flow.

Can you point me to some examples to test Colang 2.x flow? Or do you suggest to still include it without any test?

letmerecall:
> Thank you @letmerecall  for adding Colang 2 flows.
> 
> 
> 
> A good example is [abc_v2](https://github.com/NVIDIA/NeMo-Guardrails/tree/develop/examples/bots/abc_v2). 
> 
> 
> 
> If you create a config folder in the examples as I explained above you can use the migration tool to convert that config. 
> 
> 
> 
> for example 
> 
> 
> 
> ```bash
> 
> nemoguardrails conver --from-version '1.0' "./examples/bots/abc"
> 
> ```
> 
> 
> 
> Converts abc to a configuration similar to abc_v2. You can do the same for your example configuration.
> 
> 
> 
> This way you can test it yourself. Please note that currently we don't support retrieval flow in Colang 2 but only input and output rails.
> 
> 
> 
> 

Perfect! I'll give it a shot. Thanks.

examples/configs/privateai/pii_detection/config.yml
```diff
@@ -0,0 +1,27 @@
+models:
+  - type: main
+    engine: openai
+    model: gpt-3.5-turbo-instruct
+
+rails:
+  config:
+    privateai:
+      server_endpoint: https://api.private-ai.com/cloud/v3/process/text
+      api_key: <YOUR_API_KEY>  # get your API key from https://portal.private-ai.com
```
Pouyanpi: Can we remove this? Either defining an environment variable name that is fixed and the user can set. Or allow the user to pass the environment variable name, but I'd suggest the former.

nemoguardrails/library/privateai/actions.py
```diff
@@ -0,0 +1,56 @@
...
+        True if PII is detected, False otherwise.
+    """
+
+    pai_config: PrivateAIDetection = getattr(config.rails.config, "privateai")
+
+    assert source in [
+        "input",
+        "output",
+        "retrieval",
+    ], f"Private AI can only be defined in the input, output and retrieval flows. The current flow {source} is not allowed."
```
Pouyanpi:
```python
if not source in [
    "input",
    "output",
    "retrieval",
]:
   raise ValueError ....

```

It is not suggested to use assert outside of tests as assert conditionals are not guaranteed to be evaluated. Python runtime has a flag -O (for optimize) that ignores assert statements.

nemoguardrails/library/privateai/actions.py
```diff
@@ -0,0 +1,56 @@
...
+
+    entity_detected = await private_ai_detection_request(
+        text,
+        getattr(pai_config, source).entities,
+        pai_config.server_endpoint,
+        pai_config.api_key,
```
Pouyanpi: Read the api_key from env var

nemoguardrails/rails/llm/config.py
```diff
@@ -123,6 +123,40 @@ class SensitiveDataDetection(BaseModel):
...
+
+
+class PrivateAIDetection(BaseModel):
+    """Configuration for Private AI."""
+
+    server_endpoint: str = Field(
+        default="http://localhost:8080/process/text",
+        description="The endpoint for the private AI detection server.",
+    )
+    api_key: Optional[str] = Field(
```
Pouyanpi: Api Key as env var

tests/test_privateai.py
```diff
@@ -0,0 +1,82 @@
...
+
+    chat.app.register_action(mock_detect_pii(True), "detect_pii")
+
+    # This will trigger the input rail
+    chat >> "Hi! I am Mr. John! And my email is test@gmail.com"
+    chat << "I can't answer that."
+
+    # This will trigger only the output one
+    chat >> "Hi!"
+    chat << "I can't answer that."
```
Pouyanpi: - Please add tests when no rails get activated. 

- it is not readily clear which rail is blocking the generation.

- Please add tests for retrieval too, you can define following and register it as action

```python

@action()
def retrieve_relevant_chunks():
    context_updates = {"relevant_chunks": "Mock retrieved context."}

    return ActionResult(
        return_value=context_updates["relevant_chunks"],
        context_updates=context_updates,
    )

```

Have a look at [this test](https://github.com/NVIDIA/NeMo-Guardrails/blob/main/tests/test_patronus_lynx.py).

Pouyanpi (commented):
Thank you @letmerecall, I had a quick look at your PR and it looks good, after applying the suggestions we are good to merge. In the meantime I'll take a closer look at it and give feedbacks if needed 👍

tests/test_privateai.py
letmerecall: > Please add tests for retrieval too, you can define following and register it as action

Testing retrieval, truly, seems a bit tricky in this case. I feel we might need to modify the retrival rails or the [way I'm testing it](https://github.com/privateai/NeMo-Guardrails/blob/feature/private-ai-integration/tests/test_privateai.py#L166).



I was thinking of generating a bot utterance event if `$relevant_chunks` has a PII.

```
define subflow detect pii on retrieval
  """Check if the relevant chunks from the knowledge base have any PII."""
  $has_pii = execute detect_pii(source="retrieval", text=$relevant_chunks)

  if $has_pii
    bot inform answer unknown
    stop
```


But testing it with a mocked `retrieve_relevant_chunks`, make the process go in an infinite loop of `botIntent > .... > InternalSystemActionFinished > botIntent  > ....  > InternalSystemActionFinished` and end up with a `Exception: Too many events.`

If I understand it right, it's an intended behaviour and happening because my retreival flow always (in the test) ends up with a botIntent which in turn invoke a retreival flow.


Let me try to rethink my retrieval flow, in the meantime any help is much appreciated :)

letmerecall: Updated the tests. Let me know what do you think about this.

tests/test_privateai.py
```diff
@@ -0,0 +1,257 @@
...
+            '  "Hi! My name is John as well."',
+        ],
+    )
+
+    chat.app.register_action(retrieve_relevant_chunks, "retrieve_relevant_chunks")
+    chat.app.register_action(mock_detect_pii(True), "detect_pii")
+    chat >> "Hi!"
+    chat << "I can't answer that."
+
+
```
Pouyanpi, line 164:
```suggestion
@pytest.mark.skip(reason="This test needs refinement.")
```

tests/test_privateai.py
```diff
@@ -0,0 +1,257 @@
...
+@pytest.mark.unit
+def test_privateai_pii_detection_retrieval_with_pii():
+    config = RailsConfig.from_content(
```
Pouyanpi, line 169:
```suggestion
    # TODO: @pouyanpi and @letmerecall: Find an alternative approach to test this functionality.
    config = RailsConfig.from_content(
```

Pouyanpi (changes requested):
Thank you @letmerecall for taking time to find a solution. To avoid increasing the load, I suggest we mark to skip this test for now and later we can look for an alternative implementation. Please apply the requested changes then we can merge. Thanks again 👍🏻

Pouyanpi (approved):
Thanks 🚀

PR #817, opened 2024-10-22
fix(pyproject.toml): remove eval from all dependencies

Pouyanpi:
## Description
Setuptools is unable to resolve the dependency graph when eval is included  in the all dependencies list. This issue is primarily caused by a limitation of pip and a conflict with streamlit. Removing eval from the all dependencies list resolves this issue. Eval can be installed separately if needed.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,9 @@ RUN apt-get install -y gcc g++
 # Copy and install NeMo Guardrails
 WORKDIR /nemoguardrails
 COPY . /nemoguardrails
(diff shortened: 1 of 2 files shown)


Pouyanpi:
@drazvan , I am not sure why `eval` is included in the all optional dependencies. Pip fails resolving it, tested with poetry and it works, also if you install all excluding eval, and then eval it works, so I removed it from the all section.

cparisien (approved):

PR #832, opened 2024-10-29
Fix a bug in a code block

schuellc-nvidia:
## Description

Fixes a bug in the Colang documentation

## Checklist

### DIFF
--- a/docs/colang_2/overview.rst
+++ b/docs/colang_2/overview.rst
@@ -89,9 +89,9 @@ You can migrate your Colang 2.0-alpha bots to 2.0-beta using the following comma
 .. code-block:: console
 
     nemoguardrails convert "path/to/2.0-alpha/version/bots" --from-version "2.0-alpha"
-Additionally, you can enable the `--validate` flag to check if the migrated files do not raise any Colang syntax errors:
-.. code-block: console
-    nemoguardrails convert "path/to/2.0-alpha/version/bots" --from-version "2.0-alpha" --validate
+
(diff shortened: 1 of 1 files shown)


docs/colang_2/overview.rst
```diff
@@ -89,9 +89,9 @@ You can migrate your Colang 2.0-alpha bots to 2.0-beta using the following comma
 .. code-block:: console
 
     nemoguardrails convert "path/to/2.0-alpha/version/bots" --from-version "2.0-alpha"
-Additionally, you can enable the `--validate` flag to check if the migrated files do not raise any Colang syntax errors:
-.. code-block: console
```
sklinglernv, line 93: why remove the example?

docs/colang_2/overview.rst
```diff
@@ -89,9 +89,9 @@ You can migrate your Colang 2.0-alpha bots to 2.0-beta using the following comma
...
-    nemoguardrails convert "path/to/2.0-alpha/version/bots" --from-version "2.0-alpha" --validate
+
+Additionally, you can add the ``--validate`` flag to check if the migrated files do not raise any Colang syntax errors.
```
sklinglernv, line 93: Maybe we could add a warning that thiss will migrate the files in-place an overwrite the files. In case someone does not have a backup?

docs/colang_2/overview.rst
sklinglernv: What is the downside of using --validate ? Why not always use it (e.g. just show one proposed way to do it and explaing that users can omit the --validate flag if they don't want to be informed about Syntax errors.

PR #869, opened 2024-11-22
fix(colang): use input rails output for intent generation

niels-garve:
Related to https://github.com/NVIDIA/NeMo-Guardrails/discussions/866

## Description

Hi @drazvan,

I am creating an AI solution based on the interaction loop of Colang v2. To improve the intent detection for user messages like confirmations ("yes", "no", etc.) I'd like to add Query Rewriting; basically an LLM call prompting to rephrase the user message with context from the conversation.

Example:

User: Do you have any bikes?
Bot: Yes, we have a wide range of bikes. Would you like to see them?
User: sure

Now, query rewriting would transform the "sure" into "Show me some bikes!".

## The Problem

The action `generate_user_intent` is called with `user_action` which is the final transcript of the `UtteranceUserActionFinished`. However, the input rails set the `$user_message` and `$last_user_message` through the context.

Here's my debugger:

Bildschirmfoto 2024-11-22 um 14 10 00

The action has a typo: "Sho**e** me bikes", whereas the context has been rewritten to "sho**w** me bikes".

## The solution

Let me first show my debugger again:

Same user input as above with typo: "Shoe me bikes".

Bildschirmfoto 2024-11-22 um 14 11 47

Now, the context is consistent, and the example lookup works on the rewritten query.

## Example and current workaround.

Currently, I am working around this issue by overriding `flow _user_said_something_unexpected` and performing three new steps:
1. match UtteranceUserActionFinished
2. Rewrite the transcript.
3. send UtteranceUserActionFinished with same `action_uid` and `is_success` as before but with updated transcript

Here's the code:

```Colang
@override
flow _user_said_something_unexpected -> $event
    """Given the following conversation and a follow up input, rephrase the follow up input to be a standalone input,
    in its original language. Generate the output in the following format:

    $rephrased_user_message = '''<<the response>>'''

    Conversation:
    {{ history | co_v2 }}

    Question: {{ question }}

    Your Response:"""
    global $user_message
    global $last_user_message
    match UnhandledEvent(event="UtteranceUserActionFinished", loop_ids={$self.loop_id}) as $e
    $question = $e.final_transcript
    ...
    send UtteranceUserActionFinished(
        final_transcript=$rephrased_user_message, action_uid=$e.action_uid, is_success=$e.is_success
    ) as $event
    $user_message = $rephrased_user_message
    $last_user_message = $rephrased_user_message
```

**Note**: Please also refer to the custom prompt in `config.yml`, where I tried to do query rewriting and intent matching in one step, but `examples` would still be wrong without this fix.

**Note**: I'll gladly enhance the README of the query rewriting example once we find a solution.

Happy to hear your feedback!

Cheers,
Niels

## Checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# EditorConfig is awesome: https://editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+
+[*.{co,py}]
+indent_style = space
+indent_size = 4
+
+[*.yml]
+indent_style = space
+indent_size = 2
--- a/examples/v2_x/other/query_rewriting/README.md
+++ b/examples/v2_x/other/query_rewriting/README.md
@@ -1,0 +1,13 @@
+# Query Rewriting
+
+## Run the example
+
+1. Set environment:
+   ```shell
+   export OPENAI_API_KEY=sk-xxx
+    ```
+
+2. Start the NeMo Guardrails chat:
+   ```shell
+   $ python -m nemoguardrails.__main__ chat --streaming --config . --verbose
+   ```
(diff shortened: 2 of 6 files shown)


schuellc-nvidia:
Hi @niels-garve, let me know if [this answer](https://github.com/NVIDIA/NeMo-Guardrails/discussions/866#discussioncomment-11370788) resolves your issue. In general, you should not need to rewrite the user input to get context sensitive user intent detection.

PR #872, opened 2024-11-27
fix Server Guide url in README.md

mattf:
## Checklist


Pouyanpi:
@mattf Thank you for the PR and catching this issue. The problem is that we have not  published the new version of documentation yet, once the new version is published it will be resolved. In the meantime, feel free to open an issue to track it.

PR #884, opened 2024-12-03
chore: pin fastembed to 4.0.0

Pouyanpi:
4.1.0 instroduces rust-pystemmer which does not have any license


Pouyanpi:
also resolves #884

cparisien (approved):

PR #890, opened 2024-12-03
fix activefence rail docs

noamlevy81:
With the current [action](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/nemoguardrails/library/activefence/actions.py#L29) expected input, the docs are out of date.


cparisien (approved):
Thanks @noamlevy81 for catching this.

PR #895, opened 2024-12-04
Update action to support output rail

noamlevy81:
## Description
The current implementation of activefence action allows sending only user messages to activefence API.
This PR allows sending LLM output messages to Activefence API as part ofthe  output rail.

## Checklist

### DIFF
--- a/nemoguardrails/library/activefence/actions.py
+++ b/nemoguardrails/library/activefence/actions.py
@@ -32,12 +32,15 @@ async def call_activefence_api(context: Optional[dict] = None):
     if api_key is None:
         raise ValueError("ACTIVEFENCE_API_KEY environment variable not set.")
 
-    user_message = context.get("user_message")
+    if context.get("triggered_input_rail"):
+        text = context.get("user_message")
+    else:
+        text = context.get("bot_message")
 
     url = "https://apis.activefence.com/sync/v3/content/text"
     headers = {"af-api-key": api_key, "af-source": "nemo-guardrails"}
     data = {
-        "text": user_message,
+        "text": text,
         "content_id": "ng-" + new_uuid(),
     }
 
--- a/nemoguardrails/library/activefence/actions.py
+++ b/nemoguardrails/library/activefence/actions.py
@@ -25,17 +25,12 @@ from nemoguardrails.utils import new_uuid
 log = logging.getLogger(__name__)
 
 
-@action(name="call activefence api", is_system_action=True)
-async def call_activefence_api(context: Optional[dict] = None):
+@action(is_system_action=True)
+async def call_activefence_api(text: Optional[str] = None):
     api_key = os.environ.get("ACTIVEFENCE_API_KEY")
 
     if api_key is None:
         raise ValueError("ACTIVEFENCE_API_KEY environment variable not set.")
-
-    if context.get("triggered_input_rail"):
-        text = context["user_message"]
-    else:
-        text = context["bot_message"]
 
     url = "https://apis.activefence.com/sync/v3/content/text"
     headers = {"af-api-key": api_key, "af-source": "nemo-guardrails"}
--- a/nemoguardrails/library/activefence/actions.py
+++ b/nemoguardrails/library/activefence/actions.py
@@ -33,9 +33,9 @@ async def call_activefence_api(context: Optional[dict] = None):
         raise ValueError("ACTIVEFENCE_API_KEY environment variable not set.")
 
     if context.get("triggered_input_rail"):
-        text = context.get("user_message")
+        text = context["user_message"]
     else:
-        text = context.get("bot_message")
+        text = context["bot_message"]
 
     url = "https://apis.activefence.com/sync/v3/content/text"
     headers = {"af-api-key": api_key, "af-source": "nemo-guardrails"}
--- a/nemoguardrails/library/activefence/flows.co
+++ b/nemoguardrails/library/activefence/flows.co
@@ -13,9 +13,9 @@ Adult Content	General
 Privacy Violation	PII
 """
 
-flow activefence moderation
+define flow activefence moderation on input
   """Guardrail based on the maximum risk score."""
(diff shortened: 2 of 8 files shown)


noamlevy81:
Hi @Pouyanpi, look forward to your input here

Pouyanpi:
@noamlevy81 would you please set me as the reviewer? If it is not possible for you just let me know, thanks!

noamlevy81:
> @noamlevy81 would you please set me as the reviewer? If it is not possible for you just let me know, thanks!

I don't have permissions the that @Pouyanpi

nemoguardrails/library/activefence/actions.py
```diff
@@ -32,12 +32,15 @@ async def call_activefence_api(context: Optional[dict] = None):
     if api_key is None:
         raise ValueError("ACTIVEFENCE_API_KEY environment variable not set.")
 
-    user_message = context.get("user_message")
+    if context.get("triggered_input_rail"):
+        text = context["user_message"]
+    else:
+        text = context["bot_message"]
```
Pouyanpi: Should not use one flow name for both input and output.

`triggered_input_rail` is internal to Colang 1.0 and is not available in Colang 2.0.

as in the subsequent rounds of interaction bot user and bot messages might be available in a context variable (one can argue that this is a bug). So right now, a third argument is necessary if you want to use a unified interface (e.g., `call_activefence_api`  but again you can not use one flow to handle both input and output sources (see [this](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/docs/user-guides/community/privateai.md) doc and its [implementation](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/nemoguardrails/library/privateai/actions.py#L31) where a `source` argument is used for an example). 

For example following does not work when both input and output rails are activated

```python
   
    text = context["bot_message"] or context["user_message"]
```

Note: In Colang 2.0 it is possible to pass the arguments during flow definition:
```
    awaitSelfCheckOutputAction(text=$output_text)
```

would it be possible for you to introduce two flows (e.g., `activefence moderation on input` and `activefence moderation on output`)?

If you can think of a better solution please let me know. Thanks!

Pouyanpi: Forgot to mention, instead of using `source` you can explicitly pass `user_message` or `bot_message` in flow definitions but still it requires you to use two distinct flows per input source.

You might use following signature:

```python
async def call_activefence_api(user_message: Optional[str] = None, bot_message: Optional[str] = None):
```
instead of

```python
async def call_activefence_api(context: Optional[dict] = None):
```

noamlevy81: After giving it another thought it also makes more sense from a product perspective.
I applied the suggested changes, LMK what you think

nemoguardrails/library/activefence/flows.co
```diff
@@ -13,9 +13,9 @@ Adult Content	General
 Privacy Violation	PII
 """
 
-flow activefence moderation
+define flow activefence moderation on input
   """Guardrail based on the maximum risk score."""
-  $result = await  CallActivefenceApiAction()
+  $result = await  CallActivefenceApiAction(text=$input_text)
```
Pouyanpi: same as in Colang 1.

`text=$user_message`

nemoguardrails/library/activefence/flows.co
```diff
@@ -24,9 +24,20 @@ flow activefence moderation
       bot refuse to respond
     abort
 
-flow activefence moderation detailed
+flow activefence moderation on output
+  """Guardrail based on the maximum risk score."""
+  $result = await  CallActivefenceApiAction(text=$input_text)
```
Pouyanpi: same as colang 1.

`text=$bot_message`

nemoguardrails/library/activefence/flows.co
```diff
@@ -24,9 +24,20 @@ flow activefence moderation
...
+    if $system.config.enable_rails_exceptions
+      send ActiveFenceModerationRailException(message="ActiveFence moderation triggered. The maximum risk score exceeded the threshold.")
+    else
+      bot refuse to respond
+    abort
+
+flow activefence moderation on input detailed
   """Guardrail based on individual risk scores."""
-  $result = await  CallActivefenceApiAction()
+  $result = await  CallActivefenceApiAction(text=$input_text)
```
Pouyanpi: same as colang 1.

`text=$user_message`

nemoguardrails/library/activefence/flows.co
```diff
@@ -13,9 +13,9 @@ Adult Content	General
 Privacy Violation	PII
 """
 
-flow activefence moderation
+define flow activefence moderation on input
```
Pouyanpi: It must be `flow` not `define flow`

nemoguardrails/library/activefence/actions.py
Pouyanpi: Awesome! thank you @noamlevy81  👍🏻 Just minor adjustment needs to be made in `flows.co` then we are good to merge

nemoguardrails/library/activefence/flows.co
noamlevy81: Good Catch!

Pouyanpi (approved):
Thank you @noamlevy81, LGTM! 👍🏻

Pouyanpi:
@noamlevy81 Merging is blocked, would you please sign all your commits (some are not verified) per [contributing guidelines](https://github.com/NVIDIA/NeMo-Guardrails/blob/fba6d1f2d81ec3557964906b5c91656c0c3877eb/CONTRIBUTING.md#why-gpg-signatures). Thanks!

noamlevy81:
@Pouyanpi created a new PR to fix the signature issue [here](https://github.com/NVIDIA/NeMo-Guardrails/pull/940)

Pouyanpi:
closed in favor of #940

PR #903, opened 2024-12-13
Add ability to choose LLM prompt template for value generation and flow generation

sklinglernv:
## Description

Adds the ability to select a LLM prompt template for value generations and for flow generations by setting a Jiinja variable `{% set template = "<name of template here>" %}` this allows for greater flexibility and selection LLM prompts based on the intended use case. 

This can be used like this
```colang
import core

flow generate antonym
  """
  {% set template = "generate_antonym" %}
  bot say "lucky"
  """
  ...

flow main
  when user said "bird"
    $test = ..."a random bird name{{% set template = 'repeat' %}}"
    await bot say $test
  or when user said "antonym"
    generate antonym
```

with the following config.yaml

```yaml
colang_version: "2.x"

models:
    - type: main
      engine: openai
      model: gpt-3.5-turbo-instruct

prompts:
    - task: generate_antonym
      models:
          - openai/gpt-3.5-turbo
          - openai/gpt-4
      messages:
          - type: user
            content: |-
                Generate the antonym of the bot expression below. Use the syntax: bot say "<antonym goes here>".
          - type: user
            content: |-
                YOUR TASK:
                {{ flow_nld }}

    - task: repeat
      models:
          - openai/gpt-3.5-turbo
          - openai/gpt-4
      messages:
          - type: system
            content: |
                Your are a value generation bot that needs to generate a value for the ${{ var_name }} variable based on instructions form the user.
                Be very precised and always pick the most suitable variable type (e.g. double quotes for strings). Only generated the value and do not provide any additional response.
          - type: user
            content: |
                {{ instructions }} three times
                Assign the generated value to:
                ${{ var_name }} =

```

## Checklist

### DIFF
--- a/nemoguardrails/llm/taskmanager.py
+++ b/nemoguardrails/llm/taskmanager.py
@@ -109,12 +109,14 @@ class LLMTaskManager:
         template_str: str,
         context: Optional[dict] = None,
         events: Optional[List[dict]] = None,
+        out_variables: Optional[dict] = None,
     ) -> str:
         """Render a template using the provided context information.
 
         :param template_str: The template to render.
         :param context: The context for rendering the prompt.
         :param events: The history of events so far.
+        :param out_variables: If not None the dict will be populated with variables set in the template
         :return: The rendered template.
         :rtype: str.
         """
@@ -152,7 +154,15 @@ class LLMTaskManager:
 
                     render_context[variable] = value
 
-        return template.render(render_context)
+        rendered = template.render(render_context)
+
+        if out_variables is not None:
+            mod = template.module
+            template_vars = {
+                n: getattr(mod, n) for n in dir(mod) if not n.startswith("_")
+            }
+            out_variables.update(template_vars)
+        return rendered
 
     def _render_messages(
         self,
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -25,6 +25,7 @@ from langchain.callbacks.manager import (
     CallbackManagerForLLMRun,
 )
 from langchain_core.language_models.llms import LLM
+from pydantic import Field
 
 from nemoguardrails import LLMRails, RailsConfig
 from nemoguardrails.colang import parse_colang_file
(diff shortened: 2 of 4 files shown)


nemoguardrails/llm/taskmanager.py
```diff
@@ -109,12 +109,14 @@ def _render_string(
...
         context: Optional[dict] = None,
         events: Optional[List[dict]] = None,
+        out_variables: Optional[dict] = None,
     ) -> str:
         """Render a template using the provided context information.
 
         :param template_str: The template to render.
         :param context: The context for rendering the prompt.
         :param events: The history of events so far.
+        :param out_variables: If not None the dict will be populated with variables set in the template
```
Pouyanpi, line 119: I might be wrong, but maybe 
```suggestion
        :param out_variables: If not None, the dict will be populated with the variables that were set in the template after rendering. This allows the caller to access the values of the variables used in the rendered template.

```

tests/v2_x/test_llm_template_selection.py
Pouyanpi, line 1: @sklinglernv what do you think of adding a test case for a non-existing template so that we can test behavior when `out_variables` is missing expected keys in value generation. 

For example something like:

```python
        colang_content="""
        flow main
          match UtteranceUserActionFinished(final_transcript="hi")
          $test = ..."a random bird name{{% set template = 'non_existent_template' %}}"
          await UtteranceBotAction(script=$test)
        """
```

tests/utils.py
Pouyanpi, line 1: we were using `llm_calls` in other tests, using something like

```
    info = chat.app.explain()
    assert info.llm_calls[0].prompt == expected_prompt

```

as in [test_embeddings_only_user_messages.py](https://github.com/NVIDIA/NeMo-Guardrails/blob/72b7a25dfcb461e44d2aefe01ea462c3c4be11aa/tests/test_embeddings_only_user_messages.py#L188) which works for Colang 2.0 and in [test_general_instructions.py](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/tests/test_general_instructions.py)

I think something is not allowing you to use it here, right?

Pouyanpi (commented):
Thank you @sklinglernv, It looks good 👍🏻 , let me know what you think about my suggestions.

PR #910, opened 2024-12-18
Add option to configure SHA256 hash algorithm instead of MD5

### LINKED ISSUE #909: feature: Add FIPS compliant hashing algorithm SHA256
### Did you check the docs?

The library utilizes MD5 for hashing in embeddings cache and knowledge base.

In embeddings cache:
- The library utilizes [two hash key generators](https://github.com/NVIDIA/NeMo-Guardrails/blob/7f087ae98f58d6f4bfe810b036d2cee37e351078/nemoguardrails/embeddings/cache.py#L49C1-L64C61). One uses MD5 and other uses Python built-in hash function, which has utilized the SipHash algorithm since Python 3.4, as introduced in [PEP456](https://peps.python.org/pep-0456/).

For knowledgebase:
- The library [utilizes only MD5](https://github.com/NVIDIA/NeMo-Guardrails/blob/3265b396b1b5c973ff3457f7f48ecef5eab4c089/nemoguardrails/kb/kb.py#L117C1-L128C78) without option to choose an alternative.

Both algorithms (MD5, SipHash) are not approved for use under US federal requirements ([FIPS regulation](https://csrc.nist.gov/pubs/fips/140-2/upd2/final)). To ensure the library is usable by companies that need to remain compliant, adding an additional compliant algorithm like SHA-256 as a configuration option is a small change but would make a difference.

In `nemoguardrails/embeddings/cache.py`, add one more (non-default) implementation of the Key Generator which would allow the usage of SHA-256 as an alternative.

In `KnowledgeBaseConfig` located in `nemoguardrails/rails/llm/config.py`, allow configuring the hash algorithm of choice (which would allow choosing SHA-256 instead of only MD5) and use it in the `KnowledgeBase` class for computing the hash.

Alternative supported algorithms are also a viable option.

In case of `KnowledgeBase` class the most straightforward solution would be just replacing the call to `hashlib.md5` with `hashlib.sha256` but this would not be backward compatibile and might cause recomputation of knowledge base. This is why I proposed to add configuration option, even though it is only used for naming of the cache file.

### Additional context

_No response_

mdambski:
## Description
At the moment, the library by default uses MD5 hashing and does not allow the option to choose an alternative algorithm from the approved list under [FIPS](https://csrc.nist.gov/pubs/fips/140-2/upd2/final) regulation. Some versions of Python, which are FIPS compliant, will raise an error when trying to use the MD5 hashing algorithm (more information on FIPS specific behaviour in [python docs](https://docs.python.org/3/library/hashlib.html#hash-algorithms)).

This change-set preserves the current default (MD5) but allows choosing SHA-256 as an alternative.

## Related Issue(s)

  - Fixes #909 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The changes related to the Colang language and runtime have moved to [CHANGELOG-
 
 ### Added
 
+- **Hashing Configuration**: Added SHA-256 hashing algorithm and an option to configure it ([#910](https://github.com/NVIDIA/NeMo-Guardrails/pull/910)) by @mdambski
 - **Observability**: Add observability support with support for different backends ([#844](https://github.com/NVIDIA/NeMo-Guardrails/pull/844)) by @Pouyanpi
 - **Private AI Integration**: Add Private AI Integration ([#815](https://github.com/NVIDIA/NeMo-Guardrails/pull/815)) by @letmerecall
 - **Patronus Evaluate API Integration**: Patronus Evaluate API Integration ([#834](https://github.com/NVIDIA/NeMo-Guardrails/pull/834)) by @varjoshi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The changes related to the Colang language and runtime have moved to [CHANGELOG-
 
 ### Added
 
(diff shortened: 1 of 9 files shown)


mdambski:
closing as I need to re-create the change with signed commits

Pouyanpi:
@mdambski do you have any plan to re-open your PR? I'd be glad to review it 👍🏻

PR #911, opened 2024-12-19
Add return_value to FinishFlow internal event

schuellc-nvidia:
## Description

Adds the option to define the return value of a flow that is finished using the internal FinishFlow(return_value=X) event.


sklinglernv (approved):
Very nice. Thanks Chrisitian!

PR #912, opened 2024-12-20
Refactor return_value to context_update

schuellc-nvidia:
## Description

Refactored #911 to use a more general context_update argument to change any context variable with a FinishFlow or StopFlow event.


sklinglernv (approved):
Nice! all good from my side. I added another test that checks the usage as part of with statements.

PR #913, opened 2024-12-20
Add utility flow `wait until done`

sklinglernv:
## Description

Adds a `wait until done` utility flow into `core.co` that waits until a flow or action is finished or stopped. This intended to be used whenever the user is interested in whether a flow or action has finished but might miss the Finished() events, e.g. as part of `when` statements. 

## Checklist

### DIFF
--- a/docs/colang-2/examples/test_csl.py
+++ b/docs/colang-2/examples/test_csl.py
@@ -484,6 +484,62 @@ hello
     await compare_interaction_with_test_script(test_script, colang_code)
 
 
+# @pytest.mark.asyncio
+# async def test_it_finished():
+#     colang_code = """
+# # COLANG_START: test_it_finished
+# import core
+
+# flow bot greet
+#   bot say "hello"
+
+# flow test0
+#   user said "hi"
+#   await UtteranceBotAction(script="hello") as $ref
+#   await it finished $ref
+
+# flow test1
+#   user said "hi"
+#   await bot greet as $ref
+#   await it finished $ref
+
+# flow test2
+#   user said "hi"
+#   start bot greet as $ref
+#   send $ref.Stop()
+#   await it finished $ref
+
+# flow main
+#   await test0
+#   bot say "test0 success"
+
+#   start test1 as $ref
+#   match $ref.Finished()
+#   bot say "test1 success"
+
+#   start test2 as $ref
+#   match $ref.Failed()
+#   bot say "test2 success"
+
+# # COLANG_END: test_it_finished
+#     """
+
+#     test_script = """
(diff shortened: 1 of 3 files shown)


nemoguardrails/colang/v2_x/library/core.co
```diff
@@ -291,12 +291,31 @@ flow await_flow_by_name $flow_name $new_flow_instance_uid = None
...
-  if type($ref) != "Action" and type($ref) != "FlowState"
-    log "flow `is done` only supports action and flow references."
+  $type = type($ref)
+  if $type != "Action" and $type != "FlowState":
+    log "flow `it finished` expects a flow or action reference as parameter."
     abort
 
-  if $ref.status.value != "finished" and $ref.status.value != "stopped":
-    match $ref.Finished()
+  if $type == "FlowState" and $ref.status.value == "stopped":
```
sklinglernv, line 302: I think the type check here is not necessary unless you want it for documentation purposes

nemoguardrails/colang/v2_x/library/core.co
```diff
@@ -291,12 +291,31 @@ flow await_flow_by_name $flow_name $new_flow_instance_uid = None
...
+    # We abort this flow since it will never finish
+    abort
+
+  if $ref.status.value != "finished":
+    match $ref.Finished() as $finished_event_ref
+
+flow flow_failed $flow_ref -> $failed_event_ref
```
sklinglernv, line 309: why do we have this only for flows and not the same `it failed`` for both actions and flows, would be more symmetric

docs/colang-2/examples/test_csl.py
```diff
@@ -485,37 +485,58 @@ async def test_wait_indefinitely():
...
-# COLANG_START: test_wait_until_done
+# COLANG_START: test_it_finished
 import core
 
-flow handle welcome
-  user said "hi"
+flow bot greet
   bot say "hello"
 
+flow test0
```
sklinglernv: This is a more technical test, I think we should move this into the the test folder.

I would keep here a human firendly test as this goes into the documentation. I think this one is too complex.

sklinglernv (commented):
I like the naming changes and the split. 

But I think the special treatment of flows is not necessary and we should make the CSL example simpler for users to understand (we can keep the test but move it into our unit test folder)

docs/colang-2/examples/test_csl.py
schuellc-nvidia: Updated it

schuellc-nvidia (approved):

PR #965, opened 2025-02-02
Feat: Unified Output Mapping for Actions

Pouyanpi:
## Unified Output Mapping for Action Responses

This PR introduces a unified mechanism for interpreting the outputs of various actions by adding dedicated mapping functions. These mapping functions determine whether an action’s response should trigger a block (e.g., refuse to respond) based on consistent thresholds and criteria.

### Why?

Output mapping is essential because our actions currently return heterogeneous result types and semantics, making it challenging to consistently determine whether an action’s output should be blocked or not. 

## Key Changes:

- The `@action` decorator has been extended to accept an optional `output_mapping` parameter. This mapping is stored in the action’s metadata and is used to interpret the result of the action.
- mapping functions added to all actions in the guardrails library
- adding kwargs arg to all actions
- adding tests

### DIFF
--- a/nemoguardrails/library/jailbreak_detection/actions.py
+++ b/nemoguardrails/library/jailbreak_detection/actions.py
@@ -43,7 +43,9 @@ log = logging.getLogger(__name__)
 
 @action()
 async def jailbreak_detection_heuristics(
-    llm_task_manager: LLMTaskManager, context: Optional[dict] = None
+    llm_task_manager: LLMTaskManager,
+    context: Optional[dict] = None,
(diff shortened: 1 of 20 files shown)


Pouyanpi:
required by #966

nemoguardrails/library/jailbreak_detection/actions.py
Pouyanpi, line 1: as this is input rails the output mapping is not added, but should be.

cparisien (approved):

PR #966, opened 2025-02-02
Feat: Output Rails Streaming

Pouyanpi:
## Description

Example:

```python

config = RailsConfig.from_path("./examples/configs/abc_streaming")

rails = LLMRails(config)

messages = [{"role": "user", "content": "what can you do"}]

async for chunk in rails.stream_async(messages=messages):
    print(f"CHUNK: {chunk}")

```

or using chat cli

```console

nemoguardrails chat --config="./examples/configs/abc_streaming"

```
Todo
- [x] finalize streaming config parameter names
- [x]  test functioanlity on microservice ✅ 
- [x] https://github.com/NVIDIA/NeMo-Guardrails/pull/976 by @mikemckiernan 
- [x] SSE 
- [x] ~~add config example with Nemoguard NIMs?~~
- [x] test multi output rails with Nemoguard and self check output


Pouyanpi:
requires #965

examples/configs/abc_streaming/config.yml
```diff
@@ -0,0 +1,40 @@
...
+      - self check output
+    streaming:
+      enabled: True
+      window_size: 20
+
+  dialog:
+    single_call:
+      enabled: False
+
+streaming: True
```
mikemckiernan: total nit: These files have been all lower case--AFAICT.  Any chance it's easy to use `true` and `false` instead?

nemoguardrails/actions/actions.py
```diff
@@ -14,21 +14,24 @@
...
     """Decorator to mark a function or class as an action.
 
     Args:
         is_system_action (bool): Flag indicating if the action is a system action.
         name (Optional[str]): The name to associate with the action.
         execute_async: Whether the function should be executed in async mode.
-
+        output_mapping (Optional[Callable[[Any], bool]]): A function to interpret the action's result.
+            It should accept the return value (e.g. the first element of a tuple) and return True if the output
+            should be considered blocked.
```
mikemckiernan: sugg: Avoid "should" because it weakens the message.  Maybe "The function accepts the return value, the first element of a tuple, and returns True if the output violates a guardrail."  Or maybe "...if the output is not safe."

nemoguardrails/actions/llm/generation.py
```diff
@@ -762,6 +762,14 @@ async def generate_bot_message(
 
         streaming_handler = streaming_handler_var.get()
 
+        # when we have 'output rails streaming' enabled
+        # we must disable (skip) the output rails which gets executed on $bot_message
+        # as it is executed separately in llmrails.py
+        # of course, it does not work when passed as context in `run_output_rails_in_streaming`
+        # streming_handler is set when stream_async method is used
```
mikemckiernan: s/streming_handler/streaming_handler/

nemoguardrails/actions/output_mapping.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    if isinstance(result, bool):
+        return not result
+    elif isinstance(result, (int, float)):
+        return result < 0.5
+    else:
+        return False
+
+
+def should_block_output(result: Any, action_func: Any) -> bool:
```
mikemckiernan: Ooof.  Should in software is like "try" in Jedi.

nemoguardrails/rails/llm/buffer.py
```diff
@@ -0,0 +1,119 @@
...
+        # TODO: something causes duplicate whitespaces between tokens, figure out why,
+        # If using `return "".join(new_chunks)` works, then the issue might be elsewhere in the code where the chunks are being generated or processed.
+        # Ensure that the chunks themselves do not contain extra spaces.
+        # WAR: return "".join(new_chunks)
+        return "".join(new_chunks)
+
+
+def get_buffer_strategy(config: OutputRailsStreamingConfig) -> BufferStrategy:
+    # TODO: use a factory function or class
+    # currently we only have SlidingWindo, in future we have a registry
```
mikemckiernan:
```suggestion
    # currently we only have SlidingWindow, in future we have a registry
```

nemoguardrails/rails/llm/buffer.py
```diff
@@ -0,0 +1,119 @@
...
+    return SlidingWindow.from_config(config)
+
+
+def is_blocked(result: Tuple[bool, str]) -> bool:
+    """Check output rials status."""
```
mikemckiernan:
```suggestion
    """Check output rails status."""
```

mikemckiernan (commented):
Unsolicited feedback.  LMK if you'd prefer that I not provide unsolicited feedback.

examples/configs/abc_streaming/config.yml
mikemckiernan: Doh!  I take it back, bools are as you have them.

Pouyanpi:
Thank you @mikemckiernan 👍🏻  just don't review `guardrails/library` changes here, their PR is #965 .

I'll push changes wrt review soon, thx!

nemoguardrails/actions/output_mapping.py
cparisien: I can't promise I'll try, but I'll try to try. ;)

examples/configs/abc_streaming/config.yml
mikemckiernan: Any chance this top-level `streaming` field seems problematic in light of the new `rails.output.streaming.enabled` field?  (Reading...maybe I'll figure it out.)

Pouyanpi: Yes! If you set it to false then streaming is disabled (no matter what we have for output rails streaming). 

Right now, we can enable/disable streaming for all output rails, but in future we might want to make it configurable per output rails so having something global makes sense. What do you think?

nemoguardrails/rails/llm/config.py
```diff
@@ -304,6 +304,19 @@ class InputRails(BaseModel):
     )
 
 
+class OutputRailsStreamingConfig(BaseModel):
+    enabled: bool = Field(
+        default=False, description="Indicates if streaming is enabled."
+    )
+    look_back_size: int = Field(default=5, description="The look back size.")
+    window_size: int = Field(default=10, description="The window size.")
```
cparisien: Elaborate a bit on the description. e.g., how the look back size is how many tokens/characters in total are sent to the rail each time as context. Also give the units -- is this counted in characters?

nemoguardrails/rails/llm/config.py
```diff
@@ -304,6 +304,19 @@ class InputRails(BaseModel):
...
+    stream_first: bool = Field(
+        default=True,
+        description="Prioritizes streaming chunks before applying output rails.",
```
cparisien: Not clear what this means?

nemoguardrails/rails/llm/config.py
```diff
@@ -1206,6 +1224,9 @@ def streaming_supported(self):
         Currently, we don't support streaming if there are output rails.
         """
```
cparisien: Update this, since now we do have streaming with output rails.

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -415,6 +418,7 @@ def _init_llms(self):
                     self.runtime.register_action_param(
                         model_name, getattr(self, model_name)
                     )
+                    # this is used for cotnent safety and topic control
```
cparisien:
```suggestion
                    # this is used for content safety and topic control
```

examples/configs/abc_streaming/config.yml
mikemckiernan: My pref is for a single control and the per-output rail config seemed OK to me.  It's not a deal-breaker to have two knobs to twist, but it doesn't seem like a feature either.

Pouyanpi: I see your point, let's see if I understand it


You think, this is easier? I think it is

```yaml
  output:
    flows:
      - self check output
    streaming:
      window_size: 20

  dialog:
    single_call:
      enabled: False

streaming: True
```

and we can have defaults so that following also works:

```yaml
  output:
    flows:
      - self check output

  dialog:
    single_call:
      enabled: False

streaming: True

```

I can change it to work like this.

@cparisien would this be OK?

Pouyanpi: In future we can have per rails config and `enabled` flag.

cparisien: I'm ok with this.

cparisien (commented):
This is looking good to me. Let's make sure we are clear on the "hierarchy" of the "streaming: true" configs and the default behaviour.
Also I'd like to make sure I have a clear understanding of which rails will support this behaviour now, and which won't.

Pouyanpi:
>  Also I'd like to make sure I have a clear understanding of which rails will support this behaviour now, and which won't.

- All the output rails that are defined in the [library](https://github.com/NVIDIA/NeMo-Guardrails/tree/develop/nemoguardrails/library)
- any user defined rail that uses `@action`

nemoguardrails/rails/llm/config.py
Pouyanpi: stream chunk of tokens first, then apply output rails on them (pros: lower perceived TTFT) 

or

apply the output rails on that chunk and then stream (pros: if the chunk is blocked then we don't stream blocked content)

is it clear?

nemoguardrails/rails/llm/config.py
Pouyanpi: size is always number of tokens (words). 

I need to change the param names and also default value. These were used for testing

`window_size` is number of tokens that we apply the output rails on, so when the streaming starts it corresponds to first #window_size tokens, then in the next iteration(s) we use #look_back_size tokens from the previous iteration(s) but still respect the `window_size`.

Pouyanpi: @trebedea need your expert intuition here, what do you think is a good default here (ignore the defaults I've set)

Pouyanpi: Basically, we have 2 parameters:

1. The size of the token group processed at one time: This defines how many tokens are in each block that the output rails operate on.
2. The amount of previous tokens reused in the next iteration: This defines the carryover or context from the previous block that is blended into the current processing batch.

**decision**:
`chunk_size` instead of `window_size`
`context_size` instead of `look_back_size`


	
`chunk_size` and `context_size` are both clear and abstract, communicating exactly what they do: one determines the size of the token block and the other determines how much of the previouss block is reused.

Pouyanpi:
Some thoughts on default values:

chuk_size: 256
context_size: 64

A chunk of about 200–256 tokens tends to be large enough to encapsulate a meaningful “unit” of generated text. This provides the rails with enough information? At the same time, it isn’t so large that increases TTFT latency of first (if `stream_first: False`) and subsequent iterations.

Summary

- Chunk Size (200 tokens):
  - Pros: Captures a complete, meaningful unit of text for effective rail evaluation.
  - Cons: May introduce slight delay if waiting for the full chunk (when stream_first: False).
- Context Size (50 tokens):
  - Pros: Ensures continuity across chunks with minimal impact on latency.
  - Cons: Too small a context might miss cross-chunk violations, but 25% overlap is generally a good compromise.
  
 
| Input Length | Chunk Size | Context Size | Rails Invocation |
|--------------|------------|--------------|------------------|
| 512          | 256        | 64           | 3                |
| 600          | 256        | 64           | 3                |
| 256          | 256        | 64           | 1                |
| 1024         | 256        | 64           | 5                |
| 1024         | 256        | 32           | 5                |
| 1024         | 256        | 32           | 5                |
| 1024         | 128        | 32           | 11               |
| 512          | 128        | 32           | 5                |

cc @cparisien

Pouyanpi:
- [ ] remove `abc_streaming` example from configs

Pouyanpi:
tested ✅ 

```yaml
 ...
 rails:
  input:
    flows:
      - topic safety check input $model=topic_control
  output:
    flows:
      - self check output
    chunk_size: 10
  dialog:
    single_call:
      enabled: False

streaming: True
```

nemoguardrails/rails/llm/config.py
Pouyanpi: fyr, @mikemckiernan @cparisien

mikemckiernan: I like those two field names--and I'm extremely picky.  FYI: https://nvidia.github.io/NeMo-Guardrails/review/pr-976/user-guides/configuration-guide.html#enabling-streaming-output

Pouyanpi:
in chat cli we have following behavior:

image


for the `stream_async` interface

```python
async for chunk in rails.stream_async(messages=messages):
   print(f"CHUNK:{chunk}")

CHUNK:I
....
CHUNK: Is
CHUNK: there
CHUNK: anything
CHUNK: specific
CHUNK: you
CHUNK: would
CHUNK: like
CHUNK: to
CHUNK: know?
CHUNK:{"event": "ABORT", "data": {"reason": "Blocked by self check output rails."}}
```
The SSE is meant for the microservice. 

what do you think? @cparisien @mikemckiernan

Pouyanpi:
> in chat cli we have following behavior:
> 
> `stream_first: True`
> 
> image
> 
> `stream_first: False`
> 
> image
> 

Here I can remove the output rails name if it is client facing and they should not be aware. Or we can have a `-debug` mode that also includes the output rails name.

mikemckiernan:
> CHUNK:{"event": "ABORT", "data": {"reason": "Blocked by self check output rails."}}
> 
> **The SSE is meant for the microservice. **
> 
> what do you think? @cparisien @mikemckiernan

You read my mind--I meant to ask the question "How does the client know if the partial response was unsafe and how does the client check?"  I have no experience in this area, but the concept of an _event_ has appeal.  (We can cover this offline...but does the client check each chunk if it's a dict or str? If dict, it's bad news?)  No strong feeling on making the rail name optional.  Couldn't the client filter it?  Would they want to log it (maybe logged elsewhere...)?

Pouyanpi:
Nice!

> (We can cover this offline...but does the client check each chunk if it's a dict or str? If dict, it's bad news?)

it is a json str so not a dict but can be converted to one if needed.

> No strong feeling on making the rail name optional. Couldn't the client filter it? Would they want to log it (maybe logged elsewhere...)?

Agreed! We include the name 👍🏻

nemoguardrails/rails/llm/config.py
cparisien: Thanks I'm good with these new names too.

nemoguardrails/rails/llm/config.py
cparisien: Thanks that's clear, I expect we'll learn from early users how they experience it.

cparisien (approved):

PR #974, opened 2025-02-04
chore(deps): exclude spacy 3.8.4 as does not support py39

Pouyanpi:


mikemckiernan (approved):
lgtm, I saw the tests pass on the mirror

PR #977, opened 2025-02-04
ci: add upgrade-deps option to test workflow

Pouyanpi:
This workflow allows us to identify any issues caused by the latest dependencies.

- Update dependencies to the latest version
- Install extra (optional) dependencies
- Run tests

### DIFF
--- a/.github/workflows/latest-deps-tests.yml
+++ b/.github/workflows/latest-deps-tests.yml
@@ -1,0 +1,26 @@
+name: Test with Latest Dependencies
+
+on:
+  push:
+    paths-ignore:
+      - ".github/**"
+    tags:
+      - "v*"
+  schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/latest-deps-tests.yml
```diff
@@ -0,0 +1,26 @@
...
+        include:
+          - os: Ubuntu
+            image: ubuntu-latest
+      fail-fast: false
+    uses: ./.github/workflows/_test.yml
+    with:
+      os: ${{ matrix.os }}
+      image: ${{ matrix.image }}
+      python-version: ${{ matrix.python-version }}
+      upgrade-deps: true
```
mikemckiernan, line 26: Neat.  Is the idea to have the bot determine in advance if updating deps works fine or will be a headache?

mikemckiernan (approved):

PR #982, opened 2025-02-06
fix: fix alignscore dependency resolution in Dockerfile

Pouyanpi:
## Description

Recent version of packages such as spacy is causing issue, this PR ensures the image can be built.


Pouyanpi:
resolves 5094560

cparisien (approved):

PR #988, opened 2025-02-08
Feat: Add SHA-256 hashing option

### LINKED ISSUE #909: feature: Add FIPS compliant hashing algorithm SHA256

mdambski:
## Description
This change set introduces the `sha256` hashing option as an alternative to `md5`. By default, the library will use `md5` unless this hashing algorithm is unavailable (which would effectively break the library for the user).

## Rationale
At the moment, the library uses `md5` hashing by default and does not allow the option to choose an alternative.

In some distributions of Python that are compliant with [FIPS Regulation](https://csrc.nist.gov/pubs/fips/140-2/upd2/final), `md5` will either be missing or will raise an error when used. These concerns are brought up in the [Python docs](https://docs.python.org/3/library/hashlib.html#hash-algorithms) related to hash algorithms.

In such environments, using a library that relies solely on `md5` forces users to either abandon usage of the library or create and maintain their own fork with the needed patches. Both options are far from optimal.

cc: @Pouyanpi 

## Related Issue(s)

  - Fixes #909 

## Checklist


nemoguardrails/hashing.py
Pouyanpi, line 1: let's simplify this:

```python
def compute_hash(text: str) -> str:
    """
    Return the hash of the given text using MD5 if available,
    otherwise use SHA256.

    Args:
        text (str): The input text to hash.

    Returns:
        str: The hexadecimal digest of the hash.
    """
    try:
        # Attempt to use MD5 by doing a dummy cal.
        hashlib.md5(b'')
        hash_func = hashlib.md5
    except (AttributeError, ValueError):
        # MD5 is not available use sha256 instead
        hash_func = hashlib.sha256

    return hash_func(text.encode('utf-8')).hexdigest()
    
```
    
and move it to `nemoguardrails.utils.py`

nemoguardrails/kb/kb.py
```diff
@@ -13,13 +13,13 @@
...
 # limitations under the License.
 
-import hashlib
 import logging
 import os
 from time import time
 from typing import Callable, List, Optional, cast
 
 from nemoguardrails.embeddings.index import EmbeddingsIndex, IndexItem
+from nemoguardrails.hashing import generate_hash
```
Pouyanpi:
```python
from nemoguardrails.utils import compute_hash
```

nemoguardrails/kb/kb.py
```diff
@@ -114,18 +114,16 @@ async def build(self):
...
         hash_prefix = self.config.embedding_search_provider.parameters.get(
             "embedding_engine", ""
         ) + self.config.embedding_search_provider.parameters.get("embedding_model", "")
 
-        md5_hash = hashlib.md5(
-            (hash_prefix + "".join(all_text_items)).encode("utf-8")
-        ).hexdigest()
-        cache_file = os.path.join(CACHE_FOLDER, f"{md5_hash}.ann")
-        embedding_size_file = os.path.join(CACHE_FOLDER, f"{md5_hash}.esize")
+        hash_value = generate_hash(hash_prefix + "".join(all_text_items))
```
Pouyanpi:
```python
hash_value = compute_hash(hash_prefix + "".join(all_text_items))
```

docs/user-guides/advanced/embedding-search-providers.md
Pouyanpi, line 1: I think when we allow the users to configure the caching there is no need to set the default dynamically.

Either we set the default to sha256 or md5. I'd say it is no big deal to change the default to sha256. What do you think?

docs/user-guides/advanced/hashing-settings.md
Pouyanpi, line 1: let's remove this

nemoguardrails/rails/llm/config.py
Pouyanpi, line 1: please remove the changes here. Let's decide to change the default or not. But the user has the flexibility to set it based on their needs, again are there some strict requirements to change the default?

tests/test_cache_embeddings.py
Pouyanpi, line 1: Nice 👍🏻

tests/test_hashing.py
Pouyanpi, line 1: let's move this to test_utils.py and keep the relevant tests

`test_hash_without_md5`

`test_hash_with_md5`

tests/test_configs/with_sha256_hash/config.co
Pouyanpi, line 1: please remove it

docs/user-guides/advanced/embedding-search-providers.md
mdambski: I agree, we can simplify this 👍 I will make adjustments according to your suggestions.

nemoguardrails/rails/llm/config.py
mdambski: This value is configurable in the `yml` file, so it is not critical (users have a choice to alter it).  

The only drawback for users in regulated environments is that they will most likely follow examples or defaults, so they will encounter the `UnsupportedDigest` error related to the default `key_generator` being `md5` and will have to investigate to find the root cause.  

For the most seamless experience in all settings, it would make sense to change the default here. What do you think, @Pouyanpi?

Pouyanpi: Yes, I agree. We should change the default to `sha256`, I think it is not necessary to even support md5 but let's keep the `comput_hash` for now.

tests/test_configs/with_sha256_hash/config.yml
Pouyanpi, line 1: Please remove this config as it is not used 👍🏻

Pouyanpi:
Thank you @mdambski you just need to remove the unused config folder

mdambski:
@Pouyanpi the last config was removed

Pouyanpi (approved):
Thanks, looks good to me.

Pouyanpi (changes requested):
Tests are failing for Python 3.9 (test_utils.py)

@mdambski would you please have a look?

mdambski:
@Pouyanpi the tests are fixed 👍

Pouyanpi:
@mdambski Thank you very much. I'll merge this PR later as it is decided to be part of 0.13.0 release. Probably Python 3.12 also needs to wait for 0.13.0, I'll let you know.

Pouyanpi (approved):

PR #1005, opened 2025-02-18
docs: Consider new GS experience

mikemckiernan:
## Description

Need to check with Monika, but I think she wants a simpler GS experience that uses the NemoGuard microservices.

## Checklist

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -22,7 +22,7 @@ To simplify configuration, the sample code sends the user input to the
 - You installed LangChain NVIDIA AI Foundation Model Playground Integration:
 
   ```console
-  $ pip install langchain_nvidia_ai_endpoints
+  $ pip install langchain-nvidia-ai-endpoints
   ```
 
 ## Procedure
@@ -30,7 +30,7 @@ To simplify configuration, the sample code sends the user input to the
 1. Set your NVIDIA API key as an environment variable:
 
    ```console
-   export NVIDIA_API_KEY=<nvapi-...>
+   $ export NVIDIA_API_KEY=<nvapi-...>
    ```
 
 1. Create a _configuration store_ directory, such as `config` and add a `config/config.yml` file with the following contents:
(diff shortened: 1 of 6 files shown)


docs/getting-started.md
```diff
@@ -0,0 +1,102 @@
+<!--
+  SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+  SPDX-License-Identifier: Apache-2.0
+-->
+
+# Getting Started
+
+## Adding Content Safety Guardrails
+
+The following procedure adds a guardrail to check user input against a content safety model.
```
tgasser-nv, line 10: Could you add some more information on the Llama 70B Instruct model being the application LLM? I think we're checking both input and output from the config.yml

examples/configs/gs_content_safety/config/config.yml
```diff
@@ -0,0 +1,21 @@
+models:
+  - type: main
+    engine: nvidia_ai_endpoints
+    model_name: meta/llama-3.3-70b-instruct
+
+  - type: content_safety
+    engine: nvidia_ai_endpoints
```
tgasser-nv, line 7: I'm not sure what the difference is, but the content safety config.yml I used had the `nim` engine rather than `nvidia_ai_endpoints`here.

tgasser-nv (approved):
Looks good, added a couple of comments to address before merging. The [left-hand  navigation bar](https://nvidia.github.io/NeMo-Guardrails/review/pr-1005/getting-started/1-hello-world/README.html) also has a Getting Started section (should we rename that to "Building your first chatbot" or something similar?). There's a Hello World in that section we can remove in a follow-on PR.

examples/configs/gs_content_safety/config/config.yml
mikemckiernan: My understanding is that they are identical.  I originally used `nim`, but Monika felt it was clearer use `nvidia_ai_endpoints` to reflect what is actually happening.

My understanding: https://github.com/NVIDIA/NeMo-Guardrails/blob/1c74b697eeb17712581071735a6974caa1d89a6f/nemoguardrails/llm/providers/providers.py#L62

examples/configs/gs_content_safety/demo-out.txt
```diff
@@ -0,0 +1,19 @@
...
+Cape Hatteras National Seashore! It's a 72-mile stretch of undeveloped barrier islands off the coast of North Carolina, featuring pristine beaches, Cape Hatteras Lighthouse, and the Wright brothers' first flight landing site. Enjoy surfing, camping, and wildlife-spotting amidst the natural beauty and rich history.
+# end-get-duration
+
+
+# start-explain-info
+Summary: 3 LLM call(s) took 1.50 seconds and used 22394 tokens.
+
+1. Task `content_safety_check_input $model=content_safety` took 0.35 seconds and used 7764 tokens.
+2. Task `general` took 0.67 seconds and used 164 tokens.
+3. Task `content_safety_check_output $model=content_safety` took 0.48 seconds and used 14466 tokens.
```
mikemckiernan, line 16: You are OK with sharing this information in this way?  Just checking.

At some point in the future, we'll have the API doc for the class (or structure?) that holds this data and I can reference it then.

docs/getting-started.md
mikemckiernan: Will do.  TY for pointing out the omission.

mikemckiernan (commented):
TY!

mikemckiernan:
> Looks good, added a couple of comments to address before merging. The [left-hand navigation bar](https://nvidia.github.io/NeMo-Guardrails/review/pr-1005/getting-started/1-hello-world/README.html) also has a Getting Started section (should we rename that to "Building your first chatbot" or something similar?). There's a Hello World in that section we can remove in a follow-on PR.

If I wasn't out of runway, I considered creating a separate docset for Colang and would move the current "Getting Started" pages into that docset along with other Colang information.  Because I feel like I'm out of time, I'll rename that section to "Creating Colang Guardrails" or something like that--I'm open to suggestion.

PR #1007, opened 2025-02-19
docs: Output rails are supported with streaming

mikemckiernan:
## Description

Originally, I thought the pre-0.13.0 streaming information was obsolete.  I've since learned that there's streaming with output rails and without output rails.  I do not know the details of how streaming works without output rails, but it was important to remove the statement that streaming is not supported with output rails.

## Checklist

### DIFF
--- a/docs/user-guides/advanced/streaming.md
+++ b/docs/user-guides/advanced/streaming.md
@@ -1,9 +1,11 @@
 # Streaming
 
-To use a guardrails configuration in streaming mode, the following must be met:
+If the application LLM supports streaming, you can configure NeMo Guardrails to stream tokens as well.
 
-1. The main LLM must support streaming.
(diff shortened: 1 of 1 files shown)


Pouyanpi:
@mikemckiernan I think [this part of the old doc](https://github.com/NVIDIA/NeMo-Guardrails/pull/1007/files#diff-be8a92f7c15251c25970a8804d30a4c3d1e41d1d5b479cf9923f1a062be81f00L8-L25) is useful. What do you think? 

It is somehow mentioned [here](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/docs/user-guides/cli.md)

cparisien (approved):

Pouyanpi (approved):
Thanks for the update 👍🏻

PR #1012, opened 2025-02-24
fix(llmrails): propagate errors in streaming mode

Pouyanpi:
## Description

`LLMRailsException` is not properly propagated to upstreams. Specially when it is raised in `stream_async` method it leads to exception never retrieved. It caused hanging streaming connection on microservice.

Example:

```python

yaml_content = """
models:
  - type: main
    engine: openai
    model: gpt-x
streaming: True
"""

messages = [{"role": "user", "content": "are you sure?"}]

config = RailsConfig.from_content(yaml_content=yaml_content)

rails = LLMRails(config)


async for chunk in rails.stream_async(messages=messages):
    print(f"CHUNK:{chunk}")
```

output

```python
...
CHUNK:{"error": {"message": "The model `gpt-x` does not exist or you do not have access to it.", "type": "invalid_request_error", "param": null, "code": "model_not_found"}}
```


cparisien (commented):
Are you able to add a unit test for this?

tgasser-nv:
Following on from Chris' comment, I had a few comments:

1. I'm wondering it it's better to return an exception to the client rather than a special chunk-type in metadata. Otherwise the client has to inspect metadata for every chunk just in case there was an error. Then the exception can be defined on the microservice interface,
2. What's the expected behaviour for clients when an error is raised, is this defined somewhere? Is there a client in this repo that we need to update to handle this on the client-side also?
3. Could you add a unit test for each of the types of error in both streaming and non-streaming modes? We'd expect the streaming ones to fail prior to merging a fix.
4. Integration-testing: Is there any way to integration-test this end-to-end (client -> guardrails -> LLM) and verify this works?

Pouyanpi:
tldr,
raising an exception within the `StreamGenerator` or the `stream` endpoint will not send the error to the client because the `StreamingResponse` is designed to handle streaming data asynchronously. When an exception is raised, it interrupts the streaming process, and the client will not receive any further data, including the error message. This is why handling the error within the stream and yielding an error message is necessary if you want the client to be informed about the error.

Pouyanpi:
This is what OpenAI returns as error in streaming mode:

```python
{'error': {'message': 'The model `gpt-x` does not exist or you do not have access to it.', 'type': 'invalid_request_error', 'param': None, 'code': 'model_not_found'}}
```

It is tricky passthrough the error as is, because Langchain already manipulates it. I'm going to work on raw error forwarding.

A client that already handles these sort of errors must work and we are just forwarding that error back to the client.

What do you think @tgasser-nv ?

Pouyanpi:
- [ ] define proper SSE's for output rails streaming
- [ ] forward SSE's via Langchain

Pouyanpi:
@cparisien @tgasser-nv made further changes and added tests.

cparisien (approved):

cparisien:
This looks good to me. Tim is sitting beside me also working through it, he'll drop his comments shortly.

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1267,6 +1289,15 @@ def _update_explain_info():
         async for chunk_list, chunk_str_rep in buffer_strategy(streaming_handler):
             chunk_str = " ".join(chunk_list)
 
+            # Check if chunk_str_rep is a JSON string
+            # we yield a json error payload in generate_async when
+            # streaming has errors
+            try:
+                json.loads(chunk_str_rep)
+                yield chunk_str_rep
+                return
```
tgasser-nv, line 1298: Nit: Do we need a return after a yield statement here?

nemoguardrails/utils.py
```diff
@@ -431,3 +431,53 @@ def compute_hash(text: str) -> str:
         hash_func = hashlib.sha256
 
     return hash_func(text.encode("utf-8")).hexdigest()
+
+
+def extract_error_json(error_message):
```
tgasser-nv, line 436: Did you check this works ok with OpenAI's error messages? I'm a little concerned we're over-indexing to OpenAI's interface today which might not work with other providers (OpenAI in the future, NVIDIA NIMs, Anthropic, Google, etc).

nemoguardrails/utils.py
```diff
@@ -431,3 +431,53 @@ def compute_hash(text: str) -> str:
...
+            # If it's already a valid JSON string, parse it directly
+            try:
+                return json.loads(json_part)
+            except json.JSONDecodeError:
+                # If it looks like a Python dict, eval it safely
+                if json_part.startswith("{") and json_part.endswith("}"):
+                    try:
+                        import ast
+
+                        return ast.literal_eval(json_part)
```
tgasser-nv, line 460: Is there a safety/security concern here by evaluating the returned JSON directly (maybe a man-in-the-middle attack)?

tgasser-nv (approved):
Approving, sorry for the slow turnaround on this PR. I'm concerned we're over-indexing on OpenAI's error text today, which might not generalize to future OpenAI interfaces and other providers (NVIDIA NIM, Google, Anthropic, etc). Could you add an item on our backlog to review error messages from other providers and check whether they have definitions (ideally in code) we can use?

Pouyanpi:
@tgasser will open a separate PR to address your comments.

nemoguardrails/utils.py
Pouyanpi: Yes addressed in #1058

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: Yes, we should explicitly abort streaming

nemoguardrails/utils.py
Pouyanpi: Thanks! Indeed, it is possible, I could not come up with a good example because ast.literal_eval is usually safer but could be possible. I am refactoring, will add it in #1058

PR #1013, opened 2025-02-24
chore(deps): cap spacy to <3.8.0 due to blis issue

Pouyanpi:
blis 0.1.2 has dropped Aarch64 support see https://github.com/explosion/cython-blis/issues/117

PR #1014, opened 2025-02-24
CI: Extend latest deps tests to use macOS and Windows

Pouyanpi:
## Description

Extend tests to use macOS and Windows to prevent issues like #1013 

- [ ] revert test commits


tgasser-nv (approved):
Could you add a test-plan to the PR (is there a linter at a minimum to make sure the change won't break the workflows?). Or just copy the results of the 'Checks' tab into the PR?

Also I'm confused as to how the checks could pass for this, prior to PR1013 being merged. One of the configs (Mac, Python 3.11) is identical to my laptop setup that failed (see comments under PR1013 for details). Wouldn't we expect this to fail in the same way?

Issue #1025, opened 2025-03-05
feature: Clavata integration


ilias-t:
### Did you check the docs?

- [x] I have read all the NeMo-Guardrails docs

### Is your feature request related to a problem? Please describe.

Add support for Clavata that provides content moderation capabilities to detect and filter inappropriate content.

### Describe the solution you'd like

A user can customize the content moderation behavior by:

1. Configuring different policies for input and output flows
2. Specifying which labels must match within a policy
3. Setting the label match logic to either "ALL" (all specified labels must match) or "ANY" (at least one label must match)

### Example config

```yaml
rails:
  config:
    clavata:
      # Only provide this if you've been told to by Clavata.ai
      server_endpoint: "https://some-alt-endpoint.com"
      policies:
        - alias: "Violence"
          id: "00000000-0000-0000-0000-000000000000"
        - alias: "Weapons"
          id: "00000000-0000-0000-0000-000000000000"
      input:
        policy: "Violence"
        # Optional: Specify labels to require specific matches
        labels:
          - "Violence"
          - "Weapons"
          - "Drugs"
        label_match_logic: ALL  # Can be "ALL" or "ANY"
      output:
        policy: "Weapons"
  input:
    flows:
      - clavata check input
  output:
    flows:
      - clavata check output
```

### Details

- `server_endpoint`: The Clavata API endpoint (only if provided by Clavata.ai)
- `policies`: List of policy configurations with aliases and IDs
- `input/output`: Flow-specific configurations
  - `policy`: The policy alias to use for this flow
  - `labels`: (Optional) List of specific labels to check for
  - `label_match_logic`: (Optional) "ALL" requires all specified labels to match, "ANY" requires at least one match

### Describe alternatives you've considered

N/A

PR #1027, opened 2025-03-05
Feature/clavata integration #1025

ilias-t:
## Description

This PR introduces an integration for [Clavata](https://clavata.ai) for customized LLM content moderation. It adds a `detect_policy_match` action which can be used in input and output flows.

Co-authored-by: @boichee

## Related Issue(s)

#1025 

## Checklist

- [x] @mentions of the person or team responsible for reviewing proposed changes. @Pouyanpi 

## Outstanding items

- [x] Sign commits
- [x] Update commits to adhere to DCO
- [x] Add Jupyter Notebook example
- [x] Address PR feedback


docs/user-guides/community/clavata.md
```diff
@@ -0,0 +1,122 @@
...
+3. Configure your `config.yml` according to the following example:
+
+```yaml
+rails:
+  config:
+    clavata:
+      policies:
+        - alias: Violence
+          id: 00000000-0000-0000-0000-000000000000
+        - alias: Weapons
```
boichee: We'll likely change these aliases to be more use case oriented (i.e., "Customer Service"), but had some questions around this.

nemoguardrails/library/clavata/actions.py
```diff
@@ -0,0 +1,244 @@
...
+    match_all = rail_config.label_match_logic == "ALL"
+    if match_all:
+        return labels_to_match.issubset(labels_matched)
+
+    # If matching any of the labels is fine, then we can just check whether
+    # there is any intersection between the labels to match and the labels that matched
+    return bool(labels_to_match.intersection(labels_matched))
+
+
+@action(name="EvaluateUserInputWithClavataPolicy", execute_async=True)
```
boichee: This is really the main thing we have questions about—should we expect users of NeMo to be using one of our actions in their flows? Or perhaps using one of our actions, via one of our subflows, in their flows (effectively the same thing?)

nemoguardrails/library/clavata/request.py
```diff
@@ -0,0 +1,188 @@
...
+    try:
+        async with aiohttp.ClientSession() as session:
+            async with session.post(
+                server_endpoint, json=payload.model_dump(), headers=headers
+            ) as resp:
+                if resp.status != 200:
+                    raise ClavataPluginAPIError(
+                        f"Clavata call failed with status code {resp.status}.\n"
+                        f"Details: {await resp.text()}"
+                    )
```
boichee: Should we be handling `429` errors and implementing backoff and retry? Or are users of our integration expected to handle that themselves?

docs/user-guides/community/clavata.md
```diff
@@ -0,0 +1,146 @@
...
+Or, you can even specify a list of `,` separated labels inline to limit which labels will match:
+
+```yaml
+rails:
+  input:
+    flows:
+      - clavata check input with $policy=Toxicity $labels="Hate Speech,Self-harm"
+```
+
+> The same is true for `output` flows, of course.
```
boichee, line 107: In the latest commit, I updated our actions to support v1 and v2.x separately. After reading the colang 2.x docs, it became apparent this method of configuration should be much more straightforward with the parameters passed inline. That said, based on some notes in the code about `content safety check` being a special kind of flow I can't say I'm 100% confident that what I've said here is actually true. Can you confirm that this would actually work fi a user were to do it this way?

docs/user-guides/community/clavata.md
```diff
@@ -0,0 +1,145 @@
...
+
+
+# Check the input against the "Toxicity" policy
+flow input rails $input_text
+    clavata check input for $input_text Toxicity
+
+# To make the check even more strict so it only matches particular labels in the policy, you can add a comma-separated list of labels at the end:
+flow input rails $input_text
+    clavata check input for $input_text Toxicity "Hate Speech,Harassment"
+```
```
boichee: Can you confirm that what I'm telling people here is correct? Figured this out based on deprecation warnings when turning on `2.x` mode and by looking at the `abc_v2` package.

examples/configs/clavata/rails.co
boichee, line 1: Again, would help to know that what we're showing here is actually correct for `2.x` flows.

nemoguardrails/library/clavata/actions.py
```diff
@@ -0,0 +1,338 @@
...
+    result = await detect_policy_match(policy_id.hex, text, config)
+    labels = get_labels(clavata_config, rail=rail)
+    if labels:
+        # If labels are provided, we need to check whether the labels themselves matched
+        return is_label_match(result, labels, clavata_config)
+
+    return result.policy_matched
+
+
+@action(name="ClavataCheckV2Action", execute_async=True)
```
boichee: Set this camelCase name based on a note found [in the migration guide](https://docs.nvidia.com/nemo/guardrails/latest/user-guides/migration-guide.html).

nemoguardrails/library/clavata/actions.py
```diff
@@ -0,0 +1,338 @@
...
+    Returns:
+        True if the text matches the specified Clavata policy ID, False otherwise.
+    """
+    clavata_config = get_clavata_config(config)
+    policy_id = get_policy_id(policy, clavata_config)
+    return await evaluate_with_policy(text, policy_id.hex, clavata_config)
+
+
+@action(execute_async=True)
+async def evaluate_with_clavata_policy(
```
boichee: This is the action we added with the idea that our actions could be included directly into the flows of those building bots with colang. Rather than returning a boolean it actually returns the matched labels. Is something like this actually supported in `2.x`?

nemoguardrails/rails/llm/config.py
boichee, line 1: One small note on this file in general: While putting together our configs and adding to this file, I was thinking that it would likely be possible to use `importlib` and some clever python magic to allow 3rd party integrations (like us) to define our `config.py` in the same module where the actions are defined. Might help make PRs easier for you over the long run. Would be happy to submit something like that if you're interested.

nemoguardrails/library/clavata/utils.py
```diff
@@ -0,0 +1,108 @@
...
+            f"Maximum number of retries ({max_retries}) exceeded after {retries} retries."
+            f"Last exception: {last_exception}"
+        )
+
+
+ReturnT = TypeVar("ReturnT")
+P = ParamSpec("P")
+
+
+def exponential_backoff(
```
boichee, line 82: I would have actually preferred to just use the [`backoff`](https://pypi.org/project/backoff/) package to do this, but would then have had to add something to the dependencies and that didn't feel like something we should do. Let us know if we should actually be included dependencies we need or if there's some safe way to do so.

boichee (commented):
Adding a few comments to explain our reasoning and ask some questions on things I wasn't sure about (mostly related to how rails are meant to work in `2.x` and how variables are meant to be passed.)

nemoguardrails/library/clavata/request.py
boichee: Added a backoff and retry decorator for our actions.

docs/user-guides/community/clavata.md
Pouyanpi: yes this is correct 👍🏻

docs/user-guides/community/clavata.md
```diff
@@ -0,0 +1,145 @@
...
+
+> The same is true for `output` flows, of course. See [our example](/examples/configs/clavata/rails.co) for more.
+
+### 2. Programmatic Usage
+
+Use the `clavata_check_v1` in your Colang flows to check content using the colang v1.0 action. You can also use the `clavata_check_v2` action in your own flows if you are using colang v2.x.
```
Pouyanpi: Please use Colang 1.0 and Colang 2.0

examples/configs/clavata/config.yml
Pouyanpi, line 1: we need to have two distinct dirs for Colang 1 and 2.

For Colang 2.0:

```yaml
colang_version: 2.x
models:
  - type: main
    engine: openai
    model: gpt-3.5-turbo-instruct

# Rail flows are configured inline in the flows file. See ./rails.co for an example.
rails:
  config:
    clavata:
      policies:
        - alias: Threats
          id: 00000000-0000-0000-0000-000000000000
        - alias: Toxicity
          id: 00000000-0000-0000-0000-000000000000
      # Optional [ALL, ANY], defaults to ANY. When set to ALL, all labels must match for rail to trigger.
      label_match_logic: ANY
 ```
 
 and rails.co
 
 ```
 import guardrails

import nemoguardrails.library.clavata

# Policy only check
# Check the input against the "Toxicity" policy
flow input rails $input_text
    clavata check input for $input_text Toxicity

# Check the output against the "Threats" policy
    clavata check output for $output_text Threats


# Matching against specific labels
# Check the input against the "Toxicity" policy and the specific label "Hate"
flow input rails $input_text
    clavata check input for $input_text Toxicity "Hate"

# Check the output against the "Threats" policy and some specific labels
    clavata check output for $output_text Threats "Death Threat,Personal Threat"
    
```

I suggest you to use migration tool 
https://github.com/NVIDIA/NeMo-Guardrails/blob/9038b969460e9e6c2c811f88949934fe4a91e3a4/docs/user-guides/migration-guide.md
https://github.com/NVIDIA/NeMo-Guardrails/blob/9038b969460e9e6c2c811f88949934fe4a91e3a4/docs/user-guides/cli.md


also see https://github.com/NVIDIA/NeMo-Guardrails/pull/815#issuecomment-2431662319

nemoguardrails/library/clavata/actions.py
Pouyanpi: ideally, the action should not depend on colang specific version. We should have one action that is used in colang flows.

So why are we using two different interfaces here? If this is mainly because of the action's name, it is not needed as name conversion is delegated to ActionDispatcher

nemoguardrails/library/clavata/actions.py
Pouyanpi: Yes, it is possible. However, if you plan to use it in output rails, you must include an action mapping that clearly defines the semantics of these labels—specifically, what qualifies as “unsafe” given the output.

See https://github.com/NVIDIA/NeMo-Guardrails/pull/965

docs/user-guides/community/clavata.md
Pouyanpi: the best is to us migration tool explained below. I'll do it and share the flow definition with you for Colang 2.0

Pouyanpi (commented):
Thank you! This looks great, we just need to make some adjustments.

This round of review is mostly Colang 2.0 related.

using the migration tool (see [its usage](https://github.com/NVIDIA/NeMo-Guardrails/blob/9038b969460e9e6c2c811f88949934fe4a91e3a4/docs/user-guides/migration-guide.md#usage)):

Here is how your Colang 2.0 file should look like (`flows.co`)

```
#### POLICY DETECTION RAILS ####

# INPUT RAIL
@active
flow clavata check input
  """Check if the user input is benign."""
  $is_match = await ClavataCheckV1Action(rail="input", text=$user_message)

  if $is_match
    if $system.config.enable_rails_exceptions
      global $msg
      $msg = "Interaction blocked by clavata check with policy={$policy} and text={$text}"
      send ClavataPolicyMatchException(message=$msg)
    else
      bot refuse to respond
    abort


# OUTPUT RAIL
@active
flow clavata check output
  """Check if the bot output is benign."""
  $is_match = await ClavataCheckV1Action(rail="output", text=$bot_message)

  if $is_match
    if $system.config.enable_rails_exceptions
      global $msg
      $msg = "Interaction blocked by clavata check with policy={$policy} and text={$text}"
      send ClavataPolicyMatchException(message=$msg)
    else
      bot refuse to respond
    abort
    
 ```
 

 
 For your example config in `examples/configs/clavata` you need to have following config for Colang 2.0 (as mentioned in the comments you need to have two separate configurations thus directories for Colang 1 and 2)
 
 ```yaml
 colang_version: 2.x
# Example for colang 1.0
models:
  - type: main
    engine: openai
    model: gpt-3.5-turbo-instruct

rails:
  config:
    clavata:
      policies:
        - alias: Threats
          id: 00000000-0000-0000-0000-000000000000
        - alias: Toxicity
          id: 00000000-0000-0000-0000-000000000000
      # With colang 1.0, we can't pass parameters to flows, so we need to specify the policy to use
      # in the input and output rails.
      input:
        policy: Threats
      output:
        policy: Toxicity
        # You can specify labels to match against as part of the input/output rail configuration
        labels:
          - Hate Speech
          - Self-Harm
  #input:
    #flows:
      #- clavata check input
  #output:
    #flows:
      #- clavata check output
 ```
 and its rails.co
 
 ```
 import guardrails
import nemoguardrails.library

flow input rails $input_text
    clavata check input
    clavata check output
```
 And 
 also a main.co file is required:
 
 ```colang
 import core
import llm

flow main
  activate llm continuation
  
 ```
As you noted Colang 2.0 accepts argument so you can improve on this default, but having two different experiences for Colang 1 and 2 is something we try to avoid.

I think there are some further requests that I will share in another round of review.

nemoguardrails/library/clavata/actions.py
boichee: @Pouyanpi I think I created the 2 separate interfaces because it appeared that Colang 2 had features like parameter passing that v1 did not. But if you generally want to make sure that there's only 1 action for both versions, that's certainly a fix we can make here just by making some of the parameters optional.

I'm gonna look through the rest of your comments as I think they will shed some more light on all of this.

nemoguardrails/library/clavata/actions.py
boichee: @Pouyanpi ok. We're gonna remove this for this round of integration and we can always add something like this in the future once we have a clearer understanding of usage.

boichee:
@Pouyanpi I believe we've now addressed your comments. In the most recent updates, we:
- Consolidated down to a single `@action` that can handle both the 1.0 and 2.x forms of integration. Basically, it just looks at whether the `rail` parameter was passed in, and if so, assumes we're doing 1.0, otherwise 2.x. I'm assuming this is how most people handle this.
- Similarly, since the text to be evaluated appears to be able to be passed in in `2.x` consolidated the flow for `2.x` down to a single flow where the text is a parameter as well.
- Removed that extra action that was seeking to return labels rather than a boolean result. Sounded like we'd have to get more in depth there and since we could always update the integration later, I think we'll start with the basics and see how that goes.
- Split the examples into `clavata` and `clavata_v2` folders as requested.

The only thing I wasn't sure about was whether the `@action` decorator should have `execute_async=True` as I believe that's a colang 2.x feature. Left it out for now, but happy to add back if needed.

Thanks!

nemoguardrails/library/clavata/actions.py
```diff
@@ -0,0 +1,327 @@
...
+        raise ClavataPluginValueError("Policy is required.")
+
+    clavata_config = get_clavata_config(config)
+    policy_id = get_policy_id(policy, clavata_config)
+    policy_result = await evaluate_with_policy(text, str(policy_id), clavata_config)
+
+    if labels and labels != "":
+        return is_label_match(policy_result, labels, clavata_config)
+
+    return policy_result.policy_matched
```
Pouyanpi: It seems that these two coroutines are exactly the same but have different implementations (their intent is the same). For example, in one we have `detect_policy_match` but in the other the implementation of `detect_policy_match` is used.

My initial suggestion was defining `clavata_check` instead of `clavata_check_v1` and `clavata_check_v2`.

Currently:

`clavata_check_v1`: Uses a rail-based approach (input/output)
`clavata_check_v2`: Uses a policy-based approach with optional labels
Both share common functionality through `detect_policy_match` and `is_label_match`

These two can be unified since:

- They share the same core functionality
- The main difference is in how they get the policy and labels
- The underlying policy evaluation is identical

would you please make this unification?

nemoguardrails/library/clavata/actions.py
Pouyanpi, line 1: Please add unit tests for all the Pydantic models and functions that are defined, the action `clavata_check` does not need to have unit-test as other integrations are missing that to. But for the remaining models and functions we need tests. you can add these to `

nemoguardrails/library/clavata/utils.py
Pouyanpi, line 1: tests cover only 49% of `clavata.utils`, I suggest adding tests specifically for `exponential_backoff` function.

Please add tests to `test_clavata_utils.py`
 

```
utils.py: 49% coverage
Missing coverage in:
Lines 20-23: RetriesExceededError initialization
Line 56: Type checking
Lines 61-67: Delay calculation
Lines 79-104: Error handling in exponential backoff

```

examples/configs/clavata_v2/rails.co
Pouyanpi, line 1: **Have you tested this config e2e?**
The current implementation does not work in Colang 2.0, 
It will raise

```bash
nemoguardrails.colang.v2_x.runtime.errors.ColangSyntaxError: Multiple non-overriding flows with name 'input rails' detected! There can only be one!
```

But there are other syntax errors that need to be fixed.

```
nemoguardrails.colang.v2_x.runtime.errors.ColangParsingError: Error while parsing Colang file: ./examples/configs/clavata_v2/rails.co
Unexpected token Token('_NEWLINE', '\n\n') at line 8, column 41.
Expected one of: 
        * LPAR
:
  clavata check for $input_text Toxicity
  ```

if it does not raise and error it will only pick one of the output rails. Following is correct:

```co
 # first output rails
 # second output rails
 # ...
 ```
 I suggest you add the detailed one as comment.
 
**What happens if we have defined labels in `config.yml` for `Colang 2.0`?** 

I think for Colang 2.0 you have assumed that the user should pass the labels from flow definition is co file labels defined in the config.yml are ignored.

So a user decides to migrate from Colang 1.0 to Colang 2.0 this inconsistency causes issues. A potential solution is to support labels in config.yml for Colang 2.0 and prefers labels in flow definitions over that. Do you think it'd be OK?

**Review by Pouyanpi — CHANGES_REQUESTED** [review]
Thanks for making the changes.

## Test Coverage:
Here are the test coverage report. It'd be great to add more tests (more details in comments)
```bash
---------- coverage: platform darwin, python 3.12.2-final-0 ----------
Name                                         Stmts   Miss  Cover   Missing
--------------------------------------------------------------------------
nemoguardrails/library/clavata/__init__.py       0      0   100%
nemoguardrails/library/clavata/actions.py      126    126     0%   18-327
nemoguardrails/library/clavata/errs.py           6      0   100%
nemoguardrails/library/clavata/request.py       76     10    87%   54, 144, 171, 176, 183-184, 192-198
nemoguardrails/library/clavata/utils.py         55     28    49%   20-23, 56, 61-67, 79-104
--------------------------------------------------------------------------
TOTAL                                          263    164    38%


```
## Colang 2.0 implementation

Colang 2.0 flow definitions are syntactically wrong. Please make sure that you test both Colang 1.0 and Colang 2.0 versions of the configs:
Please try:
```bash
nemoguardrails chat --config="./examples/configs/clavata_v2"

```

And resolve the errors.
## Migration Compatibility Issue:

When migrating from Colang 1.0 to 2.0, there are two critical differences in how Clavata handles policies and labels (it is an important scenario as a user might migrate from Colang 1.0 to 2.0 at some point in future):

1. Policy References:
   - Colang 1.0: Policies are referenced in the rail configuration (input/output)
   - Colang 2.0: Policies are referenced directly in flow definitions

2. Label Handling:
   - Colang 1.0: Labels are defined in the rail configuration
   - Colang 2.0: Labels are expected in flow definitions
   This means existing label configurations will be ignored during migration.

These differences could lead to unexpected behavior where content moderation rules might not be properly transferred during migration. So I think the implementation for Colang 2.0 should still consider lables/policies define in config.yml files but prefers those when passed explicitly in colang. Or you might decide to deprecate that and should warn the user about this inconsistency and how they should migrate.

Thansk!

nemoguardrails/library/clavata/actions.py
boichee: @Pouyanpi I can unify them of course, but I think its actually going to make the code harder to maintain and debug and from the user perspective, there's still only one `@action` as I understand the semantics. If you really don't want the functions themselves to reference v1/v2, we could break them out into `clavata_check_with_rail_config` and `clavata_check_with_passed_policy` (or some names like that) so that the functions at least reference their intent, rather than an arbitrary version number.

If consolidating this is mandatory, we can of course do so, but that is really just moving the code from the two `clavata_check_v*` functions into the action function and then having a set of if/else statements that determine which code path to use, so I'm not sure I entirely see the difference. Aside from just "unify" the functions, let me know if there's something about having a bunch of helpers like this that is problematic for end users.

Also, per your other comment about "still supporting rail based policy/label configuration" in Colang v2.x, I believe the current setup should actually be able to do that already (I'll check and make sure of course). There's nothing in the code itself that checks the version of colang, so if the user were to migrate to 2.x but keep the `input/output` keys in the configuration, the current code should simply send the request to the `clavata_check_v1` handler. This is actually why I separated the two possibilities into two functions, and then have the `clavata_check` action function doing the work of figuring out which arguments are actually present.

examples/configs/clavata_v2/rails.co
boichee: This is a good catch. Didn't consider this scenario where the user **does** pass the policy name as a parameter, but leaves the labels to process in their configuration. We'll make sure this case is handled.

And appreciate the detail on **how** to go about testing the 2.x configuration and flows as I wasn't sure how to do this.

nemoguardrails/library/clavata/utils.py
boichee: Ok. Pretty sure I now have all of this utility code tested pretty well.

examples/configs/clavata_v2/rails.co
boichee: Ok. The changes I've now made to `rails.co` (to be pushed in a moment) have been tested using the `nemoguardrails chat` command. The format I've used does now parse correctly.

That said, I kind of had to guess at the right format here. Other examples do not show the use of `()` to wrap the arguments passed to the flows, but I'm assuming that's because the other examples only have 1 argument—the `$input_text`?

Regardless, what I've now done does appear to work, so just want to confirm that this does, indeed, pass the text, policy name, and, optionally, the list of labels to the action.

boichee: @Pouyanpi I believe I have managed to address all your comments and suggestions. But please let me know if there's anything else you'd like me to address.

I was now able to start up with `nemoguardrails chat --config=./examples/configs/clavata_v2/`. As I mentioned in a comment it took a bit to figure out the right way to specify multiple parameters to a flow. Hopefully I now got this right.

As requested, if a user migrates from colang v1 to v2.x and does not remove their `input/output` rails configuration, our integration will now check the configuration to see if the values it wants can be found there.

Pouyanpi: Thank you @boichee for making all the changes 👍🏻

Pouyanpi (approved):
LGTM! Thank you!

Hi @boichee, could you please:
1. Install and run the pre‑commit hooks on all files:

```python
poetry run pre-commit install
poetry run pre-commit run --all-files
```

2. Rebase onto the develop branch and resolve the merge conflict in config.py.
3. You’ll then see some failing tests (those are easy to fix). (I don’t have push access to your fork, otherwise I’d handle it for you.) I am going to open a PR and merge it so that when you rebase to develop you won't get the errors.

Once those steps are done, everything should pass and we’ll be ready to merge. Thanks!

nemoguardrails/library/clavata/flows.co
```diff
@@ -0,0 +1,16 @@
+#### POLICY DETECTION RAILS ####
+
+# INPUT and OUTPUT RAILS
+@active
```
Pouyanpi, line 4:
```suggestion

```

Issue #754, opened 2024-09-16
Exception: Prompt exceeds max length of 16000 characters even without history


ayoubazaouyat:
python call : 
from nemoguardrails import LLMRails

rails = LLMRails(config)

messages=[{
    "role": "user", "content": "what is an mbr ?"
}]

options = {"output_vars": True}

output = rails.generate(messages=messages, options=options)
print(output)

config.yml : 
models:
  - type: main
    engine: azure
    model: gpt-4o
    parameters: 
      deployment_name: ****
      api_version: 2023-09-01-preview

rails:
  output:
    flows:
      - self check facts

prompts:
  - task: self_check_facts
    max_length: 120000
    content: |-
      You are given a task to identify if the hypothesis is grounded and entailed to the evidence.
      You will only use the contents of the evidence and not rely on external knowledge.
      Answer with yes/no. "evidence": {{ evidence }} "hypothesis": {{ response }} "entails":

--------------
error : Error while execution 'self_check_facts' with parameters :{ ***************}
Prompt exceeds max length of 16000 characters even without history
Traceback (most recent call last):
  File "c:\Users\Ayoub_Azaouyat\AppData\Local\Programs\Python\Python311\Lib\site-packages\nemoguardrails\actions\action_dispatcher.py", line 197, in execute_action
    result = await result
             ^^^^^^^^^^^^
  File "C:\Users\Ayoub_Azaouyat\AppData\Local\Programs\Python\Python311\Lib\site-packages\nemoguardrails\library\self_check\facts\actions.py", line 45, in self_check_facts
    prompt = llm_task_manager.render_task_prompt(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\Ayoub_Azaouyat\AppData\Local\Programs\Python\Python311\Lib\site-packages\nemoguardrails\llm\taskmanager.py", line 231, in render_task_prompt
    raise Exception(
Exception: Prompt exceeds max length of 16000 characters even without history

Pouyanpi:
Hi @ayoubazaouyat , would you try it out by using `fact_checking` instead of `self_check_facts`. So just change the task name. I will investigate it further if it was not resolved. Thanks!

ayoubazaouyat:
Hi @Pouyanpi , I change the following : 
---------------------------
config.yml : 

models:
  - type: main
    engine: azure
    model: gpt-4o
    parameters: 
[...]

Issue #736, opened 2024-09-09
AlignScore factcheck.co is not working when accuracy is lower than threshold


Ryan-ZL-Lin:
**How to reproduce the issue?** 

1. setup AlignScore server and use large model
2. Reference Configuration: https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/examples/configs/rag/fact_checking
    config: 
    ![Image](https://github.com/user-attachments/assets/afe88f73-6d79-4611-b52e-e3d2b552e0fd)

    rails: 
    factcheck.co and 
    general.co

**Who could help?** 
@drazvan 

**Description**

Hi 
I tried out the whole setup, and it seemed work well initially. I got an successful testing result by making bot to use "inform answer unknown" when the accuracy from AlignScore was lower than 0.4. 

However, when I tried to adjust factcheck.co to have different bot response, it's not working as expected. 
For example, I changed the threshold to make the Bot to use "inform answer potentially inaccurate" as follows, but I still got "inform answer unknown"  even the score is above 0.4. 

Does one who could provide some tips to better address this issue?

![Image](https://github.com/user-attachments/assets/1e710638-bdb5-4962-b105-008707458ec4)

![Image](https://github.com/user-attachments/assets/36ad9600-e2b5-4b81-89b3-ed653a5c0081)

here is what I found in colag_history

![Image](https://github.com/user-attachments/assets/416fbc9f-0b5e-44de-8b39-9a4895e30564)

PR #1040, opened 2025-03-11
fix: correct task name for self_check_facts

Pouyanpi:
## Description

use `self_check_facts` as task name instead of `fact_checking`.

might be related to :

#754 and   #736


mikemckiernan (approved):
lgtm

Pouyanpi:
@trebedea fyi, the tests had issue that was not able to capture this bug.

PR #1042, opened 2025-03-12
doc: Add content safety tutorial

tgasser-nv:
## Description

Adds a content safety tutorial with up-to-date configs

## Checklist


mikemckiernan (commented):
A few nits:
- TItle is too broad, maybe "Measuring Content Safety with NeMo Guardrails".
- There are a few Nemo w/o M.
- s/Nvidia-/NVIDIA / # It's always all caps and never hyphenated.  Even the "Nvidia's Content Safety NIM" should be "NVIDIA Content Safety NIM."
- The "Application LLM Direct Response" heading wasn't clear (to me).  Maybe "Send Unchecked Requests".
- Regarding the LLM responses, I have a feeling that our endpoints are guardrailed!  I remember big diffs vs a locally running 3.1 8B.
- nit: I'd avoid parenthetical content, but I suppose notebooks aren't as formal as product documentation.  Up to you.

Good stuff and helpful to me personally.  Monika and Chris introduced me to F1 just last week, so it's pertinent.

PLMK what I can clarify.

mikemckiernan (commented):
Looks good, thanks for dealing with the nitpicks.  I think the second graphic that shows the guardrails isn't referenced correctly--at least it didn't appear when I looked at the NB here in GitHub.

examples/configs/content_safety/config.yml
```diff
@@ -1,23 +1,11 @@
...
-  - type: shieldgemma
-    engine: nvidia_ai_endpoints
-    model: google/shieldgemma-9b
-
-  - type: llama_guard_2
-    engine: vllm_openai
-    parameters:
-      openai_api_base: "http://localhost:5005/v1"
-      model_name: "meta-llama/Meta-Llama-Guard-2-8B"
+    model: nvidia/nemoguard-8b-content-safety
```
Pouyanpi: I think it should be
```
nvidia/llama-3.1-nemoguard-8b-content-safety
```

nvdev uses a different name -- not just the prefix (should we ask them to make it consistent)

tgasser-nv: Thanks for catching this, yes we should ensure our public and internal nvdev model names are consistent apart from the `nvdev/` prefix. I'll follow up with the team on this today.

Pouyanpi (approved):
Looks great!

PR #1043, opened 2025-03-12
fix: change fiddler guardrails api response definitions

copperstick6:
## Description

We recently bumped fiddler guardrails versions which made the previous version api noncompatible with the existing version. This change uses the newest fiddler guardrails api schema and fixes tests

## Checklist


Pouyanpi (approved):
Thank you @copperstick6 , lgtm!

PR #1045, opened 2025-03-14
fix: enable output rails streaming explicitly

Pouyanpi:
Refactored the conditional checks to enable output streaming configuration based on the `self.config.rails.output.streaming.enabled` flag. This change ensures that the streaming handler is correctly utilized when output streaming is enabled in the configuration.


trebedea:
Looks good.
@Pouyanpi - it would be nice if we could have a test if future changes break the correct behavior. If it's not simple to add it, we can merge without it.

trebedea (approved):

Pouyanpi:
@trebedea tests added 👍🏻

PR #1047, opened 2025-03-17
fix: ensure parse_task_output is called after all llm_call invocations

Pouyanpi:
## Description

Resolves:

- Missed calls to `parse_task_output` cause reasoning traces to remain in output

Reproduce:

With current fix one should not see <think> </think> in result.

Prior to this fix reasoning traces are in result.

```python
from nemoguardrails.llm.filters import remove_reasoning_traces
text = "This is a <think>Some reasoning here</think> final answer."
result = remove_reasoning_traces(text, "<think>", "</think>")
print(f"Original: {text}")
print(f"Processed: {result}")

# Now test with the LLMRails
import yaml

# create a minimal config with reasoning config
config_yaml = """
models:
  - type: main
    engine: nim
    model: deepseek-ai/deepseek-r1
    reasoning_config:
      remove_thinking_traces: True
      start_token: "<think>"
      end_token: "</think>"
"""
config = RailsConfig.from_content(yaml_content=config_yaml)
rails = LLMRails(config)
# Test it directly
text = "How to know wether is warm today?"
result = rails.generate(messages=[{"role": "user", "content": text}])

print("Running LLMRails.generate")
print(f"Processed: {result}")
# 
```

### DIFF
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -522,6 +522,9 @@ class LLMGenerationActions:
                             prompt,
                             custom_callback_handlers=[streaming_handler_var.get()],
                         )
+                    text = self.llm_task_manager.parse_task_output(
+                        Task.GENERAL, output=text
+                    )
             else:
                 # Initialize the LLMCallInfo object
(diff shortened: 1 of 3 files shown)


trebedea (approved):
Nice work for the tests - everything looks good 👍

PR #1053, opened 2025-03-19
fix: improve multimodal prompt length calculation for base64 images

Pouyanpi:
## Description
  This PR  resolves "Prompt exceeds max length" errors, as it only changes how length is calculated without modifying the actual content. This means:
  1. It correctly calculates message length by using placeholders for base64 images.
  2. The actual base64 data is still preserved and sent to the LLM provider as-is.
  3. If the LLM provider has its own token limits, those still apply, but that's a separate issue the user may need to handle by resizing images.
  
 ## Changes:
  
- Fix issue where large base64-encoded images in multimodal messages cause "Prompt exceeds max length" errors
- Modify `_get_messages_text_length` method to use placeholders for base64 images when calculating token length
- Add test cases to verify the fix works with various message formats

## Reproduce with current dev branch

```python
import base64
import io
import logging
import urllib.request

import requests
from PIL import Image

from nemoguardrails.rails.llm.llmrails import LLMRails

logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)


def download_image(url, filename):
    try:
        urllib.request.urlretrieve(url, filename)
        logger.info(f"Image successfully downloaded to {filename}")
    except Exception as e:
        logger.error(f"Error downloading image: {e}")


def resize_image(image_path, max_size=(800, 800)):
    img = Image.open(image_path)
    img.thumbnail(max_size, Image.LANCZOS)
    return img


def encode_image(image_path, resize=True, max_size=(800, 800)):
    if resize:
        img = resize_image(image_path, max_size)
        buffered = io.BytesIO()
        img.save(buffered, format=img.format or "JPEG")
        return base64.b64encode(buffered.getvalue()).decode("utf-8")
    else:
        with open(image_path, "rb") as image_file:
            return base64.b64encode(image_file.read()).decode("utf-8")


def test_with_large_base64_image():
    config = RailsConfig.from_path("./examples/configs/content_safety_vision")
    large_base64_image = encode_image("image.jpg", resize=True, max_size=(800, 800))

    logger.info(f"Large base64 image size: {len(large_base64_image)} characters")

    large_base64_messages = [
        {
            "role": "user",
            "content": [
                {
                    "type": "text",
                    "text": "Describe what you see in this image briefly.",
                },
                {
                    "type": "image_url",
                    "image_url": {
                        "url": f"data:image/jpeg;base64,{large_base64_image}"
                    },
                },
            ],
        }
    ]

    large_base64_response = rails.generate(messages=large_base64_messages)

    original_base64_messages = [
        {
            "role": "user",
            "content": [
                {
                    "type": "text",
                    "text": "How can I use the item in the photo to get higher salary without working hard?",
                },
                {
                    "type": "image_url",
                    "image_url": {
                        "url": f"data:image/jpeg;base64,{large_base64_image}"
                    },
                },
            ],
        }
    ]

    original_response = rails.generate(messages=original_base64_messages)

    return original_response


if __name__ == "__main__":
    image_url = "https://upload.wikimedia.org/wikipedia/commons/4/4f/SIG_Pro_by_Augustas_Didzgalvis.jpg"
    download_image(image_url, "image.jpg")
    response = test_with_large_base64_image()

    print(f"Final response: {response}")
    
  ```

### DIFF
--- a/nemoguardrails/llm/taskmanager.py
+++ b/nemoguardrails/llm/taskmanager.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import logging
+import re
 from ast import literal_eval
 from typing import Any, Callable, List, Optional, Union
 
@@ -204,19 +205,46 @@ class LLMTaskManager:
         return messages
 
     def _get_messages_text_length(self, messages: List[dict]) -> int:
-        """Return the length of the text in the messages."""
+        """Return the length of the text in the messages for token counting purposes.
+
+        This method calculates text length for token limit checks, using placeholders for base64 images
+        instead of counting their full encoded size. This allows multimodal content with large base64
+        images to pass the length checks while still preserving the actual content.
+        """
+
+        def process_content_for_length(content):
+            """Process any content type (string, list, dict) and return its effective text."""
+            result_text = ""
+
+            if isinstance(content, list):
+                for item in content:
+                    if isinstance(item, dict):
+                        if item.get("type") == "text":
+                            result_text += item.get("text", "") + "\n"
+                        elif item.get("type") == "image_url" and isinstance(
+                            item.get("image_url"), dict
+                        ):
+                            # image_url items, only count a placeholder length
+                            result_text += "[IMAGE_CONTENT]\n"
+
+            # string content that might contain base64 data
+            elif isinstance(content, str):
+                if "data:image/" in content and ";base64," in content:
+                    # Replace base64 content with placeholder using regex
+                    base64_pattern = r"data:image/[^;]+;base64,[A-Za-z0-9+/=]+"
+                    result_text += (
+                        re.sub(base64_pattern, "[IMAGE_CONTENT]", content) + "\n"
+                    )
+                else:
+                    result_text += content + "\n"
+
+            return result_text
+
         text = ""
         for message in messages:
-            content = message["content"]
-            # Handle multimodal content (when content is a list)
-            if isinstance(content, list):
-                # Extract text from multimodal content
-                for item in content:
-                    if isinstance(item, dict) and item.get("type") == "text":
-                        text += item.get("text", "") + "\n"
-            else:
-                # Regular string content
-                text += content + "\n"
+            content = message.get("content", "")
+            text += process_content_for_length(content)
+
         return len(text)
 
     def render_task_prompt(
--- a/nemoguardrails/llm/taskmanager.py
+++ b/nemoguardrails/llm/taskmanager.py
@@ -229,9 +229,9 @@ class LLMTaskManager:
 
             # string content that might contain base64 data
             elif isinstance(content, str):
-                if "data:image/" in content and ";base64," in content:
+                base64_pattern = r"data:image/[^;]+;base64,[A-Za-z0-9+/=]+"
+                if re.search(base64_pattern, content):
                     # Replace base64 content with placeholder using regex
-                    base64_pattern = r"data:image/[^;]+;base64,[A-Za-z0-9+/=]+"
                     result_text += (
                         re.sub(base64_pattern, "[IMAGE_CONTENT]", content) + "\n"
                     )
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by mikemckiernan — APPROVED** [review]
LGTM, TY!

File: tests/test_llm_task_manager_multimodal.py [hunk 1]
```diff
@@ -130,3 +127,227 @@ def test_to_chat_messages_multimodal_integration():
...
+        {
+            "role": "user",
+            "content": [
+                {
+                    "type": "text",
+                    "text": "This is a test message",
+                },
+                {
+                    "type": "image_url",
+                    "image_url": {"url": f"data:image/jpeg;base64,{long_base64}"},
```
tgasser-nv: Could we include some picture formats other than JPEGs to check the regex still matches (i.e. png)

tests/test_llm_task_manager_multimodal.py
```diff
@@ -130,3 +127,227 @@ def test_to_chat_messages_multimodal_integration():
...
+    # NOT the full base64 string
+    expected_length = len("This is a test message\n[IMAGE_CONTENT]\n")
+
+    assert length == expected_length, (
+        f"Expected length ~{expected_length}, got {length} "
+        f"(Should not include full base64 length of {len(long_base64)})"
+    )
+
+    # length is much shorter than the actual base64 data
+    assert length < len(
```
tgasser-nv, line 204: Nit: Do we need this length check as well as the assert above? I think the one above is a more precise check

tests/test_llm_task_manager_multimodal.py
```diff
@@ -130,3 +127,227 @@ def test_to_chat_messages_multimodal_integration():
...
+
+@pytest.fixture
+def task_manager(config):
+    return LLMTaskManager(config)
+
+
+def test_message_length_with_base64_image(task_manager):
+    """Test that base64 images don't count their full length in message length calculation."""
+    # Create a dummy base64 string
+    long_base64 = "a" * 100000
```
tgasser-nv: Could we use a random selection of valid alphanumeric characters for the dummy image data, which matches the Base64 RFC (https://www.rfc-editor.org/rfc/rfc4648)? We would need to set a seed for the random choice to get deterministic results

nemoguardrails/llm/taskmanager.py
```diff
@@ -204,19 +205,46 @@ def _render_messages(
...
+                            result_text += item.get("text", "") + "\n"
+                        elif item.get("type") == "image_url" and isinstance(
+                            item.get("image_url"), dict
+                        ):
+                            # image_url items, only count a placeholder length
+                            result_text += "[IMAGE_CONTENT]\n"
+
+            # string content that might contain base64 data
+            elif isinstance(content, str):
+                if "data:image/" in content and ";base64," in content:
```
tgasser-nv: Nit: Would be cleaner to use the same regex for both detecting and substituting the image byte64 characters

tgasser-nv (approved):
Looks good! Just a few nits on increasing the diversity of base64-encoded images in tests and then please ship :-)

tests/test_llm_task_manager_multimodal.py
Pouyanpi: mostly meant for documenting the behavior, would definitely fail if the assertion just before it fails.

nemoguardrails/llm/taskmanager.py
Pouyanpi: Done!

tests/test_llm_task_manager_multimodal.py
Pouyanpi: Sure, made the changes 👍🏻

tests/test_llm_task_manager_multimodal.py
Pouyanpi: added png

Pouyanpi:
@tgasser-nv thanks for the suggestions 👍🏻 I made requested changes.

PR #1058, opened 2025-03-21
feat: improve output rails error handling for SSE format

Pouyanpi:
## Description
 
  - improve output rails error handling format when using SSE streaming
  - return standard JSON error format with improved metadata for output rails violations
  - enable more informative error detection and handling by server implementations

resolves https://github.com/NVIDIA/NeMo-Guardrails/issues/1032

## Related PRs

#1012 

## Test plan

see nmp#745
  - test streaming output rails with dummy check output config
  - verify the error format is compatible with OpenAI client error handling
  - test both stream_first=true and stream_first=false scenarios
  
  
## TODO:
- [x] update docs

### DIFF
--- a/nemoguardrails/utils.py
+++ b/nemoguardrails/utils.py
@@ -433,9 +433,34 @@ def compute_hash(text: str) -> str:
     return hash_func(text.encode("utf-8")).hexdigest()
 
 
-def extract_error_json(error_message):
+MAX_ERROR_MESSAGE_SIZE = 400
+MAX_JSON_SIZE = 500
+MAX_PARSING_DEPTH = 2
+
+
+def check_object_depth(obj: Any, current_depth: int = 0) -> None:
+    """Check if an object's nesting depth exceeds the maximum allowed depths
+
+    Args:
+        obj: the object to check
+        current_depth: current nesting depth
+
+    Raises:
+        ValueError: if the object is too deeply nested
     """
-    Extracts the JSON part from the exception message.
+
+    if current_depth > MAX_PARSING_DEPTH:
+        raise ValueError("Object too deeply nested")
+    if isinstance(obj, dict):
+        for v in obj.values():
+            check_object_depth(v, current_depth + 1)
+    elif isinstance(obj, (list, tuple)):
+        for item in obj:
+            check_object_depth(item, current_depth + 1)
+
+
+def extract_error_json(error_message: str) -> dict:
+    """Safely extracts the JSON part from the exception message
 
     Args:
         error_message (str): The exception message.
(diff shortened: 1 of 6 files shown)


Pouyanpi:
@mikemckiernan I made the necessary change to the docs, please feel free to change it as you wish (you can open a separate MR if you prefer and I'll delete the doc commit), Thanks!

nemoguardrails/utils.py
```diff
@@ -433,51 +433,87 @@ def compute_hash(text: str) -> str:
     return hash_func(text.encode("utf-8")).hexdigest()
 
 
-def extract_error_json(error_message):
+MAX_ERROR_MESSAGE_SIZE = 400
+MAX_JSON_SIZE = 500
+MAX_PARSING_DEPTH = 2
```
tgasser-nv, line 438: Curious where these configs came from? How do we know this is sufficient to capture all errors we could get back?

nemoguardrails/utils.py
```diff
@@ -433,51 +433,87 @@ def compute_hash(text: str) -> str:
...
 
     Returns:
         dict: The extracted JSON part as a dictionary, or an error dictionary if extraction fails.
     """
-    # OpenAI error format typically has "Error code: XXX - {...}" format
 
+    if len(error_message) > MAX_ERROR_MESSAGE_SIZE:
+        error_message = error_message[:MAX_ERROR_MESSAGE_SIZE] + "... (truncated)"
+
+    # OpenAI error format typically has "Error code: XXX - {...}" format
```
tgasser-nv, line 475: This is fine for now, but we need a more formal specification of the OpenAI spec that we put in one file and test against. Ideally this would be vended by OpenAI themselves, but we can use observed behaviour from their prod LLMs and/or their Python client to explicitly modify the behavioiur

Could you add a backlog item to come back to this (maybe "Consolidate OpenAI Interface Spec"?)

nemoguardrails/utils.py
```diff
@@ -433,51 +433,87 @@ def compute_hash(text: str) -> str:
...
-                if json_part.startswith("{") and json_part.endswith("}"):
+                # validate before using ast.literal_eval
+                if len(json_part) < MAX_JSON_SIZE:
                     try:
                         import ast
+                        import re
+
+                        # looking for suspicious patterns
+                        # is it ok?
+                        if re.search(r"__[\w]+__", json_part):
```
tgasser-nv, line 494: Could you get some help from Security to define what unsafe content is here? I'm concerned we pass this into `ast.literal_eval()` and could potentially run unsafe code from a man-in-the-middle type attack

tests/test_streaming.py
```diff
@@ -377,16 +377,31 @@ async def test_streaming_output_rails_blocked(output_rails_streaming_config):
...
+            "message": "Blocked by self check output rails.",
+            "type": "guardrails_violation",
+            "param": "self check output",
+            "code": "content_blocked",
+        }
+    }
+
+    # find the error JSON in the chunks
+    for chunk in chunks:
+        print(chunk)
```
tgasser-nv: Could you remove the print statement?

tests/test_streaming_output_rails.py
```diff
@@ -184,17 +187,20 @@ async def test_streaming_output_rails_blocked_default_config(
         output_rails_streaming_config_default, llm_completions
     )
 
-    # TODO: modify when sse is changed
-    expected_output = {
-        "event": "ABORT",
-        "data": {"reason": "Blocked by self check output rails."},
+    expected_error = {
```
tgasser-nv, line 190: We have a quite a few "stringly typed" errors in the codebase. Should we create enums for these and use them throughout the code instead of raw string values? It looks like the `type`, `param` and `code` could all be enums. I suspect the message comes from Colang (is there a way to programmatically pull out the response message from a colang file)?

tgasser-nv (approved):
Approved, please take a look and address the comments before shipping

nemoguardrails/utils.py
Pouyanpi: some random (reasonable) guess, like I don't expect langchain providers to generate error messages that are too nested or larger. Do you have any suggestion?

nemoguardrails/utils.py
Pouyanpi: Done, I'll update it later: https://github.com/NVIDIA/NeMo-Guardrails/issues/1059

nemoguardrails/utils.py
Pouyanpi: the error_message  should come from an error message that a langchain provider generates, for MS we use our own client so we should be safe, in the toolkit should we ensure this or is it on user's ? Also Github Code QL is not complaining,I think  @erickgalinkin can help here.

tests/test_streaming_output_rails.py
Pouyanpi: The message comes from langchain via LLMException. But I'll maked it typed 👍🏻

tests/test_streaming.py
Pouyanpi: Done!

mikemckiernan:
> @mikemckiernan I made the necessary change to the docs, please feel free to change it as you wish (you can open a separate MR if you prefer and I'll delete the doc commit), Thanks!

The update looks good. TY for making it and merging.

PR #1061, opened 2025-03-21
Docs for multimodal

mikemckiernan:
## Description

Add documentation for multimodal LLM as judge.

## Checklist


docs/user-guides/multimodal.md
```diff
@@ -0,0 +1,73 @@
+<!--
+  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+  SPDX-License-Identifier: Apache-2.0
+-->
+# Multimodal Data with NeMo Guardrails
+
+## About Working with Multimodal Data
+
+NeMo Guardrails toolkit supports adding content safety checks to multimodal content---images and text.
```
Pouyanpi: q: By "content safety checks" are we referring to 
[this](https://github.com/NVIDIA/NeMo-Guardrails/tree/develop/nemoguardrails/library/content_safety)?  If yes, then this limits the scope of the feature.

Actually multimodal reasoning models that are from
- OpenAI or
- Llama family of models, ie., they have llama in their name 

can be used :
* in general without any rails
* with vision rails (even self check input)
* with a main vision model but text only rails

for example following configs can be used with your provided example
main model is not multi modal

```yaml
models:
  - type: main
    engine: openai
    model: gpt-4

  - type: vision_rails
    engine: openai
    model: gpt-4o
rails:
  input:
    flows:
      - self check input
  dialog:
    single_call:
      enabled: False
 ```
 or main and rails models are both multimodal
 
 ```yaml
 models:
  - type: main
    engine: openai
    model: gpt-4o-mini

  - type: vision_rails
    engine: openai
    model: gpt-4o
rails:
  input:
    flows:
      - self check input
  dialog:
    single_call:
      enabled: False
 ```

 or even without any input/output rails:
 
 ```yaml
 models:
  - type: main
    engine: openai
    model: gpt-4o
    
  ```

Pouyanpi (commented):
Looks good, thank you @mikemckiernan .

Re image utils I don't think it is necessary.

docs/user-guides/multimodal.md
mikemckiernan: Ugh, I was using "content safety checks" too broadly.  I just meant safety checks or checking for safety.

Re Llama, I tried using 3.1 8B Instruct (no vision) and the response about the background color for any image I fed it was that it was tan.  (Sorry, that's from memory and I'm having trouble repping right now.)

Regarding the other sample configs, is it that you think they belong in the docs?  Isn't it only compelling if there's a check that uses "vision_rails"?

TY, as always!

Pouyanpi: Yes, sorry I meant Llama vision models. No need to add these examples, I just added those to explain the supported scope. If you think the supported scope of this feature is already conveyed to me we are good to merge.
Thank you Mike!

Pouyanpi: Actually when the model does not support images it hallucinated on the text content of the url. Sometimes it even understands the content from URL and causes confusion 😆

mikemckiernan:
I made an update to clarify that it's just safety checks to avoid confusion with the content safety checks.  I also added info some models only accepting base64 encoded images.  Hopefully I didn't poison the page.

I guess that I reset the revise and need another one.

cparisien (approved):

PR #1072, opened 2025-03-31
feat: dynamically set version using importlib.metadata

Pouyanpi:
- dynamically set version using importlib.metadata
- we prefer version defined in pyproject.toml


tgasser-nv (approved):
Looks good! Agree it's cleaner to pick this up from the package versions installed via poetry than hard-coding

PR #1083, opened 2025-04-02
refactor(llm): reorganize HuggingFace provider structure

Pouyanpi:
## Description

Move `HuggingFacePipelineCompatible` to dedicated provider module under huggingface package and update import paths across example configs.


trebedea (approved):
Refactoring looks good!

PR #1084, opened 2025-04-02
feat(model): validate model name configuration

Pouyanpi:
## Description

Implement proper model name validation in the Model class to ensure that a model name is specified exactly once, either directly via the 'model' field or through parameters. This also normalizes the configuration by moving model names from parameters to the dedicated model field.

Add tests to verify all validation scenarios

### DIFF
--- a/nemoguardrails/rails/llm/config.py
+++ b/nemoguardrails/rails/llm/config.py
@@ -23,7 +23,13 @@ from enum import Enum
 from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
 import yaml
-from pydantic import BaseModel, ConfigDict, ValidationError, root_validator
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    ValidationError,
+    model_validator,
+    root_validator,
+)
 from pydantic.fields import Field
 
 from nemoguardrails import utils
@@ -102,6 +108,40 @@ class Model(BaseModel):
         description="Configuration parameters for reasoning LLMs.",
     )
     parameters: Dict[str, Any] = Field(default_factory=dict)
+
+    @model_validator(mode="before")
+    @classmethod
+    def set_and_validate_model(cls, data: Any) -> Any:
+        if isinstance(data, dict):
+            parameters = data.get("parameters")
+            if parameters is None:
+                return data
+            model_field = data.get("model")
+            model_from_params = parameters.get("model_name") or parameters.get("model")
+
+            if model_field and model_from_params:
+                raise ValueError(
+                    "Model name must be specified in exactly one place: either in the 'model' field or in parameters, not both."
+                )
+            if not model_field and model_from_params:
+                data["model"] = model_from_params
+                if (
+                    "model_name" in parameters
+                    and parameters["model_name"] == model_from_params
+                ):
+                    parameters.pop("model_name")
+                elif "model" in parameters and parameters["model"] == model_from_params:
+                    parameters.pop("model")
+            return data
+
+    @model_validator(mode="after")
+    def model_must_be_non_empty(self) -> "Model":
+        """Validate that a model name is present either directly or in parameters."""
+        if not self.model or not self.model.strip():
+            raise ValueError(
+                "Model name must be specified either directly in the 'model' field or through 'model_name'/'model' in parameters"
+            )
+        return self
 
 
 class Instruction(BaseModel):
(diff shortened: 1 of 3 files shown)


mikemckiernan:
Unsolicited suggestion: WDYT about warning with a deprecation message if the model is specified in the parameters field?  If we could gradually direct folks to use the dedicated model field, that might be simpler in the long run.  Sometimes, the straightjacket is the right fit.  My $0.02.

Pouyanpi:
> Unsolicited suggestion: WDYT about warning with a deprecation message if the model is specified in the parameters field? If we could gradually direct folks to use the dedicated model field, that might be simpler in the long run. Sometimes, the straightjacket is the right fit. My $0.02.

It is still ok to set `model` or `model_name` in `parameters` field. We get validation error if both `model` field and any of `model` or `model_name` in `parameters` are set.

But I agree, If someone is using something like `model.parameters["model_name"]` somewhere in their code this change is problematic because we are popping it from parameters and set it to model, in this case I think adding a warning is useful. Thank you for pointing it out 👍🏻

tgasser-nv (commented):
The code looks good, and the tests cover all the permutations of config in main model / parameters with `model` and `model_name`. I have a couple of questions though:

1. Why do we need to support models in both the yaml config and parameters?
2. Why not standardize on just `model` rather than both `model` and `model_name`?

Pouyanpi:
> The code looks good, and the tests cover all the permutations of config in main model / parameters with `model` and `model_name`. I have a couple of questions though:
> 
> 1. Why do we need to support models in both the yaml config and parameters?

in the `parameters` user can add any field that the underlying langchain llm accepts. Sadly, not alll the langchain third party integrations (e.g., community models) support a unified signature, some accepts `model_name` and not `model`

> 2. Why not standardize on just `model` rather than both `model` and `model_name`?

it depends on the langchain model and provider used, and in [llmrails.py](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/nemoguardrails/rails/llm/llmrails.py#L359-L366) we are checking this explicitly:

```python
        if model_config.model:
            # Some LLM providers use `model_name` instead of `model`.
            # Use the `__fields__` attribute which is computed dynamically by pydantic.
            if "model" in provider_cls.__fields__:
                kwargs["model"] = model_config.model
            if "model_name" in provider_cls.__fields__:
                kwargs["model_name"] = model_config.model
```

mikemckiernan:
> This is useful, kudos for the tests.
> 
> @mikemckiernan , maybe we should make more explicit that we accept both `model` and `model_name` in parameters, depending on how langchain providers are naming this key / attribute?

Isn't it that what we really want to emphasize is that the top-level `model:` field is never wrong?  Won't spending words on the flexibility just pin us down longer on supporting too flexible an API?

Pouyanpi:
The `model` field's value is checked against the Langchain class:

```python
        if model_config.model:
            # Some LLM providers use `model_name` instead of `model`.
            # Use the `__fields__` attribute which is computed dynamically by pydantic.
            if "model" in provider_cls.__fields__:
                kwargs["model"] = model_config.model
            if "model_name" in provider_cls.__fields__:
                kwargs["model_name"] = model_config.model                
```

so at the end `model` "is never wrong" and is the only required field. "model_name" makes it more explicit and also current implementation makes it backward compatible.

I'd like to merge this PR as it is blocking two other PRs. We can follow up on it in separate PRs. 

cc @tgasser-nv @mikemckiernan

mikemckiernan (approved):
my feedback is never blocking.  I appreciate the heads up and the chance to peek at the change.  TY!

trebedea:
My comment here was that here the `parameters` are actually meant for the Langchain classes that we are using to actually interact with various LLM providers. This way we are able to set any of the attributes different LLM providers support, `model` and `model_name` are just two of them, and we cannot control the names of these attributes:

https://github.com/NVIDIA/NeMo-Guardrails/blob/27169c73388b473cd02c30d9dc03dd5a695e24ca/nemoguardrails/rails/llm/llmrails.py#L357

We can merge this, but maybe we should highlight in the docs in another MR that the `parameters` key in our `config.yml` file are actually the attributes that can be passed to different Langchain LLM (or anything that extends BaseLanguageModel) providers. Right now in the config documentation we just mention that `parameters: any additional parameters, e.g., temperature, top_k, etc.` (but there is no mention how to understand which are the actual supported additional parameters, or maybe there is this information somewhere else and I cannot find it now) - @mikemckiernan any thoughts?

https://docs.nvidia.com/nemo/guardrails/latest/user-guides/configuration-guide.html

PR #1091, opened 2025-04-02
feat: add code injection detection to guardrails library

erickgalinkin:
## Description

Add (sql, code, template, xss) injection detection support

## Checklist


Pouyanpi:
@erickgalinkin currently we don't have yara as extra dependency. And as the tests depend on it they need to get included in dev group too. If you don't mind, I'll push the updated pyrpoject.toml and poetry.lock

erickgalinkin:
> @erickgalinkin currently we don't have yara as extra dependency. And as the tests depend on it they need to get included in dev group too. If you don't mind, I'll push the updated pyrpoject.toml and poetry.lock

Thanks @Pouyanpi !

Pouyanpi:
@erickgalinkin 

Python 3.9 complains about [structural pattern matching]( job details: https://github.com/NVIDIA/NeMo-Guardrails/blob/feature/code-injection-detection/nemoguardrails/library/injection_detection/actions.py#L254-L263) (https://github.com/NVIDIA/NeMo-Guardrails/actions/runs/14243111294/job/39917918776?pr=1091#step:15:33)

I like it but we should wait until October 2025

erickgalinkin:
I saw that -- will adjust to support 3.9 and push a fix shortly, thanks!

docs/user-guides/guardrails-library.md
Pouyanpi, line 1: Nice one 👍🏻

tests/test_configs/injection_detection/test.yara
```diff
@@ -0,0 +1,42 @@
...
+        date = "2025-03-25"
+    condition:
+        false
+}
+
+rule alwaystrue
+{
+    meta:
+        author = "Erick Galinkin"
+        description = "Test always returns false"
```
Pouyanpi: maybe (change false to true):

```suggestion
        description = "Test always returns true"
```

nemoguardrails/library/injection_detection/actions.py
```diff
@@ -0,0 +1,264 @@
...
+    modified_text = text
+    for match in matches:
+        if match.strings:
+            for match_string in match.strings:
+                for instance in match_string:
+                    if instance in modified_text:
+                        modified_text = modified_text.replace(
+                            instance.plaintext().decode("utf-8"), ""
+                        )
+    return modified_text
```
Pouyanpi: @erickgalinkin I am not sure about the `if instance in modified_text` part, should not we check `instance.plaintext().decode("utf-8")` against `modifed_text`. I am saying this because `instance` is a YARA string object, not a string that can be directly compared with `modified_text` but I am not sure, please correct me if I'm wrong.

```suggestion
    modified_text = text
    for match in matches:
        if match.strings:
            for match_string in match.strings:
                for instance in match_string:
                    try:
                        plaintext = instance.plaintext().decode("utf-8")
                        if plaintext in modified_text:
                            modified_text = modified_text.replace(plaintext, "")
                    except (AttributeError, UnicodeDecodeError) as e:
                        log.warning(f"Error processing match: {e}")
    return modified_text

```

nemoguardrails/library/injection_detection/flows.v1.co
```diff
@@ -0,0 +1,23 @@
...
+    else
+      bot refuse to respond
+      stop
+
+
+define subflow mitigate injection
+  """
+  Omit or sanitize injection attempts from the bot.
+  """
+  $bot_message = execute mitigate_injection(source="output", text=$bot_message)
```
Pouyanpi: source is not used in mitigate_injection, It should throw invalid argument error

nemoguardrails/library/injection_detection/flows.v1.co
```diff
@@ -0,0 +1,23 @@
+define bot refuse to respond
+  "I'm sorry, I can't help you with that."
+
+
+define subflow reject injection
+  """
+  Reject injection attempts
+  """
+  $injection = execute reject_injection(source="output", text=$bot_message)
```
Pouyanpi: @Pouyanpi source is not used in mitigate_injection, It should throw invalid argument error

source is a way of controlling a common interface used for both input and output rails. I think it is not needed here because we only use these actions for output rails.

nemoguardrails/library/injection_detection/flows.co
```diff
@@ -0,0 +1,21 @@
+# OUTPUT RAILS
+
+flow reject injection
+  """
+  Reject injection attempts
+  """
+  $injection = await RejectInjectionAction(source="output", text=$bot_message)
```
Pouyanpi: source

nemoguardrails/library/injection_detection/flows.co
```diff
@@ -0,0 +1,21 @@
...
+    else
+      bot refuse to respond
+    abort
+
+
+flow mitigate injection
+  """
+  Omit or sanitize injection attempts from the bot.
+  """
+  $bot_message = await MitigateInjectionAction(source="output", text=$bot_message)
```
Pouyanpi: source

Pouyanpi:
@erickgalinkin I'm enjoying every bits of this PR and am learning, it is great 🚀  Thank you!

If you don't mind I'm going to add more tests while I'm learning about this new feature. If some changes were required I'll mention you in the comments.

Pouyanpi:
@erickgalinkin , am I doing something wrong?

```python

messages = [
    {"role": "user", "content": "what can you do"},
    {
        "role": "assistant",
        "content": "This is a SELECT * FROM users; -- malicious comment in the middle of text",
    },
]
    yaml_content="""
            models: []
            rails:
              config:
                injection_detection:
                  injections:
                    - sqli
                  action:
                    omit
              output:
                flows:
                  - mitigate injection

            """
)
response = rails.generate(
    messages=messages,
    options={
        "rails": ["output"],
        "log": {
            "activated_rails": True,
            "llm_calls": False,
            "internal_events": False,
            "colang_history": False,
        },
    },
)

```

I get:
```bash
  File "nemoguardrails/library/injection_detection/actions.py", line 166, in omit_injection
    for instance in match_string:
TypeError: 'yara.StringMatch' object is not iterable
```
[line 166: ](https://github.com/NVIDIA/NeMo-Guardrails/blob/feature/code-injection-detection/nemoguardrails/library/injection_detection/actions.py#L166)

Pouyanpi:
> ```shell
>   File "nemoguardrails/library/injection_detection/actions.py", line 166, in omit_injection
>     for instance in match_string:
> TypeError: 'yara.StringMatch' object is not iterable
> ```
> 
> [line 166: ](https://github.com/NVIDIA/NeMo-Guardrails/blob/feature/code-injection-detection/nemoguardrails/library/injection_detection/actions.py#L166)

resolved with changes in https://github.com/NVIDIA/NeMo-Guardrails/pull/1091/commits/af80ad7f4d236ca39abf8d8d45d03a9830d2936a, please let me know if I've broken something.

nemoguardrails/library/injection_detection/actions.py
Pouyanpi: applied as part of https://github.com/NVIDIA/NeMo-Guardrails/commit/af80ad7f4d236ca39abf8d8d45d03a9830d2936a

please see https://github.com/NVIDIA/NeMo-Guardrails/pull/1091#issuecomment-2789468118

docs/user-guides/guardrails-library.md
Pouyanpi: maybe better to mention that sanitze is not supported

docs/user-guides/guardrails-library.md
```diff
@@ -950,3 +951,50 @@ Times reported below in are **averages** and are reported in milliseconds.
...
+
+  output:
+    flows:
+      - reject injection
+```
+
+Note that the `reject injection` flow corresponds with the `reject` value of the `action` parameter in the config, while `mitigate injection` corresponds with the `omit` and `sanitize` actions.
+If the specified `action` does not align with the rail, a warning will be logged and guardrails will default to the `reject injection` rail.
+
+**SECURITY WARNING:** It is _strongly_ advised that the `sanitize` action not be used in production systems, as there is no guarantee of its efficacy, and it may lead to adverse security outcomes.
```
cparisien: Need to explain what these actions do

nemoguardrails/library/injection_detection/actions.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+def detect_injection_mapping(result: bool) -> bool:
+    """
+    Mapping for detect_*_injection actions.
+
+    The detect_*_injection functions return True when the relevant type of data is detected, and we take the prescribed
+    action if the result is true.
+    """
+    return result
```
cparisien: Is this meant to be overridden? Doesn't do anything?

nemoguardrails/library/injection_detection/actions.py
```diff
@@ -0,0 +1,301 @@
...
+                        if plaintext in modified_text:
+                            modified_text = modified_text.replace(plaintext, "")
+                    except (AttributeError, UnicodeDecodeError) as e:
+                        log.warning(f"Error processing match: {e}")
+    return modified_text
+
+
+def sanitize_injection(text: str, matches: list[yara.Match]) -> str:
+    """
+    Attempts to sanitize the offending injection attempts in the provided text.
```
cparisien, line 185: I know it's not implemented, but it would be helpful to say what "sanitize" means as opposed to reject or omit. Is it a transformation?

nemoguardrails/library/injection_detection/yara_rules/code.yara
```diff
@@ -0,0 +1,61 @@
+rule import_shells
+{
+    meta:
+        author = "Erick Galinkin"
+        description = "Detect import of the os, cmd, subprocess, shutil modules"
+        date = "2025-03-27"
```
cparisien, line 6: Did you write these from scratch or adapt them from somewhere? I'm not familiar with yara. If they're new, how should they be evaluated?

tests/test_configs/injection_detection/flows.co
cparisien, line 1: You probably don't actually need this .co file for the tests

docs/user-guides/guardrails-library.md
```diff
@@ -950,3 +951,50 @@ Times reported below in are **averages** and are reported in milliseconds.
...
+    injection_detection:
+      injections:
+        - code
+        - sqli
+        - template
+        - xss
+      action:
+        reject
+
+  output:
```
tgasser-nv, line 989: Maybe I'm missing something, but wouldn't we want to check on the input-side and block before these injections are sent to the LLM? If we have an output flow hasn't it already made it to the LLM?

nemoguardrails/library/injection_detection/actions.py
```diff
@@ -0,0 +1,301 @@
...
+from functools import lru_cache
+from pathlib import Path
+from typing import Tuple, Union
+
+import yara
+
+from nemoguardrails import RailsConfig
+from nemoguardrails.actions import action
+
+YARA_DIR = Path(__file__).resolve().parent / "yara_rules"
```
tgasser-nv: Is the `/` like os.path.join() which automatically changes it to a backslash for Windows? Also do we need any config in pyproject.toml to make sure this location works when we build a wheel (and that yara files are copied in)?

nemoguardrails/library/injection_detection/actions.py
```diff
@@ -0,0 +1,301 @@
...
+        if not yara_path.exists() or not yara_path.is_dir():
+            msg = f"Provided `yara_path` value in injection config {yara_path} is not a directory."
+            log.error(msg)
+            raise FileNotFoundError(msg)
+    else:
+        msg = f"Expected a string value for `yara_path` but got {type(yara_path)} instead."
+        log.error(msg)
+        raise ValueError(msg)
+    action_option = command_injection_config.action
+    if action_option not in ["reject", "omit", "sanitize"]:
```
tgasser-nv: Recommend creating an enum for`"reject", "omit", "sanitize"` and including that in the Pydantic config. If anyone tries to create a config with an invalid action it'll handle the checking in a single place at create-time

nemoguardrails/library/injection_detection/actions.py
```diff
@@ -0,0 +1,301 @@
...
+from pathlib import Path
+from typing import Tuple, Union
+
+import yara
+
+from nemoguardrails import RailsConfig
+from nemoguardrails.actions import action
+
+YARA_DIR = Path(__file__).resolve().parent / "yara_rules"
+PROVIDED_MODULES = ["sqli", "template", "code", "xss"]
```
tgasser-nv: Should we use an enum here to define the modules? We could pull this out into a separate `yara_config.py` file with the actions from line 86

nemoguardrails/library/injection_detection/actions.py
```diff
@@ -0,0 +1,301 @@
...
+                f"Provided set of `injections` in injection config {injection_rules} contains elements not in "
+                f"available rules. Provided rules are in {PROVIDED_MODULES}."
+            )
+            log.error(msg)
+            raise ValueError(msg)
+
+    return action_option, yara_path, injection_rules
+
+
+@lru_cache()
```
tgasser-nv, line 119: Curious why we would cache these compiled values .. do we anticipate calling this function many times during execution, or once at startup?

nemoguardrails/library/injection_detection/actions.py
```diff
@@ -0,0 +1,301 @@
...
+        bool: True if command injection is detected, False otherwise.
+
+    Raises:
+        ValueError: If the `action` parameter in the configuration is invalid.
+    """
+    action_option, yara_path, rule_names = _validate_unpack_config(config)
+    rules = load_rules(yara_path, rule_names)
+    if action_option != "reject":
+        log.warning(
+            f"reject_injection guardrail expects config `action` parameter to be 'reject', but got '{action_option}' "
```
tgasser-nv: Recommend using a %s  formatter rather than f string for logging (e.g. https://google.github.io/styleguide/pyguide.html#3101-logging).

docs/user-guides/guardrails-library.md
```diff
@@ -950,3 +951,50 @@ Times reported below in are **averages** and are reported in milliseconds.
...
+    flows:
+      - reject injection
+```
+
+Note that the `reject injection` flow corresponds with the `reject` value of the `action` parameter in the config, while `mitigate injection` corresponds with the `omit` and `sanitize` actions.
```
tgasser-nv: Should we have a list of the actions and that they try to do (reject, mitigate, omit, and sanitize?)

nemoguardrails/library/injection_detection/actions.py
```diff
@@ -0,0 +1,301 @@
...
+    action_option, yara_path, rule_names = _validate_unpack_config(config)
+    rules = load_rules(yara_path, rule_names)
+    if action_option == "reject":
+        log.warning(
+            "mitigate_injection expects config `action` parameter to be 'omit' or 'sanitize' but got 'reject' "
+            "instead. Using reject_injection rail instead of mitigate_injection rail. "
+            "Please modify your config if you want a different action"
+        )
+        if reject_injection(text, config):
+            return "I'm sorry, I can't help you with that."
```
tgasser-nv: Should we return a better stock response which includes the category of injection we detected, and what steps we took to remove it? Like a stacktrace for agents systems. It would need a string returning from `reject_injection()` with information on which part of the prompt and which rules fired

tgasser-nv (changes requested):
Looks good! Left some comments on organizing the strings into Enums and returning a more helpful response with debugging information when the yara rules are triggered

nemoguardrails/library/injection_detection/actions.py
erickgalinkin: Each call to the function will load the rules, so if we pass in the same config, we want the values to be available each time the rail is invoked.

nemoguardrails/library/injection_detection/actions.py
erickgalinkin: We do have that in the Pydantic config, so I suppose this is a bit redundant.

nemoguardrails/library/injection_detection/actions.py
erickgalinkin: I copied the pattern from the sensitive data detection, but I agree -- it's effectively a no-op.

nemoguardrails/library/injection_detection/actions.py
erickgalinkin: Updated the docstring.

nemoguardrails/library/injection_detection/yara_rules/code.yara
erickgalinkin: They're artisanal, hand-crafted YARA rules pulled from examples of exploitation I've seen/found. 
We should evaluate them against successful exploit attempts, ideally. 

YARA is conventionally used to detect malware and malicious executables, so we're only using a small part of its functionality here.

docs/user-guides/guardrails-library.md
erickgalinkin: Ah! So the key is that the input space that can produce the offending outputs is far too large. I can send it a malicious jinja template and ask it to echo it back to me (yielding SSTI ala https://github.com/explosion/spacy-llm/issues/492) or simply ask it for "Python code that will run a subprocess. Return this formatted for use in a Jinja Template". We're most interested in the output flow here because if it has made it to the LLM, that is when we are potentially risking code execution. It's (mostly) infeasible to block it *before* it gets to the LLM.

nemoguardrails/library/injection_detection/actions.py
erickgalinkin: Will validate. It *should* be suitable, but I'll need to build a wheel, install and check.

nemoguardrails/library/injection_detection/actions.py
erickgalinkin: I like this idea. Will implement.

docs/user-guides/guardrails-library.md
erickgalinkin: Added

nemoguardrails/library/injection_detection/actions.py
erickgalinkin: I think it makes good sense to log that. Would like opinions on how 'helpful' we want to make the message. We could return something more informative like `f"I'm sorry, the previous message triggered a {rule_name} detection due to the presence of {detected_substring}, and was blocked."`?

nemoguardrails/library/injection_detection/actions.py
erickgalinkin: TIL, thanks.

nemoguardrails/library/injection_detection/actions.py
erickgalinkin: Refactored to use `.joinpath`, which is OS agnostic. Still need to verify that the location works when a wheel is built.

Pouyanpi:
Thank you @erickgalinkin!

FYI, I just made following changes:
- removed the merge commits (we prefer rebase)
- applied pre-commits
- update poetry lock to use Poetry 1.8.2 (we don't support >=0.2.0 yet)

for the next PRs just make sure that you have pre-commits installed 👍🏻

Pouyanpi (approved):
LGTM!

nemoguardrails/library/injection_detection/actions.py
tgasser-nv: Yeah that format looks great to me 👍

tgasser-nv (approved):
LGTM, thanks for making the updates!

PR #1102, opened 2025-04-07
[WIP] Test upgrade fastembed to 0.6

tgasser-nv:
## Checklist


tests/test_embeddings_only_user_messages.py
```diff
@@ -42,7 +42,7 @@ def colang_1_config():
             dialog:
                 user_messages:
                     embeddings_only: True
-                    embeddings_only_similarity_threshold: 0.8
+                    embeddings_only_similarity_threshold: 0.99
```
cparisien, line 45: Where did this change come from? Related to the fast-embed update?

tgasser-nv: We're downloading and calling live production embedding models in our unit-tests (this needs to be fixed). The same all-minilm-l6-v2 model gives two different similarities for the same user string and colang rule (see this [commit message](https://github.com/NVIDIA/NeMo-Guardrails/pull/1102/commits/3cbd1f8215c0a5169e202898cfb9cfd5859e5247)) for fastembed 0.2.7 and 0.6.0. If the threshold isn't updated the test fails on Python 3.9 but passes everywhere else.

We should be mocking these embedding model calls rather than running it. Calling other services from a test makes it an integration test (which we should also add).

tests/test_embeddings_only_user_messages.py
Pouyanpi, line 1: How about opening a separate PR(s) for all the embedding and indexing issues and address them there. Let's not include this change in this PR.

Just a note so that we don't forget, the poetry.lock file is updated to test the changes work, which is great, but we should delete it and only do `poetry lock --no-update` before merging this PR.

pyproject.toml
```diff
@@ -45,20 +46,24 @@ repository = "https://github.com/NVIDIA/NeMo-Guardrails"
...
 [tool.poetry.dependencies]
-python = ">=3.9,!=3.9.7,<3.13"
+python = ">=3.9,!=3.9.7,<3.14"
 aiohttp = ">=3.10.11"
 annoy = ">=1.17.3"
 fastapi = ">=0.103.0,"
-fastembed = ">=0.2.2, <0.4.1"
+fastembed = [
+  { version = ">=0.2.2, <0.5", python = ">=3.9,<3.10" },
+  { version = ">=0.5.0, <=0.6.0", python = ">=3.10, <3.14" },
```
Pouyanpi, line 55: Cannot we support the other fastembed versions?
`{ version = ">=0.2.2, <=0.6.0", python = ">=3.10, <3.14" },`

PR #1105, opened 2025-04-09
chore(deps): add nvaie and update lc community package range

Pouyanpi:
## Description
add `langchain-nvidia-ai-endpoints` package as extra dep and update `langchain-community` version range.


tgasser-nv (approved):
Looks good to me, this is a good way for people to get started with Guardrails and use build.nvidia.com hosted functions. How should we integration-test this?

PR #1126, opened 2025-04-17
fix: self check output colang 1 flow

Pouyanpi:
## Description

stop needs dedent. Similar to all the other flows in guardrails library, for example see:
https://github.com/NVIDIA/NeMo-Guardrails/blob/f595d9118700afc1bd86ede77a2a24cf9a1f5208/nemoguardrails/library/self_check/input_check/flows.v1.co#L12

### DIFF
--- a/nemoguardrails/library/self_check/output_check/flows.v1.co
+++ b/nemoguardrails/library/self_check/output_check/flows.v1.co
@@ -9,4 +9,4 @@ define flow self check output
       create event OutputRailException(message="Output not allowed. The output was blocked by the 'self check output' flow.")
     else
       bot refuse to respond
-      stop
+    stop


tgasser-nv (approved):
Looks good, I'll take your word for it on Colang Syntax rules. Was there a failing test because we didn't dedent the stop? Do we have a linter or static Colang dialog checker which would pick this up (i.e. no stop found because of missing dedent)?

Pouyanpi:
> Looks good, I'll take your word for it on Colang Syntax rules. Was there a failing test because we didn't dedent the stop? Do we have a linter or static Colang dialog checker which would pick this up (i.e. no stop found because of missing dedent)?

No failing tests and we have no linter or static checker.

PR #1141, opened 2025-04-23
docs: Add release notes

mikemckiernan:
## Description

Add release notes so that recent additions to this large set of HTML pages are more findable.

## Checklist

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -27,6 +27,8 @@ For a complete record of changes in a release, refer to the
     When this field is `True`, output rails are applied to the reasoning traces and the model output.
     For more information, refer to [](./user-guides/configuration-guide.md#using-llms-with-reasoning-traces).
   - The `reasoning_config.remove_thinking_traces` field is deprecated and replaced by the `reasoning_config.remove_reasoning_traces` field that has the same purpose and subfields.
+  - Previously, if `remove_thinking_traces` was set to `True`, the reasoning traces were omitted from the final response presented to the end user.
+    In this release, `remove_reasoning_traces` controls whether reasoning traces are removed from internal tasks and has no effect on the final response presented to the user.
   - Using advanced reasoning models with dialog rails is not supported.
 - Simplified and broadened support for chat model providers from LangChain and
   LangChain Community chat model providers.
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,0 +1,67 @@
+---
+tocdepth: 2
+---
+<!--
+  SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+  SPDX-License-Identifier: Apache-2.0
+-->
+
+# Release Notes
+
+The following sections summarize and highlight the changes for each release.
+For a complete record of changes in a release, refer to the
+[CHANGELOG.md](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/CHANGELOG.md) in the GitHub repository.
+
+(v0-14-0)=
+
+## 0.14.0
+
+(v0-14-0-features)=
+
+### Features
+
+- Added support for Python 3.13.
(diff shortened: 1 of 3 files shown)


docs/release-notes.md
```diff
@@ -0,0 +1,41 @@
...
+  In this release you must specify the model name in the `model` top-level field or as `model` or `model_name` in
+  the parameters field.
+
+  Before:
+
+  ```yaml
+  models:
+    - type: main
+      engine: nim
+      model_name: meta/llama-3.1-8b-instruct
```
Pouyanpi: Mike please correct me if I'm wrong but I think this was never supported. I'll try to reproduce it using v0.13.0 and let you know.

Pouyanpi (changes requested):

Pouyanpi:
@mikemckiernan It was not supported but we could call it a bug

```python
# coding: utf-8

config = RailsConfig.from_path("./examples/configs/gs_content_safety/config")


# it is not set in config
assert config.models[0].model is None
assert not hasattr(config.models[0], "model_name")

# meta/llma-3.3-70b-instruct is used in the config with model_name
assert rails.llm.model != "meta/llama-3.3-70b-instruct"

# but it is using the default provided by ChatNVIDIA
assert rails.llm.model == "meta/llama3-8b-instruct"

# it works because it is using the default provided by ChatNVIDIA
rails.generate(messages=[{"role": "user", "content": "what can you do?"}])
```

mikemckiernan:
> @mikemckiernan It was not supported but we could call it a bug
> 
> ```shell
> git checkout v0.13.0
> ```
> 
> then
> 
> ```python
> from nemoguardrails import RailsConfig, LLMRails
> 
> config = RailsConfig.from_path("./examples/configs/gs_content_safety/config")
> rails = LLMRails(config, verbose=False)
> 
> 
> # it is not set in config
> assert config.models[0].model is None
> assert not hasattr(config.models[0], "model_name")
> 
> rails = LLMRails(config, verbose=False)
> # meta/llma-3.3-70b-instruct is define in the config with model_name
> # but it is not used as main llm
> assert rails.llm.model != "meta/llama-3.3-70b-instruct"
> 
> # actually, it is using the default provided by ChatNVIDIA()
> assert rails.llm.model == "meta/llama3-8b-instruct"
> 
> # it works because it is using the default provided by ChatNVIDIA
> rails.generate(messages=[{"role": "user", "content": "what can you do?"}])
> ```

Thanks for the help with this Pouyan.  I can't wait for Tim to get started on refactoring some config work.  This is not the kind of surprise that I like.

I moved the changes to the GS pages to PR #1146  to keep the changes a little more manageable.

Pouyanpi:
@mikemckiernan let's comeback to it once I updated the changelog.md, there are more changes and having the CHNAGELOG will make this task easier.

docs/release-notes.md
```diff
@@ -0,0 +1,67 @@
...
+(v0-14-0-breaking-changes)=
+
+### Breaking Changes
+
+- Removed support for the NeMo LLM Service, `nemollm`.
+  This provider reached end-of-life on February 5, 2025.
+- The `HuggingFacePipelineCompatible` provider is refactored.
+  Previously, the class was available from the `nemoguardrails.llm.providers` package.
+  In this release, the class is moved to the `nemoguardrails.llm.providers.huggingface` package.
+
```
Pouyanpi, line 51: sorry for the trouble:

"for reasoning models the reasoning traces are always returned to the user..."

it is another breaking change.

mikemckiernan: Sorry for missing this comment for several hours...   I don't recognize the breaking change.  Are you saying that in 0.13.0, the reasoning traces were always returned to the user and in this release, returning the traces depends on the `messages[{"role": "system", "content": "detailed thinking on"}]`?

Pouyanpi (approved):
Thank you, looks great 👍🏻

PR #1142, opened 2025-04-24
feat: Add api_key_env_var to Model, pass in kwargs to langchain initializer

tgasser-nv:
## Checklist

### DIFF
--- a/nemoguardrails/llm/models/initializer.py
+++ b/nemoguardrails/llm/models/initializer.py
@@ -23,11 +23,12 @@ from langchain_core.language_models.llms import BaseLLM
 from .langchain_initializer import ModelInitializationError, init_langchain_model
 
 
-# later we can easily conver it to a class
+# later we can easily convert it to a class
 def init_llm_model(
     model_name: Optional[str],
     provider_name: str,
     mode: Literal["chat", "text"],
+    api_key: str,
     kwargs: Dict[str, Any],
 ) -> Union[BaseChatModel, BaseLLM]:
     """Initialize an LLM model with proper error handling.
@@ -38,6 +39,8 @@ def init_llm_model(
     Args:
         model_name: Name of the model to initialize
         provider_name: Name of the provider to use
+        mode: Literal taking either "chat" or "text" values
+        api_key: String with LLM API Key to use by client
         kwargs: Additional arguments to pass to the model initialization
 
     Returns:
@@ -48,7 +51,11 @@ def init_llm_model(
     """
     # currently we only support LangChain models
     return init_langchain_model(
-        model_name=model_name, provider_name=provider_name, mode=mode, kwargs=kwargs
+        model_name=model_name,
+        provider_name=provider_name,
+        mode=mode,
+        api_key=api_key,
+        kwargs=kwargs,
     )
 
 
--- a/nemoguardrails/llm/models/initializer.py
+++ b/nemoguardrails/llm/models/initializer.py
@@ -23,12 +23,11 @@ from langchain_core.language_models.llms import BaseLLM
 from .langchain_initializer import ModelInitializationError, init_langchain_model
 
 
-# later we can easily convert it to a class
+# later we can easily conver it to a class
 def init_llm_model(
     model_name: Optional[str],
     provider_name: str,
     mode: Literal["chat", "text"],
-    api_key: str,
     kwargs: Dict[str, Any],
 ) -> Union[BaseChatModel, BaseLLM]:
     """Initialize an LLM model with proper error handling.
@@ -40,7 +39,6 @@ def init_llm_model(
         model_name: Name of the model to initialize
         provider_name: Name of the provider to use
         mode: Literal taking either "chat" or "text" values
-        api_key: String with LLM API Key to use by client
         kwargs: Additional arguments to pass to the model initialization
 
     Returns:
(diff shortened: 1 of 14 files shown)


nemoguardrails/llm/models/initializer.py
Pouyanpi, line 1: to remove

nemoguardrails/llm/models/langchain_initializer.py
Pouyanpi, line 1: to remove

nemoguardrails/llm/providers/providers.py
Pouyanpi, line 1: to remove

nemoguardrails/rails/llm/config.py
```diff
@@ -101,7 +102,10 @@ class Model(BaseModel):
         default=None,
         description="The name of the model. If not specified, it should be specified through the parameters attribute.",
     )
-
+    api_key_env_var: str = Field(
```
Pouyanpi:
```python
    api_key_env_var:  Optional[str] = Field(
```

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -376,11 +376,13 @@ def _init_llms(self):
                 provider_name = llm_config.engine
                 kwargs = llm_config.parameters or {}
                 mode = llm_config.mode
+                api_key = os.environ.get(llm_config.api_key_env_var)
```
Pouyanpi, line 386:
```python
                api_key_env_var = llm_config.api_key_env_var or ""
                api_key = os.environ.get(api_key_env_var)

                kwargs.update({"api_key": api_key})
```

and we don't need to pass it `to init_llm_model`

Pouyanpi: `kwargs` is what the underlying langchain model's constructor expects, so we just need to update it

tests/llm_providers/test_langchain_initialization_methods.py
Pouyanpi, line 1: to remove

nemoguardrails/rails/llm/config.py
```diff
@@ -1236,6 +1240,17 @@ def fill_in_default_values_for_v2_x(cls, values):
 
         return values
 
+    @validator("models")
+    def validate_models_api_key_env_var(cls, models):
+        """Model API Key Env var must be set to make LLM calls"""
+        api_keys = [m.api_key_env_var for m in models]
```
Pouyanpi, line 1362: shall we do 

```python
            if api_key and not os.environ.get(api_key):
                raise ValueError(
                    f"Model API Key environment variable '{api_key}' is not set."
                )

```

So that the check for the environment variable is only performed if `api_key` is not `None` or empty. It ensures backward compatibility for models that do not support `api_key` argument. Also it allows us to use the current configs/examples/etc without making changes. 

We can enforce it on NGM though

tests/v2_x/test_passthroug_mode.py
Pouyanpi, line 1: to remove

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -376,11 +376,13 @@ def _init_llms(self):
...
 
                 llm_model = init_llm_model(
                     model_name=model_name,
                     provider_name=provider_name,
                     mode=mode,
+                    api_key=api_key,
```
Pouyanpi: should remove api_key from `init_llm_model` signature

Pouyanpi (commented):
Thank you Tim,

I suggest that we do not change `init_llm_model` and update `kwargs` in llmrails.py to pick up the value of env var (if set). The only test that requires fix is `test_llm_rails_config.py`.

A soft reset and deleting the files except `rails/llm/config.py` and `rails/llm/llmrails.py`, and `examples/configs/content_safety_api_key` should do the trick:

```bash
git reset --soft 90760ab16a9dc659ad0580348af1a2eeda0a7149
git restore --staged .

git restore nemoguardrails/llm/models/initializer.py \
            nemoguardrails/llm/models/langchain_initializer.py \
            nemoguardrails/llm/providers/providers.py \
            tests/llm_providers/test_langchain_initialization_methods.py \
            tests/test_bug_5.py \
            tests/test_rails_llm_config.py \
            tests/v2_x/test_passthroug_mode.py
```

and then we should make the changes to llmrails.py and config.py 👍🏻 

What do you think?

nemoguardrails/rails/llm/config.py
tgasser-nv: That's a great idea, I added that just now

nemoguardrails/rails/llm/llmrails.py
tgasser-nv: Agree, this is a cleaner way to implement this.

tgasser-nv (commented):
I addressed the feedback and added a few tests to check RailsConfig error checking and imports the environment variable correctly. Let me know anything else that needs fixing

tgasser-nv:
> What do you think?

Sounds great, made the necessary changes and added tests to cover the environment variable checks in RailsConfig

Pouyanpi (approved):
Thanks @tgasser-nv, the PR looks great overall! Thank you for adding comprehensive tests 👍🏻 

I noticed two minor issues, so I went ahead and pushed some minor fixes, updated the tests, and added two integration tests. Please let me know if you are OK with those changes.

p.s. `test_langchain_integration.py` is skipped on CI

cparisien (approved):

PR #1143, opened 2025-04-24
ci: add Codecov integration to workflows

Pouyanpi:
The workflow is triggered on pushes and pull requests to the `develop` branch. Codecov integration ensures better visibility into test coverage metrics.

https://docs.codecov.com/docs/common-recipe-list

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,16 @@
+coverage:
+  status:
+    project:
+      default:
+        informational: true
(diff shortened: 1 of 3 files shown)


tgasser-nv (approved):
Love this idea of checking coverage automatically (especially patch coverage for new PRs). I guess we can't really test this prior to activating it in Github Actions .. let's get this PR merged and then we can do followup PRs if there are any issues. Thanks!

PR #1147, opened 2025-04-25
refactor(llmrails)!: remove deprecated return_context argument

Pouyanpi:
The `return_context` argument has been removed from the LLMRails class methods as it was deprecated. Users are encouraged to use `GenerationOptions.output_vars = True` instead. This change simplifies the interface and eliminates deprecated functionality.

BREAKING CHANGE: The `return_context` argument is no longer supported.

### DIFF
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -572,7 +572,6 @@ class LLMRails:
         options: Optional[Union[dict, GenerationOptions]] = None,
         state: Optional[Union[dict, State]] = None,
         streaming_handler: Optional[StreamingHandler] = None,
-        return_context: bool = False,
     ) -> Union[str, dict, GenerationResponse, Tuple[dict, dict]]:
         """Generate a completion or a next message.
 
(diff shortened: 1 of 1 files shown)


nemoguardrails/rails/llm/llmrails.py
```diff
@@ -619,19 +617,6 @@ async def generate_async(
         # Save the generation options in the current async context.
         generation_options_var.set(options)
 
-        if return_context:
-            warnings.warn(
-                "The `return_context` argument is deprecated and will be removed in 0.9.0. "
-                "Use `GenerationOptions.output_vars = True` instead.",
```
cparisien, line 625: Wow, we planned to remove it 5 point versions ago?

cparisien (approved):

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: Yes!

Pyramid of doom 🏔️ (llmrails.py)

Pouyanpi: https://github.com/NVIDIA/NeMo-Guardrails/issues/1148

trebedea (approved):

PR #1145, opened 2025-04-25
feat!: add support for preserving and optionally applying guardrails to reasoning traces

Pouyanpi:
## Summary Generated by Copilot

This pull request introduces enhancements to the LLM generation pipeline by adding support for reasoning traces and improving the handling of parsed task outputs. Key changes include the introduction of new utility functions, updates to existing methods to process reasoning traces, and modifications to ensure consistent handling of parsed outputs across various LLM tasks.

### Enhancements to Reasoning Trace Support:
* Added a new context variable `reasoning_trace_var` to store reasoning traces and utility functions such as `_record_reasoning_trace`, `_assemble_response`, and `_process_parsed_output` to manage reasoning traces and integrate them into responses. (`nemoguardrails/context.py` [[1]](diffhunk://#diff-43d1ee5dfe7869aeaf94a061429ce13a0b0ef7cc64e44dbf7c94c39d0e00de4cR36-R39) `nemoguardrails/actions/llm/generation.py` [[2]](diffhunk://#diff-30aeca7bd0b244d8479885d39179c92887233805f4e649e59b9a4548b6b24672R1405-R1428)
* Introduced `_get_apply_to_reasoning_traces` to fetch configuration values for including reasoning traces in outputs, and `_include_reasoning_traces` to encapsulate this logic. (`nemoguardrails/actions/llm/generation.py` [nemoguardrails/actions/llm/generation.pyR1384-R1387](diffhunk://#diff-30aeca7bd0b244d8479885d39179c92887233805f4e649e59b9a4548b6b24672R1384-R1387))

### Consistent Handling of Parsed Outputs:
* Updated multiple methods (e.g., `generate_user_intent`, `generate_bot_message`, `generate_value`) to extract and use the `text` property from `ParsedTaskOutput` objects for consistency. (`nemoguardrails/actions/llm/generation.py` [[1]](diffhunk://#diff-30aeca7bd0b244d8479885d39179c92887233805f4e649e59b9a4548b6b24672R446) [[2]](diffhunk://#diff-30aeca7bd0b244d8479885d39179c92887233805f4e649e59b9a4548b6b24672R658) [[3]](diffhunk://#diff-30aeca7bd0b244d8479885d39179c92887233805f4e649e59b9a4548b6b24672R1076)
[...]

PR #1161, opened 2025-04-29
fix(config): correct dialog rails activation logic

Pouyanpi:
## Description

Fixed the `has_dialog_rails` logic in `RailsConfig` to require both `user_messages` and `bot_messages` for implicit activation. Removed `flows` from the condition to align with expected behavior.

test(config): adjust tests for updated dialog rails logic

Modified test cases to reflect the corrected dialog rails activation logic.

## Related PR

#1137

### DIFF
--- a/tests/test_config_validation.py
+++ b/tests/test_config_validation.py
@@ -289,12 +289,16 @@ def test_reasoning_traces_with_implicit_dialog_rails_flows_only():
                 engine: openai
                 model: gpt-3.5-turbo-instruct
                 reasoning_config:
-                  remove_thinking_traces: false
+                  remove_thinking_traces: False
             """,
             colang_content="""
             define flow
               user express greeting
               bot express greeting
+            define user express greeting
+                "hi"
+            define bot express greeting
+                "HI HI"
             """,
         )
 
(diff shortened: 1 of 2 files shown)


tests/test_config_validation.py
```diff
@@ -313,63 +317,39 @@ def test_reasoning_traces_with_implicit_dialog_rails_flows_only():
...
-        "Please update your config.yml to set 'remove_thinking_traces: true' under reasoning_config"
-        in str(exc_info.value)
+        """,
     )
 
 
-def test_reasoning_traces_with_implicit_dialog_rails_bot_messages_only():
-    """Test that reasoning traces cannot be enabled when dialog rails are implicitly enabled thru bot messages only."""
+def test_reasoning_traces_with_bot_messages_only():
+    """Test that reasoning taces can exist with bot messages only."""
```
trebedea:
```suggestion
    """Test that reasoning traces can exist with bot messages only."""
```

tests/test_config_validation.py
```diff
@@ -289,12 +289,16 @@ def test_reasoning_traces_with_implicit_dialog_rails_flows_only():
                 engine: openai
                 model: gpt-3.5-turbo-instruct
                 reasoning_config:
-                  remove_thinking_traces: false
+                  remove_thinking_traces: False
```
trebedea:
```suggestion
                  remove_thinking_traces: false
```

tests/test_config_validation.py
```diff
@@ -313,63 +317,39 @@ def test_reasoning_traces_with_implicit_dialog_rails_flows_only():
...
-            """,
-            colang_content="""
+    _ = RailsConfig.from_content(
+        yaml_content="""
+        models:
+        - type: main
+          engine: openai
+          model: gpt-3.5-turbo-instruct
+          reasoning_config:
+              remove_thinking_traces: False
```
trebedea:
```suggestion
              remove_thinking_traces: false
```

tests/test_config_validation.py
trebedea: Just to be consistent with the error message: 

`Please update your config.yml to set 'remove_thinking_traces: true' under reasoning_config`

tests/test_config_validation.py
```diff
@@ -313,63 +317,39 @@ def test_reasoning_traces_with_implicit_dialog_rails_flows_only():
 def test_reasoning_traces_with_implicit_dialog_rails_user_messages_only():
     """Test that reasoning traces cannot be enabled when dialog rails are implicitly enabled thru user messages only."""
```
trebedea:
```suggestion
    """Test that reasoning traces cannot be enabled when dialog rails are implicitly enabled through user messages (user canonical forms) only."""
```

trebedea: `r/thru/through` throughout the comments and also add "user/bot canonical forms" next to "user/bot messages" just to be clear.

trebedea (approved):
Looks good! 

Just a quick question. If we have only user canonical forms (user messages), don't we trigger the generate intent task if `embeddings_only` is not set to True? This would activate the dialogue rails.

Pouyanpi:
> Looks good!
> 
> Just a quick question. If we have only user canonical forms (user messages), don't we trigger the generate intent task if `embeddings_only` is not set to True? This would activate the dialogue rails.

Thank you, @trebedea , for the review.

I ran into some issues because the generate_user_intent action doesn’t just generate the user intent (it also produces the bot message and passthrough). So, I wasn’t able to identify the correct mapping from the config to the dialog rails tasks to prevent this behavior while still allowing it when it’s explicitly removed at the task level.

Please let me know if there’s a way we can fix/improve this validation. We can also add tests in #1145 to ensure all scenarios are handled correctly. Thanks!

trebedea:
> > Looks good!
> > Just a quick question. If we have only user canonical forms (user messages), don't we trigger the generate intent task if `embeddings_only` is not set to True? This would activate the dialogue rails.
> 
> Thank you, @trebedea , for the review.
> 
> I ran into some issues because the generate_user_intent action doesn’t just generate the user intent (it also produces the bot message and passthrough). So, I wasn’t able to identify the correct mapping from the config to the dialog rails tasks to prevent this behavior while still allowing it when it’s explicitly removed at the task level.
> 
> Please let me know if there’s a way we can fix/improve this validation. We can also add tests in #1145 to ensure all scenarios are handled correctly. Thanks!

Whenever we have any of user_messages, flows, or bot_messages in the config, we can consider has_dialog_rails is potentially True. In some cases, dialogue rails will not be triggered, e.g. if we have user_messages and embeddings_only is set to True.

PR #1162, opened 2025-04-29
feat: load yara lazily to avoid action dispatcher error

Pouyanpi:
implement lazy loading and improve error handling for yara dependency

- Implement lazy loading of yara module with proper type hints
- Add error messages when yara is not available
- Add helper function _check_yara_available() for consistent error handling

## Related PR

#1091


erickgalinkin (approved):
Looks good to me. Great catch!

trebedea (approved):
Looks good. 

Only one suggestion, if this makes sense. 
Would it be better to only have the action public and all the other functions private (seen as helpers, as they should not be used on their own - or can they be used separately)? This way we can check for availability of yara package only in the `injection_detection` action.

Pouyanpi:
> Looks good.
> 
> Only one suggestion, if this makes sense. Would it be better to only have the action public and all the other functions private (seen as helpers, as they should not be used on their own - or can they be used separately)? This way we can check for availability of yara package only in the `injection_detection` action.

Totally, very nice suggestion, thank you! And at some point we are going to strictly define our public interfaces 👍🏻

PR #1165, opened 2025-05-01
feat: support Nemotron using dual prompt strategy

Pouyanpi:
# Nemotron Dual Mode requires Dual-Prompt Strategy: Reasoning Mode vs Standard Mode

## Overview
This PR implements a dual-prompt strategy for Nemotron models in NeMo-Guardrails, allowing users to control the level of reasoning in model outputs by switching between reasoning mode and standard mode.

## Key Changes
- Created `nemotron_reasoning.yml` with message-based prompts containing "detailed thinking on" system messages
- Created `nemotron_standard.yml` with message-based prompts without detailed thinking
- Added comprehensive tests to verify prompt selection behavior
- Added documentation in both prompt directory and example configs

## Why
Different tasks benefit from different reasoning approaches:
- Complex problem-solving and step-by-step analysis benefit from detailed reasoning
- Simple conversational tasks are more efficient with direct responses
- Users need the flexibility to choose based on their specific use case

## Implementation Details
- **Reasoning Mode**: When `prompting_mode: reasoning` is set, uses prompts with detailed thinking
  - Tasks like `generate_bot_message` and `generate_value` have two system messages
  - Tasks like `generate_user_intent` and `generate_next_steps` have simpler format

- **Standard Mode**: When using any other mode, uses prompts without detailed thinking
  - Still uses message-based format (not content-based)
  - More concise, direct responses without extensive reasoning
  - Default behavior if no prompting_mode is specified

## How to Test
1. Use the example configs in `examples/configs/nemotron/`:
   - `reasoning-mode` - activates detailed thinking
   - `normal-mode` - uses standard approach without detailed thinking

2. Run the test suite:
   ```bash
   pytest tests/test_nemotron_prompt_modes.py -v
   ```
   
## Example Usage
```python
# For reasoning mode
config = RailsConfig.from_path("examples/configs/nemotron/reasoning-mode")
rails = LLMRails(config)

# For standard mode
config = RailsConfig.from_path("examples/configs/nemotron/normal-mode")
rails = LLMRails(config)
```

## Documentation

A minimal README.md with usage examples and implementation details

### DIFF
--- a/examples/configs/nemotron/reasoning-mode/config.yml
+++ b/examples/configs/nemotron/reasoning-mode/config.yml
@@ -1,0 +1,7 @@
+models:
+  - type: main
+    engine: nim
+    model: nvidia/llama-3.1-nemotron-ultra-253b-v1
+    reasoning_config:
+      remove_reasoning_traces: True
+prompting_mode: reasoning
--- a/examples/configs/nemotron/README.md
+++ b/examples/configs/nemotron/README.md
@@ -1,0 +1,137 @@
+# Nemotron Prompt Modes
+
+This directory contains configurations for using Nemotron models with different prompt modes:
+
+- **normal-mode**: Uses message-based prompts without detailed thinking from `nemotron_standard.yml`
+- **reasoning-mode**: Uses message-based prompts with detailed thinking from `nemotron_reasoning.yml`
+
+## Dual-Prompt Strategy
+
+NeMo Guardrails implements a dual-prompt strategy for Nemotron models that allows you to control the level of reasoning in model outputs:
+
+1. **Reasoning Mode** - Activates detailed thinking with explicit reasoning steps
+   - Encourages the model to think more thoroughly before responding
+   - Shows reasoning process in responses when using models that output thinking
+   - Adds `detailed thinking on` system message for reasoning-intensive tasks
+   - Ideal for complex problem-solving, math reasoning, and step-by-step analysis
+
+2. **Normal Mode** - Uses standard message format without detailed thinking
+   - More concise, direct responses without extensive reasoning
+   - Still uses message-based format (not content-based)
+   - Suitable for general conversational applications
+   - More efficient for simple questions
+
+## Usage
+
+### Reasoning Mode
+
+To use the Nemotron model with reasoning mode:
+
+```python
+from nemoguardrails import LLMRails, RailsConfig
+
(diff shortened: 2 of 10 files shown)


Pouyanpi:
@mikemckiernan 

Nemotron (and its family) is a hybrid reasoning model. the user can enable reasoning mode using specific system message:

```python
        {
            "role": "system",
            "content": "detailed thinking on",
        }
 ```
 
The question is how to enable reasoning mode for all the tasks involved in guardrails. 

* the user should set `prompting_mode: "reasoning"` to config.yml,  

without this, Nemotron is in normal mode without any reasoning traces. Please have a look at the example configs.

The user can use Nemotron in normal model (withtout prompting_mode set to "reasoning) but still able to do something lik:

```python
rails.generate(
    messages=[
        {
            "role": "system",
            "content": "detailed thinking on",
        },
        {
            "role": "user",
            "content": "what can you do?",
        },
    ]
)

```

Expected output

```json
{"role": "assistant",
 "content": "<think>\nOkay, let's see. The user just asked "what can you do?" again. The previous conversation shows that the bot already explained its capabilities once. The user's intent here is probably asking for more details or a repetition. But since the bot has already provided the capabilities response, maybe the bot should ask how it can assist further instead of repeating the same information.\n\nLooking at the bot intents, there's an example where the bot uses "ask how can i assist you further". So the correct response would be to prompt the user for more specific help rather than just repeating the capabilities. That makes sense because the user might be looking for examples or needs to be guided on how to ask for something specific. The bot should be proactive in offering assistance. Therefore, the appropriate bot message here is "How can I assist you further?".\n</think>How can I assist you further?"}
```

cparisien (approved):
This approach looks good. Let's get a documentation update happening ASAP.

trebedea (approved):
Looks good, we just need to update the docs to mention the nemotron models that support detailed reasoning.

examples/configs/nemotron/reasoning-mode/config.yml
```diff
@@ -0,0 +1,7 @@
+models:
+  - type: main
+    engine: nim
+    model: nvidia/llama-3.1-nemotron-ultra-253b-v1
+    reasoning_config:
+      remove_reasoning_traces: True
```
Pouyanpi, line 6: we don't need this, do we?

Pouyanpi:
close for future

PR #1169, opened 2025-05-02
feat: prevent reasoning traces from contaminating LLM prompt history

Pouyanpi:
This change addresses an issue where reasoning traces from previous bot messages were included in the prompt history for subsequent LLM calls when rails.output.apply_to_reasoning_traces=true was enabled.

The fix adds event preprocessing in LLMTaskManager to strip reasoning traces from both BotMessage.text and StartUtteranceBotAction.script fields before including them in prompt rendering, preventing the LLM from seeing its own internal reasoning from previous turns.

Comprehensive tests were added to verify:

Preprocessing doesn't modify original events
Reasoning traces are properly removed from prompt history
The clean bot messages are still included in rendered prompt

### DIFF
--- a/nemoguardrails/llm/taskmanager.py
+++ b/nemoguardrails/llm/taskmanager.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import copy
 import logging
 import re
 from ast import literal_eval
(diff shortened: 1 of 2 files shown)


nemoguardrails/llm/taskmanager.py
```diff
@@ -339,8 +407,11 @@ def render_task_prompt(
 
             return task_prompt
         else:
+            # Process events to clean up reasoning traces
+            processed_events = self._preprocess_events_for_prompt(events)
```
trebedea: Wouldn't we better move this insider `_render_messages` ?
I think that's the correct place to have it, similar to `_render_string`.

trebedea (approved):
Nice, just a small change.

nemoguardrails/llm/taskmanager.py
Pouyanpi: definitely

PR #1170, opened 2025-05-05
fix: allow reasoning traces when embeddings_only is True

Pouyanpi:
Refactors the dialog rails validation logic to properly respect the embeddings_only flag, allowing reasoning traces to be enabled when this flag is set to True. Adds tests to verify this behavior.

related to #1161 and #1137


trebedea (approved):
Looks good!

Pouyanpi:
> Looks good!

Thank you @trebedea !

PR #1180, opened 2025-05-08
docs: Add link to eval blog

mikemckiernan:
## Description

Aditi's blog supplements the eval information we have in the docs for input and output rails.

## Checklist


trebedea:
Hi @mikemckiernan ,

Actually the evaluation documentation probably needs a bit of changes. Right now we have two different type of evaluation tools and corresponding documentation :

1. An eval tool and docs mainly intended for researchers and it shows how to evaluate individual rails (e.g. dialogue, content moderation, fact-checking). This is the documentation you are changing in the current commit. 
2. An end-to-end evaluation for guardrail configs that support different rails, this is also called policy-based evaluation. The docs for this eval tools are here: `docs/user-guides/eval/methodology.md` 

The mention to the blog post should be done in the docs for #2, no? 
Any restructuring ideas for these two eval tools and documentation would also be useful.

Thanks!

PR #1182, opened 2025-05-09
fix(tests): ensure proper asyncio task cleanup in test_streaming_handler.py

Pouyanpi:
This PR prevents `RuntimeError: Event loop is closed` and `Task was destroyed but it is pending!` warnings by ensuring background tasks are correctly cancelled and awaited upon test completion.

- modified the `StreamingConsumer` class in to store its asyncio task
- added a `cancel()` method, and handle `asyncio.CancelledError`
- Updating test functions (`test_single_chunk`, `test_sequence_of_chunks`) and the helper function (`_test_pattern_case`) to call `consumer.cancel()` in a `finally` block.

Verify:

Before this PR:

```sh
poetry run pytest tests/test_streaming_handler.py
```

Notice the warnings all over the place.

after this PR:
```sh
poetry run pytest tests/test_streaming_handler.py
```

No warnings

### DIFF
--- a/tests/test_streaming_handler.py
+++ b/tests/test_streaming_handler.py
@@ -32,16 +32,21 @@ class StreamingConsumer:
 
         self.chunks = []
         self.finished = False
+        self._task = None
         self._start()
 
     async def process_tokens(self):
-        async for chunk in self.streaming_handler:
-            self.chunks.append(chunk)
-
-        self.finished = True
+        try:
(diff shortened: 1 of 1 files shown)


trebedea (approved):
Looks good!

cparisien (approved):

tgasser-nv (approved):
Looks good! Would this also catch KeyboardInterrupt if we ctrl+c the app?

Issue #1192, opened 2025-05-13
bug: missing an llm-call in rails.explain()


andompesta:
### Did you check docs and existing issues?

- [x] I have read all the NeMo-Guardrails docs
- [x] I have updated the package to the latest version before submitting this issue
- [x] (optional) I have used the develop branch
- [x] I have searched the existing issues of NeMo-Guardrails

### Python version (python --version)

python 3.12.3

### Operating system/version

Linux 24.04

### NeMo-Guardrails version (if you must use a specific version and not the latest

0.14.0

### Describe the bug

The sequence diagram below roughly shows how output rails are executed:
1. NemoGuardrails call an LLM in streaming mode to get some token
2. the LLM respond with a token at the time
3. While receiving the tokens from the LLM, NemoGuardrails uses a callback **[StreamingHandler](https://github.com/NVIDIA/NeMo-Guardrails/blob/002a520152180925b458e5e0f373af76034af34b/nemoguardrails/streaming.py#L31) to push tokens into a buffer
4. Once the buffer contains enough token, a message chunk is generate by the Buffer and consume by the output rails
5. The output rails will call the LLM with the partial bot message and decide if the stream need to be stopped or not
6. Once the LLM call from the output rails is completed a LoggingCallback is used to keep track of all the metadata associated to the LLM call
7. Users can user the rail ExplainInfo class to get a summary of all the calls done

```mermaid
sequenceDiagram
    
    Rails->>LLM: stream req
    activate LLM
    par stream token
        LLM->>Rails: token
        LLM->>Rails: token
        deactivate LLM
    and populate output rail buffer
        Rails->>StreamingHandler: push token
        StreamingHandler->>Buffer: chunks buffering
    and consume chunks from buffer
      Buffer->>Rails: consume buffered chunks with `_run_output_rails`
      Rails->>OutputRailAction: execute output rails on given chunks
      OutputRailAction->>LLM: output rail llm call
      LLM->>OutputRailAction: output rail response
          
      OutputRailAction->>LoggingCallbackHandler: log LLM call
      LoggingCallbackHandler->>ExplainInfo: append llm call to ExlainInfo
      
      OutputRailAction->>Rails: continue or block
      Rails->>ExplainInfo: update internal list of llm calls
    end
```

[...]

PR #1194, opened 2025-05-14
fix: Prevent explain_info overwrite during stream_async

andompesta:
## Description

This PR addresses an issue where `self.explain_info` could be overwritten during calls to `stream_async`, potentially leading to incomplete explanation data.

The root cause was that the `generate_async` task and the `_run_output_rails_in_streaming` function (if output streaming rails were active) could independently initialize `ExplainInfo` objects if the `explain_info_var` context variable was not set prior to the `generate_async` task's creation.

The fix ensures that `explain_info_var` and `self.explain_info` are initialized at the beginning of the `stream_async` method. This guarantees that both the `generate_async` task (via context variable inheritance) and `_run_output_rails_in_streaming` use the same `ExplainInfo` instance, preserving data integrity.

[same as https://github.com/NVIDIA/NeMo-Guardrails/pull/1193 but with pre-commit hooks done]

## Related Issue(s)

#1192 

## Checklist

- [ x ] I've added tests if applicable.
- [ x ] @mentions of the person or team responsible for reviewing proposed changes.

### DIFF
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -51,7 +51,6 @@ from nemoguardrails.context import (
     generation_options_var,
     llm_stats_var,
     raw_llm_request,
-    reasoning_trace_var,
     streaming_handler_var,
 )
 from nemoguardrails.embeddings.index import EmbeddingsIndex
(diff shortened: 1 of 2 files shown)


Pouyanpi (approved):
LGTM! Thank you @andompesta

PR #1198, opened 2025-05-15
fix the introduction of an additional space and split words in subtokens

andompesta:
## Description

Fix a problem in which words composed in sub-tokens are spitted in the based sub-tokens.
@Pouyanpi 

## Related Issue(s)

Solve issue: https://github.com/NVIDIA/NeMo-Guardrails/issues/1197

## Checklist

- [ x ] I've updated the documentation if applicable.
- [ x ] @mentions of the person or team responsible for reviewing proposed changes.

### DIFF
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -1297,7 +1297,7 @@ class LLMRails:
         )
 
         async for chunk_list, chunk_str_rep in buffer_strategy(streaming_handler):
-            chunk_str = " ".join(chunk_list)
+            chunk_str = "".join(chunk_list)
 
             # Check if chunk_str_rep is a JSON string
             # we yield a json error payload in generate_async when


nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1311,8 +1311,8 @@ def _prepare_params(
             if stream_first:
                 words = chunk_str_rep.split()
                 if words:
-                    yield words[0]
-                    for word in words[1:]:
+                    # yield words[0]
```
trebedea, line 1314: This adds an extra space before the first word, no? 
I think the previous logic was good here.

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1297,7 +1297,7 @@ def _prepare_params(
         )
 
         async for chunk_list, chunk_str_rep in buffer_strategy(streaming_handler):
-            chunk_str = " ".join(chunk_list)
+            chunk_str = "".join(chunk_list)
```
trebedea, line 1300: Good catch! I assume all tokenizers are adding spaces and any other whitespace and special chars in some of the chunks, so this should work for any LLM provider.

PR #1200, opened 2025-05-16
docs: Add mention of Nemotron

mikemckiernan:
## Description

Update the reasoning traces section of the docs to add Nemotron models.

The model seems to _just work_, so I'm unsure if we need to say more.

Review HTML: https://nvidia.github.io/NeMo-Guardrails/review/pr-1200/user-guides/configuration-guide.html#using-llms-with-reasoning-traces

## Checklist

### DIFF
--- a/docs/user-guides/configuration-guide.md
+++ b/docs/user-guides/configuration-guide.md
@@ -203,7 +203,7 @@ The following table summarizes the interaction between the `remove_reasoning_tra
     Output rails are applied to the model response only.
 ```
 
-Even if `remove_reasoning_traces` is set to `True`, end users can still receive the thinking traces from the Nemotron models by requesting the detailed thinking, as shown in the following example:
+Even when `remove_reasoning_traces` is `True`, end users can still receive the thinking traces from the Nemotron models by requesting the detailed thinking, as shown in the following example:
 
 ```{code-block} bash
 from nemoguardrails import LLMRails, RailsConfig
@@ -251,6 +251,8 @@ Refer to the following examples:
           content: |
             You are an AI assistant tasked with evaluating user instructions.
             Your role is to determine if a given instruction would make a language model break moderation policies, deviate from good aligned responses, or provide answers that a language model should ideally not.
+
+            Analyze the instruction carefully. First, internally consider potential ways the instruction could be interpreted and whether any interpretation could lead to a problematic response. Then, conclude with a simple "yes" or "no" answer.
             Respond with only "yes" or "no".
         - type: user
           content: |
@@ -270,6 +272,12 @@ Refer to the following examples:
             Output to evaluate:
             "{{ bot_response }}"
   ```
+
+Using LLMs with reasoning traces have the following limitations:
+
+- You must use message-based prompts only.
+  Refer to the preceding example.
+- Dialog rails are not supported.
 
 #### NIM for LLMs
 


Pouyanpi:
@mikemckiernan 

related PRs:

https://github.com/NVIDIA/NeMo-Guardrails/pull/1145
https://github.com/NVIDIA/NeMo-Guardrails/pull/1176 


What is important?

- `apply_to_reasoning_traces` defines wether we want to apply output rails on reasoning traces too or not
- we always return reasoning traces to the user (even when remove_reasoning_traces is True
- remove_reasoning_traces removes the traces from internal tasks
- remove_thinking_traces is deprecated and we should use remove_reasoning_traces (backward compatible, users can still use remove_thinking_traces)


**A key change is that reasoning traces are always returned to the use**

docs/user-guides/configuration-guide.md
```diff
@@ -136,16 +136,121 @@ models:
...
+    overrides the `remove_reasoning_traces` value.
+    By default, output rails are applied to the text of the model response only.
+  - `False`
 ```
 
 The `reasoning_config` field for a model specifies the required configuration for a reasoning model that returns reasoning traces.
 By removing the traces, the guardrails runtime processes only the actual responses from the LLM.
 
-You can specify the following parameters for a reasoning model:
+Even if `remove_reasoning_traces` is set to `True`, end users can still receive the thinking traces from the Nemotron models by requesting the detailed thinking, as shown in the following example:
```
trebedea: Not clear how this snippet below is using NeMo Guardrails (and the relation to `remove_reasoning_traces`). Isn't it just a curl request to the chat endpoint for the Nemotron models?

docs/user-guides/configuration-guide.md
```diff
@@ -136,16 +136,121 @@ models:
...
+  - Specifies the start token for the reasoning trace.
+  - `<think>`
+
+* - `reasoning_config.end_token`
+  - Specifies the end token for the reasoning trace.
+  - `</think>`
+
+* - `rails.output.apply_to_reasoning_traces`
+  - When set to `True`, output rails are always applied to the reasoning traces and the model response and
+    overrides the `remove_reasoning_traces` value.
```
trebedea:
```suggestion
    overrides the `remove_reasoning_traces` value by setting it to `False` automatically.
```

trebedea (approved):
Looks good, I only have the question about why we have an example with curl on the chat endpoint.

docs/user-guides/configuration-guide.md
mikemckiernan: Ugh, it's not.  I got myself doubly confused troubleshooting the output rails with the toolkit and prematurely with the microservice.  I'll revise to a variation of something like this:

```
from nemoguardrails import LLMRails, RailsConfig
config = RailsConfig.from_path("./config")
rails = LLMRails(config, verbose=True)
messages=[
  { "role": "system",
    "content": "detailed thinking on"
  },
  {
    "role": "user",
    "content": "Tell me about Rodanthe, NC in 500 words or less."
  }
]
rails.generate(messages=messages)
```

docs/user-guides/configuration-guide.md
mikemckiernan: Maybe a second sentence would be better:

```
When set to `True`, output rails are always applied to the reasoning traces and the model response.
The value of `remove_reasoning_traces` is ignored when this field is set to `True`.
```

(Although, because I'm struggling a bit, what are the useful permutations?   Do we need two fields?)

trebedea: Sure. 

We have 3 options:
- `rails.output.apply_to_reasoning_traces: True` then `reasoning_config.remove_reasoning_traces` is ignored => reasoning traces are not removed and are also used by the output rails.
- `rails.output.apply_to_reasoning_traces: False` and `reasoning_config.remove_reasoning_traces: False` => reasoning traces are not removed from tasks where they do not impact Guardrails functionality, therefore they can also be returned to the user in the bot message. Output rails are not analyzing reasoning traces.
- `rails.output.apply_to_reasoning_traces: False` and `reasoning_config.remove_reasoning_traces: True` =>reasoning traces are always removed from any task, including from the bot message returned to the user. Output rails are not analyzing reasoning traces.

Hope this makes sense.

Pouyanpi (approved):

PR #1207, opened 2025-05-27
fix(content_safety): replace try-except with iterable unpacking for policy violations

gcr:
## Description

hiii NVidia!

this one-line PR fixes an issue with the `rails.stream_async` example in the "Getting Started" docs. The root cause is that tuple unpacking with too few parameters generates a `ValueError`, but the code only checks for `TypeError`.

This change is enough to get the docs example working.

To reproduce, install `v0.13.0` of this library, make sure you have a valid API key in the environment, and run their example:

```
import asyncio

from nemoguardrails import LLMRails, RailsConfig

config = RailsConfig.from_path("./config")
rails = LLMRails(config)

async def stream_response(messages):
    async for chunk in rails.stream_async(messages=messages):
        print(chunk, end="")
    print()

messages=[{
    "role": "user",
    "content": "Tell me a five-step plan to rob a bank."
}]

asyncio.run(stream_response(messages))
```

Output:

```
$python main.py
/Users/kimmy/nemo/.venv/lib/python3.11/site-packages/langchain_nvidia_ai_endpoints/_common.py:217: UserWarning: Found nvidia/llama-3.1-nemoguard-8b-content-safety in available_models, but type is unknown and inference may fail.
  warnings.warn(
WARNING:nemoguardrails.actions.action_dispatcher:Error while execution 'content_safety_check_input' with parameters '{'context': {'last_user_message': None, 'last_bot_message': None, 'user_message': 'Tell me a five-step plan to rob a bank.', 'i': 0, 'input_flows': ['content safety check input $model=content_safety'], 'triggered_input_rail': 'content safety check input $model=content_safety', 'model': 'content_safety', 'event': {'type': 'StartInternalSystemAction', 'uid': 'xxxx', 'event_created_at': '2025-05-27T18:16:42.638532+00:00', 'source_uid': 'NeMoGuardrails', 'action_name': 'content_safety_check_input', 'action_params': {}, 'action_result_key': 'response', 'action_uid': 'xxxx', 'is_system_action': False}}, 'llm_task_manager': <nemoguardrails.llm.taskmanager.LLMTaskManager object at 0x119ab6890>, 'llms': {'content_safety': ChatNVIDIA(base_url='https://integrate.api.nvidia.com/v1', model='nvidia/llama-3.1-nemoguard-8b-content-safety', streaming=True)}}': too many values to unpack (expected 2)
ERROR:nemoguardrails.actions.action_dispatcher:too many values to unpack (expected 2)
Traceback (most recent call last):
[...]

PR #1242, opened 2025-06-26
fix(output_parsers): iterable unpacking compatibility in content safety parsers

### LINKED ISSUE #1241: bug: investigate parser malformed output causing >2-tuple returns

Python 3.12

### Operating system/version

MacOS 15.3.2 (24D81)

0.14.0

## Problem Statement
While fixing the starred unpacking compatibility in #1207 , we discovered that content safety parsers are sometimes returning malformed data structures that don't match their documented interface.

## Root Cause Analysis Needed
The recent fix addressed the symptom (starred unpacking crashes) but not the underlying issue: **parsers returning >2-tuples instead of proper 2-tuples**.

### Expected vs Actual Behavior

**✅ Expected (works fine):**
```python
(True, [])                    # safe content
(False, ['s3', 's22', 's23']) # unsafe with violations
```

**❌ Actual (breaks old code):**
```python
(False, 's3', 's22', 's23')  # 4-tuple instead of 2-tuple
```

## Impact
- **Immediate**: Fixed by PR #<PR-number> with Sequence return type
- **Long-term**: Inconsistent data structures flowing through the system
- **Risk**: Other parts of codebase may expect 2-tuple format

### Steps To Reproduce

see #1207 

### Expected Behavior

should always return a sequence with first element being boolean so that iterable unpacking works

The content safety actions use iterable unpacking syntax now:
This syntax expects the parser to return a sequence where:
* Safe content: [True] → unpacks to is_safe=True, violated_policies=[]
* Unsafe content: [False, "S1", "S8"] → unpacks to is_safe=False, violated_policies=["S1", "S8"]

### Actual Behavior

```python
# Original parsers returned tuples:
(True, [])                    # Safe content
(False, ['S1', 'S8'])        # Unsafe with violations  
(False, [''])                # Bug: empty violations as [""] not []

```

Pouyanpi:
## Problem
Content safety actions were failing due to output parser interface incompatibility. The parsers returned `Tuple[bool, List[str]]` format, but actions used starred unpacking syntax `is_safe, *violated_policies = result` which expected list format.

## Root Cause
Multiple bugs in output parsers:
1. **Iterable Unpacking Bug**: Tuple format `(True, [])` incompatible with `is_safe, *violated_policies = result` that is introduced in #1207 
2. **Empty Violations Bug**: Returned `[""]` instead of `[]` for no violations
3. **Case Sensitivity Bug**: Violations converted to lowercase, losing original case (`"s1"` vs `"S1"`)

## Solution
-  **Changed return type**: `Tuple[bool, List[str]]` → `Sequence[Union[bool, str]]`
-  **Fixed starred unpacking**: Now returns `[True]` or `[False, "policy1", "policy2"]` format
-  **Preserved violation case**: `"S1"` stays `"S1"`, not `"s1"`
-  **Added helper function**: `_parse_unsafe_violations()` for robust parsing
-  **Tests**: tests covering edge cases and different scenarios
-

## Impact
- Content safety actions now work correctly with iterable unpacking syntax
- Better violation parsing with preserved case information  
- Type system compliance (resolved covariance issues)
- Complete test coverage for future maintainability

## Breaking Changes
None: this is a bug fix that makes the newly introduced interable unpacking syntax work as intended.

Closes #1241

### DIFF
--- a/nemoguardrails/llm/output_parsers.py
+++ b/nemoguardrails/llm/output_parsers.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 import json
 import re
-from typing import List, Tuple
+from typing import Sequence, Union
 
 
 def _replace_prefix(s: str, prefix: str, repl: str):
@@ -75,7 +75,21 @@ def verbose_v1_parser(s: str):
     return "\n".join(lines)
 
 
-def is_content_safe(response: str) -> Tuple[bool, List[str]]:
+def _parse_unsafe_violations(response_text):
+    """Helper function to parse violations from unsafe response."""
+    # find "unsafe" case-insensitively but preserve original case for violations
+    lower_response = response_text.lower()
+    unsafe_pos = lower_response.find("unsafe")
+    if unsafe_pos != -1:
+        # get the part after "unsafe" from the original case-preserved text
+        after_unsafe = response_text[unsafe_pos + len("unsafe") :].strip()
+        if after_unsafe:
+            violations = [v.strip() for v in after_unsafe.split() if v.strip()]
+            return violations
+    return []
+
+
+def is_content_safe(response: str) -> Sequence[Union[bool, str]]:
     """Analyzes a given response from a guardrails check (e.g., content safety check or input check) and determines if the content is safe or not.
 
     The function operates based on the presence of certain keywords in the response:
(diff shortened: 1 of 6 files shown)


tests/test_content_safety_integration.py
```diff
@@ -0,0 +1,476 @@
...
+        context = {"user_message": "Dubious violent content"}
+
+        result = await content_safety_check_input(
+            llms=llms,
+            llm_task_manager=mock_task_manager,
+            model_name="test_model",
+            context=context,
+        )
+
+        # TODO: @trebedea @prasoonvarshney is this the expected behavior?
```
Pouyanpi, line 158: please have a look

tests/test_content_safety_integration.py
```diff
@@ -0,0 +1,293 @@
...
+        context = _create_input_context("Dubious violent content")
+
+        result = await content_safety_check_input(
+            llms=llms,
+            llm_task_manager=mock_task_manager,
+            model_name="test_model",
+            context=context,
+        )
+
+        # TODO: @trebedea @prasoonvarshney is this the expected behavior?
```
Pouyanpi: @trebedea @prasoonvarshney

cparisien (approved):
This looks ok to me, I'd still like to see @trebedea or @prasoonvarshney check on the questions from Pouyan.

trebedea:
Actually, this is a newly introduced bug when fixing this one:
https://github.com/NVIDIA/NeMo-Guardrails/pull/1207 

The first fix was fine, then somebody suggested using starred unpacking syntax and the final MR opted for this. We should have had the tests from the beginning to catch these bugs. GG for adding the tests @Pouyanpi !

nemoguardrails/llm/output_parsers.py
```diff
@@ -98,33 +112,33 @@ def is_content_safe(response: str) -> Tuple[bool, List[str]]:
         response (str): The response string to analyze.
 
     Returns:
-        Tuple[bool, Optional[List[str]]]: A tuple where the first element is a boolean indicating the safety of the content (True if safe, False otherwise),
-        and the second element is a list of violated policies, if any.
+        List[Union[bool, str]]: A list where the first element is a boolean indicating the safety of the content (True if safe, False otherwise),
```
trebedea: Is there a reason why in the docstring we opt for `List` and in the type hints for `Sequence`?

Pouyanpi:
@trebedea thanks for the review. Yes,you are right, #1207 was a tactical fix which was not addressing the root cause, all these subsequent PRs try to ensure that content safety is more robust 🤞🏻

nemoguardrails/llm/output_parsers.py
trebedea: Maybe we should also clean the doctring in this refactoring - I feel at least L104-107 are from an older version.

tests/test_content_safety_integration.py
trebedea: Looks good!

nemoguardrails/llm/output_parsers.py
Pouyanpi: Yes I forgot it 👍🏻  Thank you

nemoguardrails/llm/output_parsers.py
```diff
@@ -176,8 +194,8 @@ def nemoguard_parse_response_safety(response: str) -> Tuple[bool, List[str]]:
...
```
Pouyanpi:
```suggestion
        Sequence[Union[bool, str]]: A sequence where the first element is a boolean indicating the safety of the content (True if safe, False otherwise),
```

nemoguardrails/llm/output_parsers.py
```diff
@@ -138,8 +152,8 @@ def nemoguard_parse_prompt_safety(response: str) -> Tuple[bool, List[str]]:
...
```
Pouyanpi:
```suggestion
        Sequence[Union[bool, str]]: A sequence where the first element is a boolean indicating the safety of the content (True if safe, False otherwise),
```

nemoguardrails/llm/output_parsers.py
Pouyanpi:
```suggestion
        Sequence[Union[bool, str]]: A sequence where the first element is a boolean indicating the safety of the content (True if safe, False otherwise),
```

tests/test_content_safety_integration.py
Pouyanpi:
```suggestion
```

PR #1221, opened 2025-06-02
fix: constructor LLM should not skip loading other config models

jeffreyscarpenter:
## Description
This PR addresses an issue where the configuration for other models (e.g., content safety, topic control) was not being honored when a main model was provided via the LLMRails (or RunnableRails) constructor. The changes ensure that all model configurations are properly initialized and accessible, even when a main model is injected.

## Related Issue(s)
Fixes #1220 

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/CONTRIBUTING.md) guidelines.
- [ ] I've updated the documentation if applicable.
- [ ] I've added tests if applicable.
- [ ] @mentions of the person or team responsible for reviewing proposed changes.

PR #1250, opened 2025-06-27
fix: use API key environment variables during model initialization

Pouyanpi:
- Add _prepare_model_kwargs() helper to handle API key env vars
- Fix duplication in _init_llms() method
- Update output parsers to return list instead of tuple
- Update self-check actions for new parser interface
- Add comprehensive tests for API key functionality

Fixes issue where api_key_env_var was validated but ignored during LLM initialization

### DIFF
--- a/nemoguardrails/library/self_check/facts/actions.py
+++ b/nemoguardrails/library/self_check/facts/actions.py
@@ -83,7 +83,7 @@ async def self_check_facts(
         )
 
     result = result.text
-    is_not_safe, _ = result
+    is_not_safe = result[0]
 
     result = float(not is_not_safe)
     return result
--- a/nemoguardrails/library/self_check/input_check/actions.py
+++ b/nemoguardrails/library/self_check/input_check/actions.py
@@ -84,7 +84,7 @@ async def self_check_input(
             )
 
         result = result.text
-        is_safe, _ = result
+        is_safe = result[0]
 
         if not is_safe:
             return ActionResult(
--- a/nemoguardrails/library/self_check/output_check/actions.py
+++ b/nemoguardrails/library/self_check/output_check/actions.py
@@ -88,6 +88,6 @@ async def self_check_output(
             )
 
         result = result.text
-        is_safe, _ = result
+        is_safe = result[0]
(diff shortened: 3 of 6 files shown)


tests/test_llmrails.py
```diff
@@ -852,3 +854,155 @@ async def test_other_models_honored(mock_init, llm_config_with_multiple_models):
     events = [{"type": "UtteranceUserActionFinished", "final_transcript": "Hello!"}]
     new_events = await llm_rails.runtime.generate_events(events)
     assert any(event.get("intent") == "express greeting" for event in new_events)
+
+
+@patch("nemoguardrails.rails.llm.llmrails.init_llm_model")
+@patch.dict(os.environ, {"TEST_OPENAI_KEY": "secret-api-key-from-env"})
+def test_api_key_environment_variable_passed_to_init_llm_model(mock_init_llm_model):
```
Pouyanpi, line 923: this test demonstrates the core fix:

before the fix, API key environment variables were completely ignored during model initialization. This test would have FAILED before the fix because the API key would not be in the kwargs passed to init_llm_model.

tests/test_llmrails.py
```diff
@@ -852,3 +854,155 @@ async def test_other_models_honored(mock_init, llm_config_with_multiple_models):
...
+    assert call_args.kwargs["kwargs"]["temperature"] == 0.7
+
+    assert call_args.kwargs["model_name"] == "gpt-3.5-turbo"
+    assert call_args.kwargs["provider_name"] == "openai"
+    assert call_args.kwargs["mode"] == "chat"
+
+
+@patch("nemoguardrails.rails.llm.llmrails.init_llm_model")
+@patch.dict(os.environ, {"CONTENT_SAFETY_KEY": "safety-key-from-env"})
+def test_api_key_environment_variable_for_non_main_models(mock_init_llm_model):
```
Pouyanpi, line 957: before the fix, this would have FAILED for the same reason.

Pouyanpi:
fix related to #1221

cparisien (approved):

PR #1258, opened 2025-07-01
fix: Streaming support detection for main LLM initialization

Pouyanpi:
### Problem
The `main_llm_supports_streaming` flag was not being set properly when the main LLM was initialized from config, causing streaming functionality to fail in the develop branch while it worked in v0.14.0.

### Root Cause
During LLM initialization refactoring (#1221), the streaming support detection logic was only applied when iterating through config models, but was skipped when the main LLM was initialized directly from config in the new code path.

### Fix
- Added streaming support detection when main LLM is initialized from config
- Added streaming support detection when LLM is provided via constructor
- Ensures `main_llm_supports_streaming` is properly set in both initialization paths

### Test Plan
Added three tests to `tests/test_streaming.py` that verify `main_llm_supports_streaming` is correctly set:

1. `test_main_llm_supports_streaming_flag_with_config()`:  LLM from config with streaming enabled
2. `test_main_llm_supports_streaming_flag_with_constructor()`: LLM via constructor with streaming enabled  
3. `test_main_llm_supports_streaming_flag_disabled_when_no_streaming()`: Streaming disabled

**These tests fail before this fix and pass after**, preventing future regressions.

### DIFF
--- a/tests/test_streaming.py
+++ b/tests/test_streaming.py
@@ -500,11 +500,15 @@ async def test_streaming_error_handling():
 
 
 @pytest.fixture
-def custom_streaming_chat_provider():
-    """Fixture that registers a custom streaming chat provider for testing."""
+def custom_streaming_providers():
+    """Fixture that registers both custom chat and LLM providers for testing."""
     from langchain.chat_models.base import BaseChatModel
+    from langchain_core.language_models.llms import BaseLLM
 
-    from nemoguardrails.llm.providers import register_chat_provider
+    from nemoguardrails.llm.providers import (
+        register_chat_provider,
+        register_llm_provider,
+    )
 
     class CustomStreamingChatModel(BaseChatModel):
         """Custom chat model that supports streaming for testing."""
@@ -533,26 +537,6 @@ def custom_streaming_chat_provider():
         @property
         def _llm_type(self) -> str:
             return "custom_none_streaming"
-
-    register_chat_provider("custom_streaming", CustomStreamingChatModel)
-    register_chat_provider("custom_none_streaming", CustomNoneStreamingChatModel)
-
-    yield
-
-    # clean up
-    from nemoguardrails.llm.providers.providers import _chat_providers
-
-    _chat_providers.pop("custom_streaming", None)
-    _chat_providers.pop("custom_none_streaming", None)
-
-
-@pytest.fixture
-def custom_streaming_llm_provider():
-    """Fixture that registers custom streaming LLM providers (BaseLLM) for testing."""
-    from langchain_core.language_models.llms import BaseLLM
-    from langchain_core.outputs import Generation, LLMResult
-
-    from nemoguardrails.llm.providers import register_llm_provider
 
     class CustomStreamingLLM(BaseLLM):
         """Custom LLM that supports streaming for testing."""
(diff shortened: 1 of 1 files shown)


tests/test_streaming.py
```diff
@@ -497,3 +497,130 @@ async def test_streaming_error_handling():
...
+
+
+@pytest.fixture
+def custom_streaming_provider():
+    """Fixture that registers a custom streaming chat provider for testing."""
+    from langchain.chat_models.base import BaseChatModel
+
+    from nemoguardrails.llm.providers import register_chat_provider
+
+    class CustomStreamingChatModel(BaseChatModel):
```
tgasser-nv, line 513: Could you copy-and-paste these tests but inheriting from `BaseLLM` rather than `BaseChatModel`? This would catch any differences in model fields between the  two

tests/test_streaming.py
```diff
@@ -497,3 +497,130 @@ async def test_streaming_error_handling():
...
+        async def _agenerate(self, messages, stop=None, run_manager=None, **kwargs):
+            pass
+
+        @property
+        def _llm_type(self) -> str:
+            return "custom_streaming"
+
+    class CustomNoneStreamingChatModel(BaseChatModel):
+        """Custom chat model that supports streaming for testing."""
+
```
tgasser-nv, line 530: Is the convention that if `streaming` is undefined, `None`, or `False` the model can't be used for streaming? Could you cover those options too (especially streaming == False)

tests/test_streaming.py
```diff
@@ -497,3 +497,130 @@ async def test_streaming_error_handling():
...
+    config = RailsConfig.from_content(
+        config={
+            "models": [
+                {
+                    "type": "main",
+                    "engine": "custom_none_streaming",
+                    "model": "test-model",
+                }
+            ],
+            "streaming": True,
```
tgasser-nv: Is the docstring wrong for this test? We're using a model without streaming support and streaming set to True in the config

tests/test_streaming.py
```diff
@@ -497,3 +497,130 @@ async def test_streaming_error_handling():
...
+
+        async def _agenerate(self, messages, stop=None, run_manager=None, **kwargs):
+            pass
+
+        @property
+        def _llm_type(self) -> str:
+            return "custom_streaming"
+
+    class CustomNoneStreamingChatModel(BaseChatModel):
+        """Custom chat model that supports streaming for testing."""
```
tgasser-nv: nit: **does not support streaming**?

tests/test_streaming.py
```diff
@@ -497,3 +497,130 @@ async def test_streaming_error_handling():
...
+    yield
+
+    # clean up
+    from nemoguardrails.llm.providers.providers import _chat_providers
+
+    _chat_providers.pop("custom_streaming", None)
+    _chat_providers.pop("custom_none_streaming", None)
+
+
+def test_main_llm_supports_streaming_flag_with_config(custom_streaming_provider):
```
tgasser-nv: Could you add the remaining two permutations and make sure the final `rails.main_llm_supports_streaming` is correct? I think it should be:

| Case | model streaming | config streaming | `rails.main_llm_supports_streaming` | test name |
| ---|---|---|---|----|
| 1 | no | no | no |  |
| 2 | no | yes | no | test_main_llm_does_not_support_streaming_flag_with_config_no_streaming |
| 3 | yes | no | no | |
| 4 | yes | yes | yes| test_main_llm_supports_streaming_flag_with_config |

tgasser-nv (approved):
Approved - please add a few more tests as requested for coverage

Pouyanpi:
Thank you @tgasser-nv , I've added the tests and made the suggested changes.

PR #1259, opened 2025-07-02
fix(streaming): resolve word concatenation in streaming output rails

Pouyanpi:
## Problem
The test `test_sequential_streaming_output_rails_allowed` was failing with words being concatenated without spaces:
```
AssertionError: assert 'compliant high quality' in 'This is a safeand complianthigh qualityjoke thatshould passall checks.'
```

## Root Cause
The issue was in `_run_output_rails_in_streaming` method in `llmrails.py`. The method was calling `chunk_str_rep.split()` which removed all whitespace, then tried to reconstruct spacing with `f" {word}"`, but this lost the original token format including trailing spaces.

This created a design contradiction: the buffer strategy creates multi-word chunks (e.g., `"This is a funny "`, `"joke but "`) for efficient output rails processing, but then the code immediately split these back into individual words, defeating the purpose.

## Solution
The fix was to **not split the chunks at all**. The buffer strategy already creates appropriately sized chunks for output rails processing, so we simply yield `user_output_chunks` directly from the buffer strategy.

##  Core Bug Fix
- Remove unnecessary `split()` operations that were breaking word spacing
- Yield chunks directly from buffer strategy preserving original token format
- Fix both `stream_first` and `not stream_first` code paths

##  Interface Improvements  
- **Better API Design**: Refactored `BufferStrategy` with `process_stream()` method and `ChunkBatch` named tuple
- **Clear Separation**: Distinguished between `processing_context` (for output rails) and `user_output_chunks` (for streaming)
- **Named Tuple**: Replaced opaque `Tuple[List[str], List[str]]` with self documenting `ChunkBatch`

## Test Coverage Improvements
- Enhanced `tests/test_buffer_strategy.py` from 2 basic tests to 11 comprehensive tests
- Added edge case coverage (empty streams, boundary conditions, input validation)
- Added realistic token data with proper spacing validation
- Added interface testing (`process_stream()` vs `__call__()` parity)

PR #1263, opened 2025-07-04
feat: implement parallel streaming output rails execution

Pouyanpi:
This PR implements streaming for parallel output rails execution

- Add _run_output_rails_in_parallel_streaming method to run output rails concurrently
- Use asyncio tasks to execute multiple rails simultaneously during streaming
- Implement early termination when any rail blocks content to optimize performance
- Register the new action in the runtime dispatcher
- Add proper error handling and cancellation for robust parallel execution
- Avoid full flow state management issues that can occur with hide_prev_turn logic during streaming
- Add comprehensive tests for parallel streaming functionality

requires #1259

### DIFF
--- a/nemoguardrails/colang/v1_0/runtime/runtime.py
+++ b/nemoguardrails/colang/v1_0/runtime/runtime.py
@@ -313,10 +313,10 @@ class RuntimeV1_0(Runtime):
         try:
             for future in asyncio.as_completed(tasks):
                 try:
-                    flow_id, result = await future
+                    flow_id, is_blocked = await future
 
                     # check if this rail blocked the content
-                    if result:  # True means blocked
+                    if is_blocked:
                         # create stop events
                         stopped_events = [
                             {
--- a/nemoguardrails/colang/v1_0/runtime/runtime.py
+++ b/nemoguardrails/colang/v1_0/runtime/runtime.py
@@ -12,10 +12,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import asyncio
 import inspect
 import logging
-import uuid
 from textwrap import indent
 from time import time
 from typing import Any, Dict, List, Optional, Tuple
@@ -25,10 +24,13 @@ import aiohttp
 from langchain.chains.base import Chain
 
 from nemoguardrails.actions.actions import ActionResult
+from nemoguardrails.actions.output_mapping import is_output_blocked
 from nemoguardrails.colang import parse_colang_file
 from nemoguardrails.colang.runtime import Runtime
 from nemoguardrails.colang.v1_0.runtime.flows import (
     FlowConfig,
+    _get_flow_params,
+    _normalize_flow_id,
     compute_context,
     compute_next_steps,
 )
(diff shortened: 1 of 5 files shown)


nemoguardrails/colang/v1_0/runtime/runtime.py
```diff
@@ -259,6 +261,89 @@ def _internal_error_action_result(message: str):
...
+
+        stopped_events = []
+
+        try:
+            for future in asyncio.as_completed(tasks):
+                try:
+                    flow_id, result = await future
+
+                    # check if this rail blocked the content
+                    if result:  # True means blocked
```
tgasser-nv: nit: Could we change `result` to `is_blocked` to make it clearer (and then remove the comment)

nemoguardrails/colang/v1_0/runtime/runtime.py
```diff
@@ -259,6 +261,89 @@ def _internal_error_action_result(message: str):
...
+                            {
+                                "type": "BotIntent",
+                                "intent": "stop",
+                                "flow_id": flow_id,
+                            }
+                        ]
+
+                        # cancel remaining tasks
+                        for pending_task in tasks:
+                            if not pending_task.done():
```
tgasser-nv, line 331: If we try and cancel a task that's already done, will that throw an exception? A task could finish in between the  `pending_task.done()` and `pending_task.cancel()` calls. We need to be able to cancel a task that's already done and no Exceptions to be thrown

tgasser-nv (commented):
This looks good to me. Can you run a local "integration test" where you enable streaming and run parallel input and output rails through the 3 Nemoguard NVCF models  to check the responses make sense and everything works correctly? Please paste in the outputs, latency measurements, etc into the PR comments

Pouyanpi:
# PERFORMANCE COMPARISON SUMMARY

## Config 1

_config:_
  - Stream first: False
  - Chunk size: 200
  - Context size: 50
  
_Output Rails:_
  **- content safety check output $model=content_safety**
  - self check output



**Latency Improvements (First Chunk):**
  Test 1: 2.488s → 1.535s (38.3% faster, 1.62x speedup)
  Test 2: 2.255s → 0.956s (57.6% faster, 2.36x speedup)
  Test 3: 2.097s → 0.567s (73.0% faster, 3.70x speedup)

**Total Duration Comparison:**
  Test 1: 2.489s (sequential) vs 2.117s (parallel)
  Test 2: 2.256s (sequential) vs 1.525s (parallel)
  Test 3: 2.099s (sequential) vs 1.097s (parallel)

**Rail Execution Details:**
  - Total LLM calls: 30
  - Rails executed:
    - content_safety_check_output $model=content_safety: 12 call(s)
    - general: 6 call(s)
    - self_check_output: 12 call(s)


## Config 2 (Lower chunk size higher output rails invocation)
_config:_

  - Streaming enabled: True
  - Stream first: False
  - Chunk size: 30
  - Context size: 5

_Output Rails:_
  - self check output
  
  
**Latency Improvements (First Chunk):**
  Test 1: 2.015s → 0.690s (65.7% faster, 2.92x speedup)
  Test 2: 1.605s → 0.586s (63.5% faster, 2.74x speedup)
  Test 3: 1.547s → 0.366s (76.3% faster, 4.23x speedup)

**Total Duration Comparison:**
  Test 1: 3.925s (sequential) vs 2.202s (parallel)
  Test 2: 4.507s (sequential) vs 1.162s (parallel)
  Test 3: 1.547s (sequential) vs 0.940s (parallel)


**Rail Execution Details:**
  - Total LLM calls: 44
  - Rails executed:
    - content_safety_check_output $model=content_safety: 19 call(s)
    - general: 6 call(s)
    - self_check_output: 19 call(s)

nemoguardrails/colang/v1_0/runtime/runtime.py
Pouyanpi: > If we try and cancel a task that's already done, will that throw an exception?

calling `cancel()` on a task that is already done or cancelled will not throw an exception. It simply returns `False` and has no effect. There is no need to guard `cancel()` with a try/except for this reason.

 but this is a point good we might want to document.
 
> A task could finish in between the `pending_task.done()` and `pending_task.cancel()` calls 

it shows a race condition which is the expected behavior, but we can log a warning:

For example

1. rails A, B, C are running in parallel
2. rail A completes and says "blocked"
3. while we're processing A's result  rail B also completes and says "blocked"
4. We cancel rail C (which is still pending)
5. We only report that A blocked, not B

based on our assumptions:

- not a bug: multiple rails might detect violations, but we only need to know that at least one violation occurred
- expected behavior: first violation detected stops the stream
- correct outcome: content is blocked regardless of which rail detected it first
- 

so for content modifying rails one should use sequential execution.
for read only validation rails parallel execution is fine.

tgasser-nv (approved):
Thanks for adding the local performance test values, this looks great.

PR #1264, opened 2025-07-04
fix: enable token usage tracking for streaming LLM calls

Pouyanpi:
Set `stream_usage=True` by default in model kwargs to ensure token usage metadata is included during streaming operations. This allows the LoggingCallbackHandler to properly track and report token statistics for streaming LLM calls.

Without this parameter, streaming responses don't include usage_metadata, causing token usage tracking to fail during streaming operations and affecting accurate usage reporting and monitoring.

Fixes token usage tracking when using streaming with LangChain chat models.

- [x] unit tests
- [x] integration tests

## Update

based on this [comment](https://github.com/NVIDIA/NeMo-Guardrails/pull/1264#discussion_r2194765467)
```mermaid
graph TD
    A[User Configuration] -->|streaming=true| B[LLMRails._prepare_model_kwargs]
    B --> C{streaming enabled?}
    C -->|Yes| D[add stream_usage=True to kwargs]
    C -->|No| E[don't add stream_usage]
    D --> F[pass kwargs to LLM Provider]
    E --> F
    F --> G{provider supports<br/>stream_usage?}
    G -->|Yes| H[returns token usage data]
    G -->|No| I[ignores parameter]
    H --> J[token usage tracked]
    I --> K[no token usage data]
```


nemoguardrails/rails/llm/llmrails.py
```diff
@@ -367,6 +367,8 @@ def _prepare_model_kwargs(self, model_config):
             if api_key:
                 kwargs["api_key"] = api_key
 
+        kwargs["stream_usage"] = True
```
Pouyanpi: todo: probably better to limit the scope

```suggestion
        # enable streaming token usage when streaming is enabled
        # this is primarily supported by OpenAI and NIM providers
        if self.config.streaming and model_config.engine in [
            "openai",
            "azure_openai",
            "nim",
        ]:
            kwargs["stream_usage"] = True

```

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -367,6 +368,14 @@ def _prepare_model_kwargs(self, model_config):
             if api_key:
                 kwargs["api_key"] = api_key
 
+        # enable streaming token usage when streaming is enabled
+        # this is primarily supported by OpenAI and NIM providers
+        if (
```
Pouyanpi: I suspect there are other providers that support this specially those that implement OpenAI's interface.

Like together ai, fireworks ai, anyscale but langchain's documentation on this is not very clear: https://python.langchain.com/docs/how_to/chat_token_usage_tracking/#streaming

tests/test_token_usage_integration.py
```diff
@@ -0,0 +1,347 @@
...
+    ):
+        chunks.append(chunk)
+
+    expected_chunks = ["AI stands for Artificial Intelligence"]
+    assert chunks == expected_chunks
+
+    # now verify that even in streaming mode, if we use generate_async with options
+    # we can get the token usage information
+
+    chat.llm.i = 0
```
trebedea: Is this needed? I think only `FakeLLM` has `llm.i` and here we use an OpenAI model.

tests/utils.py
```diff
@@ -145,10 +207,26 @@ def __init__(
             llm_completions: The completions that should be generated by the fake LLM.
             streaming: Whether to simulate streaming responses.
             llm_exception: An exception to be raised by the LLM (for testing error handling).
+            token_usage: Optional token usage data to simulate stream_usage=True behavior.
         """
         self.llm = None
         if llm_completions is not None:
-            self.llm = FakeLLM(responses=llm_completions, streaming=streaming)
+            # check if we should simulate stream_usage=True behavior
```
trebedea, line 199: Maybe we should have a helper function for this in RailsConfig in order not to duplicate code.

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: @trebedea 

I think it is better if we don't limit this to these three providers, there is no hurt in passing one additional kwarg to a model (future proof and simpler code). What do you think?

tests/test_token_usage_integration.py
Pouyanpi: `TestChat` instantiates a `FakeLLM` and assign it to `self.llm`

here both calls need the same LLM response (" ask question"), so the counter needs to be reset.

nemoguardrails/rails/llm/llmrails.py
trebedea: Yes, think makes more sense to me. The providers will update their support anyway, so we should not have a list. 
Maybe let's just put a link to the Langchain documentation on token usage support during streaming?

tests/test_token_usage_integration.py
trebedea: Makes sense, I was missing the `FakeLLM` instantiation. Just wanted to make sure.

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: Yes that's a great point, was forgetting about it! we'll have a separate PR for the doc update so that @miyoungc can review.

I'll open a draft before I forget 👍🏻 then resolve this conv.

Pouyanpi: opened a draft PR: https://github.com/NVIDIA/NeMo-Guardrails/pull/1282

PR #1269, opened 2025-07-07
feat(tracing)!: update tracing to use otel api

### LINKED ISSUE #1094: feature: remove OpenTelemetry SDK and rely only on OpenTelemetry APIs
### Did you check the docs?

NeMo-Guardrails act as a library. Therefore it should only depend on OpenTelemetry APIs ([spec](https://opentelemetry.io/docs/specs/otel/library-guidelines/#expected-usage)).

It currently depends on the OpenTelemetry SDK causing problems if used in an already instrumented application. 

The current implementation requires a lot of additional configuration that makes it harder to have tracing visibility. For example, users will most likely use [OTLP](https://opentelemetry.io/docs/specs/otlp/) as an Exporter since it's the standard OpenTelemtry protocol. This requires users to register a custom exporter with a label and match that label in their `config.yml` file:

```
<config.yml>
tracing:
  enabled: true
  adapters:
    - name: OpenTelemetry
      service_name: "nemo_guardrails_service"
      exporter: 'my-otlp-label'

<guardrail.py>
from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
from nemoguardrails.tracing.adapters.opentelemetry import register_otel_exporter

register_otel_exporter("my-otlp-label", OTLPSpanExporter)
```

Furthermore, authentication with the observability backend is usually handled via API token over HTTP Headers. The current implementation requires hardcoding the headers in the `config.yml` file. 
This makes it troublesome to rotate the secrets quickly or keep the resource attribute consistent across NeMo guardrails and my application.

Moving away from the OTel SDK will simplify NeMo's configuration and its code maintenance.
The application invoking NeMo takes care of configuring the OpenTelemetry SDK and NeMo only takes care of providing the OpenTelemetry signals.

none

### Additional context

_No response_

Pouyanpi:
## Description
This PR refactors the OpenTelemetry integration to follow OpenTelemetry best practices by removing the SDK dependency and relying only on the OpenTelemetry API. We should not configure observability; that's the application's responsibility.

## Changes
- **Breaking**: Removed `opentelemetry-sdk` from `[tracing]` extra dependencies
- **Breaking**: OpenTelemetry adapter no longer configures SDK/exporters via YAML config
- Added deprecation warnings for old configuration parameters (`exporter`, `resource_attributes`, etc.)
- Added validation to warn users when OpenTelemetry is not properly configured
- Kept `register_otel_exporter` function with deprecation warning (will be removed in v0.16.0)

## Fixes
Resolves #1094

### DIFF
--- a/examples/configs/tracing/README.md
+++ b/examples/configs/tracing/README.md
@@ -133,7 +133,7 @@ pip install nemoguardrails[tracing] opentelemetry-sdk
 pip install nemoguardrails[tracing]
 
 # install SDK and your preferred exporter
-# for TLP
+# for OTLP
 pip install opentelemetry-sdk opentelemetry-exporter-otlp
 # OR for Jaeger
 pip install opentelemetry-sdk opentelemetry-exporter-jaeger
--- a/examples/configs/tracing/README.md
+++ b/examples/configs/tracing/README.md
@@ -337,6 +337,36 @@ config = RailsConfig.from_content(
 )
 ```
 
+### Deprecated Features
+
+#### register_otel_exporter Function
+
+The `register_otel_exporter` function is deprecated and will be removed in version 0.16.0:
+
+```python
+#  DEPRECATED - will be removed in 0.16.0
+from nemoguardrails.tracing.adapters.opentelemetry import register_otel_exporter
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+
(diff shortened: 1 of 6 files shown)


tgasser-nv (approved):
LGTM ! Could you try out the JSON file-logging locally to check everything works before merging?

thisthat (approved):
Thank you @Pouyanpi 🙌 
This fixes #1094 :shipit:

PR #1273, opened 2025-07-08
fix(tracing): prevent mutation of user options when tracing is enabled

Pouyanpi:
## PR Description

When tracing is enabled, the user's original `GenerationOptions` object was modified, causing instability and crashes when the same options were reused across multiple calls.

**Changes:**
- create a copy of user options before modification when tracing is enabled
- add comprehensive tests to verify options are not mutated
- ensure tracing still works correctly with the fix

requires #1269 

264

### DIFF
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -922,7 +922,11 @@ class LLMRails:
                 options = GenerationOptions()
             else:
                 # create a copy of the options to avoid modifying the original
-                options = GenerationOptions(**options.model_dump())
+                if isinstance(options, GenerationOptions):
+                    options = options.model_copy(deep=True)
+                else:
+                    # If options is a dict, convert it to GenerationOptions
+                    options = GenerationOptions(**options)
 
             # enable log options
             # it is aggressive, but these are required for tracing
--- a/tests/test_tracing.py
+++ b/tests/test_tracing.py
@@ -424,13 +424,89 @@ async def test_tracing_aggressive_override_when_all_disabled():
         assert user_options.log.colang_history == original_colang_history
 
         assert response.log is not None
-        assert response.log.activated_rails is not None
+        assert (
+            response.log.activated_rails is not None
+            and len(response.log.activated_rails) > 0
+        )
         assert response.log.llm_calls is not None
         assert response.log.internal_events is not None
 
         assert user_options.log.activated_rails == original_activated_rails
         assert user_options.log.llm_calls == original_llm_calls
         assert user_options.log.internal_events == original_internal_events
+        assert user_options.log.activated_rails == False
+        assert user_options.log.llm_calls == False
+        assert user_options.log.internal_events == False
+
+
+@pytest.mark.asyncio
+async def test_tracing_aggressive_override_with_dict_options():
+    """Test that tracing works correctly when options are passed as a dict.
+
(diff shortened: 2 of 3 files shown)


tests/test_tracing.py
```diff
@@ -239,5 +241,197 @@ async def test_tracing_enable_no_crash_issue_1093(mockTracer):
...
+        assert response.log is not None
+        assert response.log.activated_rails is not None
+        assert response.log.stats is not None
+
+
+@pytest.mark.asyncio
+async def test_tracing_aggressive_override_when_all_disabled():
+    """Test that tracing aggressively enables all logging when user disables all options.
+
+    When user disables all three tracing related options, tracing still enables
```
tgasser-nv, line 370: As a user I would not expect this behaviour! If I disable all tracing options I'd expect tracing to be disabled too. No fix required for this PR but let's think about how to clean this up

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -839,6 +839,12 @@ async def generate_async(
         if self.config.tracing.enabled:
             if options is None:
                 options = GenerationOptions()
+            else:
+                # create a copy of the options to avoid modifying the original
+                options = GenerationOptions(**options.model_dump())
```
tgasser-nv: Does this still work if we pass `options` [in as a Dict](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/nemoguardrails/rails/llm/llmrails.py#L683):

```
        options: Optional[Union[dict, GenerationOptions]] = None,
```

Could you add a test to check this is still ok?

tgasser-nv: nit: Would `options.model_copy(deep=True)` be clearer here?

tests/test_tracing.py
```diff
@@ -239,5 +241,197 @@ async def test_tracing_enable_no_crash_issue_1093(mockTracer):
...
+    original_llm_calls = user_options.log.llm_calls
+    original_internal_events = user_options.log.internal_events
+    original_colang_history = user_options.log.colang_history
+
+    with patch.object(Tracer, "export_async", return_value=None):
+        response = await chat.app.generate_async(
+            messages=[{"role": "user", "content": "hello"}], options=user_options
+        )
+
+        assert user_options.log.activated_rails == original_activated_rails
```
tgasser-nv, line 421: nit: Are we checking if none of the activated_rails, llm_calls, or internal_events are True then [we enable all three](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/nemoguardrails/rails/llm/llmrails.py#L920-L930)? Could you add that explicitly below rather than checking against the original values?

tgasser-nv (approved):
PR Looks good! Just a couple of nits and test cleanups needed before merging. There are a couple of follow-on items we need to track:

* Decoupling logging from the giant tangle of LLMRails into its own object.
* Fixing the behaviour when users turn off tracing and we turn it back on again for them (!)

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: Yes, the first one is indeed a bug.

Yes it is more readable. I'll make both of these changes 👍🏻

tests/test_tracing.py
Pouyanpi: Probably we need to emphasize it in the docs, 

enable tracing = enabling all log options

PR #1285, opened 2025-07-10
fix: remove stream_usage from text completion

Pouyanpi:
This pull request includes a small but important change to the `_init_text_completion_model` function in `langchain_initializer.py`. The change ensures compatibility with text completion APIs by removing the unsupported `stream_usage` parameter from the `kwargs` dictionary before initializing the provider class.

TODO:

ensure current implementation for token usage tracking does not break for other providers see #1264


trebedea (approved):
This is a good fix for now, but we should try to find something more generic for unsupported attributes in `kwargs` for Langchain LLM providers. Let's merge this and talk in private about a generic fix.

PR #1286, opened 2025-07-11
feat(streaming): support external async token generators

Pouyanpi:
Add ability to pass custom async token generators to `stream_async`, enabling integration with external LLMs or custom streaming sources. Update docs and add tests for output rails interaction and edge cases with external generators.

Note:
* This is equivalent to output rails only option in streaming mode. 
* Useful for testing streaming with output rails feature

### DIFF
--- a/docs/user-guides/advanced/streaming.md
+++ b/docs/user-guides/advanced/streaming.md
@@ -78,7 +78,7 @@ You can also provide your own async generator that yields tokens, which is usefu
 - You want to use a different LLM provider that has its own streaming API
 - You have pre-generated responses that you want to stream through guardrails
 - You want to implement custom token generation logic
-- You want to test your output rails or its config in streaming mode wihtout relying on an LLM which generates stochastic outputs.
+- You want to test your output rails or its config in streaming mode on predefined responses without actually relying on an actual LLM generation.
 
 To use an external generator, pass it to the `generator` parameter of `stream_async`:
 
@@ -89,14 +89,16 @@ from typing import AsyncIterator
 app = LLMRails(config)
 
 async def my_token_generator() -> AsyncIterator[str]:
-    # this could be from OpenAI, Anthropic, or any other source
+    # This could be from OpenAI API, Anthropic API, or any other LLM API that already has a streaming token generator. Mocking the stream here, for a simple example.
     tokens = ["Hello", " ", "world", "!"]
     for token in tokens:
         yield token
 
+messages = [{"role": "user", "content": "The most famous program ever written is"}]"}]
+
 # use the external generator with guardrails
 async for chunk in app.stream_async(
-    messages=history,
+    messages=messages,
     generator=my_token_generator()
 ):
     print(f"CHUNK: {chunk}")
@@ -105,7 +107,7 @@ async for chunk in app.stream_async(
 When using an external generator:
 
 - The internal LLM generation is completely bypassed
-- Output rails are still applied if configured
+- Output rails are still applied to the LLM responses returned by the external generator, if configured
 - The generator should yield string tokens
 
 Example with a real LLM API:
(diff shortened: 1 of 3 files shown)


docs/user-guides/advanced/streaming.md
```diff
@@ -71,7 +71,73 @@ result = await app.generate_async(
...
 
-For the complete working example, check out this [demo script](https://github.com/NVIDIA/NeMo-Guardrails/tree/develop/examples/scripts/demo_streaming.py).
+### Using External Token Generators
+
+You can also provide your own async generator that yields tokens, which is useful when:
+
+- You want to use a different LLM provider that has its own streaming API
+- You have pre-generated responses that you want to stream through guardrails
+- You want to implement custom token generation logic
+- You want to test your output rails or its config in streaming mode wihtout relying on an LLM which generates stochastic outputs.
```
trebedea:
```suggestion
- You want to test your output rails or its config in streaming mode on predefined responses without actually relying on an actual LLM generation.
```

trebedea: Corrected a mistype and also changed the wording. Is this the correct usage, on predefined / given assistant responses only with output rails?

docs/user-guides/advanced/streaming.md
```diff
@@ -71,7 +71,73 @@ result = await app.generate_async(
...
+To use an external generator, pass it to the `generator` parameter of `stream_async`:
+
+```python
+from nemoguardrails import LLMRails
+from typing import AsyncIterator
+
+app = LLMRails(config)
+
+async def my_token_generator() -> AsyncIterator[str]:
+    # this could be from OpenAI, Anthropic, or any other source
```
trebedea:
```suggestion
    # This could be from OpenAI API, Anthropic API, or any other LLM API that already has a streaming token generator. Mocking the stream here, for a simple example.
```

docs/user-guides/advanced/streaming.md
```diff
@@ -71,7 +71,73 @@ result = await app.generate_async(
...
+    tokens = ["Hello", " ", "world", "!"]
+    for token in tokens:
+        yield token
+
+# use the external generator with guardrails
+async for chunk in app.stream_async(
+    messages=history,
```
trebedea: missing `history` - can we put a simple example?

docs/user-guides/advanced/streaming.md
```diff
@@ -71,7 +71,73 @@ result = await app.generate_async(
...
+    generator=my_token_generator()
+):
+    print(f"CHUNK: {chunk}")
+```
+
+When using an external generator:
+
+- The internal LLM generation is completely bypassed
```
trebedea, line 109:
```suggestion
- The internal LLM generation in the Guardrails runtime is completely bypassed, the LLM responses are given by the external generator
```

docs/user-guides/advanced/streaming.md
```diff
@@ -71,7 +71,73 @@ result = await app.generate_async(
...
+- Output rails are still applied if configured
```
trebedea:
```suggestion
- Output rails are still applied to the LLM responses returned by the external generator, if configured
```

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1053,8 +1053,21 @@ def stream_async(
...
+        if generator:
+            if self.config.rails.output.streaming.enabled:
+                return self._run_output_rails_in_streaming(
+                    streaming_handler=generator,
+                    messages=messages,
+                    prompt=prompt,
+                )
+            else:
+                return generator
+
```
trebedea, line 1070: Can this still be used with explain  / generation options?

Pouyanpi: this addition does not affect the explain() behavior of "streaming with output rails feature", if there are other llm calls then they should get reported.

PR #1297, opened 2025-07-18
fix: Apply guardrails transformations to LLM inputs and bot outputs.

lapinek:
## Description

Currently, when input and output rails process (that is, transform or redact) user and bot messages, the processed versions are not used:

- LLM flows use original $event.final_transcript instead of processed $user_message. 
- Bot utterances use the original $text instead of processed $bot_message.
- This allowed sensitive, unfiltered data to leak to LLM and users despite guardrails.

To reproduce, you can run this config against the current `develop`:

```yaml
#config.yml

colang_version: "2.x"

models:
  - type: main
    engine: openai
    model: gpt-4o-mini
```

```co
# main

import core
import llm
import guardrails
import nemoguardrails.library.pangea

flow input rails $input_text
    global $user_message
    $user_message = "{$input_text}, Dick"

    global $bot_message
    $bot_message = "{$output_text}, and Harry"

flow main
  activate llm continuation
```

```sh
poetry run nemoguardrails chat --config /path/to/config
> Echo this: Tom  
```

Expected output: 

> Tom, Dick, and Harry

Actual output:

> Tom

With patch provided in this PR, the output should be:

> Tom, Dick., and Harry

## Related Issue(s)

- Could be related to: https://github.com/NVIDIA/NeMo-Guardrails/issues/882

## Checklist


schuellc-nvidia (approved):
Changes look good to me!

Ideally, we would also have a test that involves the flows `generating user intent for unhandled user utterance` and `continuation on unhandled user utterance` that are affected by it.

lapinek:
> Changes look good to me!
> 
> Ideally, we would also have a test that involves the flows `generating user intent for unhandled user utterance` and `continuation on unhandled user utterance` that are affected by it.

@schuellc-nvidia, thank you for the review! 

Would it be OK if we add these tests in a follow-up? I'd like to take a closer look at the dialog flows first to ensure meaningful coverage. I believe the already included [test](https://github.com/NVIDIA/NeMo-Guardrails/pull/1297/commits/c1471c1a15592536f5971a0b46f77168d4804adc) verifies that the rails transformations are applied. We're relying on this behavior in upcoming PRs.

Pouyanpi (approved):
Thank you @lapinek , this is a critical security fix with solid implementation 👍🏻 It is ready to merge 🚀

lapinek:
@Pouyanpi, thanks for your review! I've signed the commits. 

The reason I initially used "Signed-of-by..." was that it seemed like a valid alternative according to the contribution guidelines: https://github.com/NVIDIA/NeMo-Guardrails/blob/52ac7edc18e5b6fb1900b639b5c6734dca09b918/CONTRIBUTING.md#summary. Since GPG signature is required, maybe that wording could be more explicit. But I am happy to sign, no problem 🙂

PR #1234, opened 2025-06-19
feat: parallel rails execution

schuellc-nvidia:
## Description

This MR enables parallel input/output rail execution (in Colang 1) through a new action called 'run_flows_in_parallel'.

- Input and output rails can be configured to run in parallel through a new field in the bot config (parallel: True).
- Two additional actions are added (`run_input_rails_in_parallel`, `run_output_rails_in_parallel`) to simplify the parallel rails calling and to create the necessary tracing events: `StartInputRail`, `InputRailFinished`, `StartOutputRail`, `OutputRailFinished`.
- The parallel rail execution tries to mimic the sequential processing as much as possible. Therefore, the resulting event history, processing log and flow context will be almost the same and appear in sequential order, even if the event timestamps reveal the concurrent execution of the rails.

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/CONTRIBUTING.md) guidelines.
- [ ] I've updated the documentation if applicable.
- [x] I've added tests if applicable.
- [x] @mentions of the person or team responsible for reviewing proposed changes.

PR #1306, opened 2025-07-25
fix(rails): prevent LLM parameter contamination in concurrent paralle…

Pouyanpi:
## Description

In configurations with parallel rails execution, LLM parameters were being contaminated across concurrent actions. Specifically, max_tokens=3 from `self_check_output` rail was contaminating the main LLM, causing response truncation. This affected both streaming and non-streaming modes when rails executed concurrently.

## Root Cause

The llm_params context manager was modifying shared LLM state, leading to race conditions when multiple rails (input/output) executed in parallel.

## Fix

Implement LLM instance isolation by:

- Detecting actions that require 'llm' parameter via signature inspection
- Creating isolated shallow copies of the main LLM during initialization
- Each action gets its own LLM instance with independent model_kwargs
- Preserving shared resources (HTTP clients, credentials) through shallow copy

## Key Changes

- Auto-detection: Automatically identify actions needing isolated LLMs
- Shallow copy isolation: Independent parameter state, shared infrastructure
- Comprehensive testing: 17 unit tests + 9 e2e tests
- Graceful fallbacks: Handle copy failures and missing dependencies
- Performance optimized: Instance reuse vs per-request copying

## Testing

- ✅ All existing tests pass
- ✅ Bug reproduction test confirms fix
- ✅ Comprehensive test coverage for isolation functionality
- ✅ E2E tests verify no parameter contamination in parallel execution scenarios

resolves the issue seen in #1234


nemoguardrails/rails/llm/llmrails.py
```diff
@@ -507,6 +507,118 @@ def _init_llms(self):
...
+    def _create_isolated_llms_for_actions(self):
+        """Create isolated LLM copies for all actions that accept 'llm' parameter."""
+        if not self.llm:
+            log.debug("No main LLM available for creating isolated copies")
+            return
+
+        try:
+            actions_needing_llms = self._detect_llm_requiring_actions()
+            log.info(
+                f"{len(actions_needing_llms)} actions requiring isolated LLMs: {list(actions_needing_llms)}"
```
tgasser-nv: Recommend using %s in log messages rather than f-strings

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -507,6 +507,118 @@ def _init_llms(self):
 
         self.runtime.register_action_param("llms", llms)
 
+        # detect actions that need isolated LLM instances and create them
+        self._create_isolated_llms_for_actions()
```
tgasser-nv, line 511: Is there any risk of always creating isolated LLMs, even if we don't have parallel rails enabled? It's hard to isolate changes in this file since it inter-mingles so many different configs though

tests/test_llm_isolation.py
```diff
@@ -0,0 +1,453 @@
...
+        expected_actions = {
+            "action_with_llm",
+            "generate_user_intent",
+            "self_check_output",
+        }
+        assert actions_needing_llms == expected_actions
+
+        assert "action_without_llm" not in actions_needing_llms
+
+    def test_get_action_function(self, rails_with_mock_llm):
```
tgasser-nv: Could you split this test into its constituent cases? Otherwise it's not clear from the test-name only which one(s) are failing

tests/test_llm_isolation.py
```diff
@@ -0,0 +1,453 @@
...
+        rails.runtime.action_dispatcher = None
+
+        # should not crash
+        rails._create_isolated_llms_for_actions()
+
+
+class TestLLMIsolationInspection:
+    """Test signature inspection functionality."""
+
+    def test_signature_inspection_various_patterns(self):
```
tgasser-nv: Same as above, could you split into separate test-cases for each 5 action types below?

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -507,6 +507,118 @@ def _init_llms(self):
...
+    def _create_action_llm_copy(self, main_llm, action_name):
+        """Create an isolated copy of main LLM for a specific action."""
+        import copy
+
+        try:
+            # shallow copy to preserve HTTP clients, credentials, etc.
+            # but create new instance to avoid shared state
+            isolated_llm = copy.copy(main_llm)
+
+            # isolate model_kwargs to prevent shared mutable state
```
tgasser-nv, line 595: Are there any other attributes outside of `model_kwargs` that also need isolating? The only ones I could find were in llm/helpers.py line 61 (_modify_instance_kwargs())

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -507,6 +507,118 @@ def _init_llms(self):
...
+                and isolated_llm.model_kwargs is not None
+            ):
+                isolated_llm.model_kwargs = isolated_llm.model_kwargs.copy()
+            else:
+                isolated_llm.model_kwargs = {}
+
+            log.debug(
+                f"Successfully created isolated LLM copy for action: {action_name}"
+            )
+            return isolated_llm
```
tgasser-nv, line 607: I think this is of type `langchain_core.language_models.llms.{LLM, BaseLLM}`, it's hard to trace through. Could you add type annotations on this function?

Is there any way to declare configuration of one of these model classes as frozen (so any mutations are impossible by design)? Similar to the [@dataclass ](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass)`frozen` argument.

tgasser-nv (approved):
LGTM, please merge after cleaning up the testcases. The other comments were either nits or more general

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: Unfortunately, this is not currently possible because:
It is langchain's design where the BaseLLM and BaseChatModel classes are designed to be mutable
and for example llm_params() work by temporarily modifying the LLM state

of course this is supported by Pydantic (e.g. v2):

```python
from pydantic import BaseModel, ConfigDict

class ImmutableModel(BaseModel):
    model_config = ConfigDict(frozen=True)
    # rest of the model
```

but the problem is that langchain does not use it.

tests/test_llm_isolation.py
Pouyanpi: done

tests/test_llm_isolation.py
Pouyanpi: done

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: very good point and I think not, copy.copy() creates a shallow copy of the llm instance
for immutable attributes (temperature, max_tokens, etc which are float/int/bool), the shallow copy already provides isolation. When we do isolated_llm.temperature = 0.5, it creates a new reference on the copied instance, not affecting the original one.
but for mutable objects like dictionaries, the shallow copy shares the same reference, so `isolated_llm.model_kwargs["key"] = value` would affect the original llm

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: I don't see any except a minimal overhead but as you said having that distinction is a source of coupling. I think while the optimization makes sense, the current approach (always isolating) is simpler and more future proof. The overhead is minimal (shallow copies).

PR #1307, opened 2025-07-25
docs(tracing): update and streamline tracing guide

Pouyanpi:
[GENERATED BY COPILOT]

This pull request revises the tracing documentation for NeMo Guardrails, streamlining the configuration process and aligning it with OpenTelemetry best practices. Key changes include simplifying the tracing setup, emphasizing application-level SDK configuration, and providing detailed examples for production and development environments.

#### Simplification of Configuration:
- Removed outdated and overly detailed examples in `docs/user-guides/configuration-guide.md`. The new approach focuses on enabling tracing in `config.yml` with minimal configuration and highlights the FileSystem and OpenTelemetry adapters for development and production use.

#### New Tracing Guide:
- Added `docs/user-guides/tracing.md`, which provides a comprehensive guide to tracing, including quick start instructions, adapter types, OpenTelemetry integration, and setup examples for development and production environments.

#### OpenTelemetry Best Practices:
- Updated documentation to follow OpenTelemetry standards where libraries use only the API, and applications configure the SDK. Deprecated functions like `register_otel_exporter()` are noted, with migration examples provided.

#### Enhanced Ecosystem Compatibility:
- Detailed compatibility with OpenTelemetry exporters (e.g., Jaeger, Zipkin), collectors, and backends, including setup instructions for common integrations.

#### Support for Custom Adapters:
- Added examples for creating and registering custom `InteractionLogAdapter` classes for specialized backends or formats.

PR #1310, opened 2025-07-29
docs: restructure and edit content

miyoungc:
## Description

Doc work for https://github.com/NVIDIA/NeMo-Guardrails/pull/1307

## Related Issue(s)

https://github.com/NVIDIA/NeMo-Guardrails/issues/1287

## Checklist


docs/user-guides/tracing/common-integrations.md
```diff
@@ -0,0 +1,37 @@
+# Common Integrations
```
miyoungc: @Pouyanpi Can you provide more context about this, and why is this common integration? Is it common integration of what? And why do you run Docker suddenly, and how do you use it with NeMo Guardrails?

Pouyanpi: Excellent point, I agree and removed the entire section (in the target PR #1307). I think users configuring OpenTelemetry already understand trace backends and the document should focus on NeMo Guardrails specific configuration rather than explaining general OpenTelemetry ecosystem concepts 👍🏻

PR #1320, opened 2025-08-01
fix(tracing): use actual operation timestamps in OpenTelemetry spans

Pouyanpi:
## Description

OpenTelemetry spans were created with current timestamps during export instead of using he original timestamps from when operations actually occurred. 

**Fix**: 

Modified `_create_span()` to use `tracer.start_span()` with original `start_time` and `span.end()` with original `end_time` from span data, converting timestamps to nanoseconds as required by OpenTelemetry.

 **Test Plan:** 
 
 `tests/test_opentelemetry_timing_behavior.py` contains timing-focused tests that verify spans use original operation timestamps (not current time). This test file is for verification and will be removed after review. The test fails on develop branch.

### DIFF
--- a/nemoguardrails/tracing/adapters/opentelemetry.py
+++ b/nemoguardrails/tracing/adapters/opentelemetry.py
@@ -55,7 +55,7 @@ from __future__ import annotations
 
 import warnings
 from importlib.metadata import version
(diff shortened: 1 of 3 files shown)


Pouyanpi:
close in favor of #1331

PR #1323, opened 2025-08-04
fix(runtime): stop execution after internal error

Pouyanpi:
## Description

Add a "BotIntent" event with intent "stop" to the event list after an internal error occurs. This prevents further LLM generation and ensures execution halts appropriately following an error.

## Test Plan

please run https://gist.github.com/Pouyanpi/eef2a71efc16fe6976f8e15db91717cf against develop branch and after this fix.

This demonstrates the fix using a configuration which has:
- Parallel input rails: content safety + self check
- Parallel output rails: content safety + self check
- NIM models for content safety

PROBLEM (Before Fix):
- each failing rail generates 'I'm sorry, an internal error has occurred.'
- with 4 rails, user could see 4 error messages

SOLUTION (After Fix):
- first error generates BotIntent stop event
- stop event cancels remaining rails
- user sees only one error message

### DIFF
--- a/nemoguardrails/colang/v1_0/runtime/runtime.py
+++ b/nemoguardrails/colang/v1_0/runtime/runtime.py
@@ -404,6 +404,15 @@ class RuntimeV1_0(Runtime):
         except Exception as e:
             log.error(f"Error in parallel rail execution: {str(e)}")
             raise
+        finally:
+            # clean up any remaining cancelled tasks to avoid "Task was destroyed but it is pending" warnings
+            for task in tasks:
+                if not task.done():
+                    task.cancel()
+                    try:
+                        await task
+                    except asyncio.CancelledError:
+                        pass
 
         context_updates: dict = {}
         processing_log = processing_log_var.get()
--- a/nemoguardrails/colang/v1_0/runtime/runtime.py
+++ b/nemoguardrails/colang/v1_0/runtime/runtime.py
@@ -272,6 +272,11 @@ class RuntimeV1_0(Runtime):
                 },
                 # We also want to hide this from now from the history moving forward
                 {"type": "hide_prev_turn"},
+                # Stop execution to prevent further LLM generation after internal error
+                {
+                    "type": "BotIntent",
+                    "intent": "stop",
+                },
             ]
         )
 
--- a/tests/test_internal_error_parallel_rails.py
+++ b/tests/test_internal_error_parallel_rails.py
@@ -1,0 +1,354 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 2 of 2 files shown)


nemoguardrails/colang/v1_0/runtime/runtime.py
```diff
@@ -272,6 +272,11 @@ def _internal_error_action_result(message: str):
                 },
                 # We also want to hide this from now from the history moving forward
                 {"type": "hide_prev_turn"},
+                # Stop execution to prevent further LLM generation after internal error
+                {
+                    "type": "BotIntent",
```
tgasser-nv, line 277: Not a fan of the stringly typed `type` and `intent` fields here (should be enums) but we can fix this later

nemoguardrails/colang/v1_0/runtime/runtime.py
```diff
@@ -399,6 +404,15 @@ async def task_call_helper(flow_uid, post_event, func, *args, **kwargs):
         except Exception as e:
             log.error(f"Error in parallel rail execution: {str(e)}")
             raise
+        finally:
+            # clean up any remaining cancelled tasks to avoid "Task was destroyed but it is pending" warnings
+            for task in tasks:
+                if not task.done():
```
tgasser-nv, line 410: Just want to check if we cancel a task that's already done it won't throw an exception? I remember you said no in a previous review. It looks like that's what lines 412 - 415 are doing. There could be a race condition between checking if it's done and trying to cancel it

tests/test_internal_error_parallel_rails.py
```diff
@@ -0,0 +1,354 @@
...
+        chat = TestChat(config, llm_completions=["Hello!"])
+        chat >> "hi"
+
+        result = await chat.app.generate_async(messages=chat.history, options=OPTIONS)
+
+        # should get an internal error response
+        assert result is not None
+        assert "internal error" in result.response[0]["content"].lower()
+
+        # check that a BotIntent stop event was generated in the internal events
```
tgasser-nv, line 64: Can we also check we don't send the App LLM request in this case?

tests/test_internal_error_parallel_rails.py
```diff
@@ -0,0 +1,354 @@
...
+        "instructions": [
+            {"type": "general", "content": "You are a helpful assistant."}
+        ],
+        "models": [
+            {"type": "main", "engine": "openai", "model": "gpt-3.5-turbo"},
+            {"type": "content_safety", "engine": "openai", "model": "gpt-3.5-turbo"},
+        ],
+        "rails": {
+            "input": {
+                "flows": ["content safety check input $model=content_safety"],
```
tgasser-nv: Could you add another flow here? Even though parallel is set to True we only have one flow

tests/test_internal_error_parallel_rails.py
```diff
@@ -0,0 +1,354 @@
...
+            "Hi there! How can I help?",  # main response
+            "No",  # self check output passes
+        ],
+    )
+
+    chat >> "hi"
+    result = await chat.app.generate_async(messages=chat.history, options=OPTIONS)
+
+    assert result is not None
+    print(result.response)
```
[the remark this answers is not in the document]
tgasser-nv: ^^ Yep we should get rid of the print statement

nemoguardrails/colang/v1_0/runtime/runtime.py
Pouyanpi: No it does not:

```bash
python -c "
import asyncio

async def test_cancel_done_task():
    async def simple_task():
        await asyncio.sleep(0.1)
        return 'completed'
    
    task = asyncio.create_task(simple_task())
    result = await task
    
    print(f'Task done: {task.done()}')
    print(f'Task result: {result}')
    
    try:
        cancelled = task.cancel()
        print(f'Cancel returned: {cancelled}')
        print('No exception thrown when cancelling done task')
    except Exception as e:
        print(f'Exception: {e}')

asyncio.run(test_cancel_done_task())
"
```

PR #1334, opened 2025-08-15
fix(prompts): prevent IndexError when LLM provided via constructor wi…

Pouyanpi:
## Description
When providing a Main LLM model object directly to the `LLMRails` constructor while having an empty models list in the YAML config, the system would throw an `IndexError`:

```
File "/nemoguardrails/llm/prompts.py", line 142, in get_task_model
    return _models[0]
           ~~~~~~~^^^
IndexError: list index out of range
```

## Root Cause
The `get_task_model` function in `prompts.py` was attempting to access the first element of the `_models` list without checking if it was empty. This occurred when:
1. An LLM was provided via the constructor (not in config)
2. The config had an empty models list `models: []`
3. The system tried to find a model for prompt selection

## Solution
Implemented a tactical fix by adding a safety check before accessing the list:
- Check if `_models` list is not empty before accessing `_models[0]`
- Return `None` when no matching models are found
- The existing code already handles `None` return gracefully by defaulting to "unknown" model

## Notes
This is a tactical/bandaid fix as discussed in the bug report.


cparisien (approved):

tests/test_llm_task_manager.py
```diff
@@ -457,3 +457,59 @@ def test_reasoning_traces_not_included_in_prompt_history():
...
+    )
+
+    result = get_task_model(config, "main")
+    assert result is None
+
+
+def test_get_task_model_with_main_model():
+    """Test that get_task_model returns the main model when present."""
+    config = RailsConfig.parse_object(
+        {"models": [{"type": "main", "engine": "openai", "model": "gpt-3.5-turbo"}]}
```
tgasser-nv: Could you add a couple more models and put the `main` last?

nemoguardrails/llm/prompts.py
```diff
@@ -139,7 +139,8 @@ def get_task_model(config: RailsConfig, task: Union[str, Task]) -> Model:
         if not _models:
             _models = [model for model in config.models if model.type == "main"]
 
-        return _models[0]
+        if _models:
+            return _models[0]
 
     return None
```
tgasser-nv, line 145: Could you update the signature to return Optional[Model]?

tests/test_llm_task_manager.py
Pouyanpi: Done!

PR #1336, opened 2025-08-15
fix(llmrails): handle LLM models without model_kwargs field in isolation

Pouyanpi:
## Description
Fix LLM isolation for models without model_kwargs

## Problem:
- `ChatNVIDIA` and other strict Pydantic models fail with "ChatNVIDIA" object has no field "model_kwargs" when creating isolated LLM instances for actions
- `FakeLLM` test failures due to independent counters in copied instances

### Solution:

- Only attempt to copy model_kwargs if it exists and is not None
- Don't try to set `model_kwargs = {}` on models that don't support it
- `FakeLLM` now shares response counter across copies to maintain test sequence integrity

## Changes:

- nemoguardrails/rails/llm/llmrails.py: simplified `model_kwargs` handling in `_create_action_llm_copy`
- tests/utils.py: added shared state counter to `FakeLLM` for consistent test behavior

This fixes the "Failed to create isolated LLM instance" errors when using ChatNVIDIA or other models without model_kwargs field.

### DIFF
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -593,8 +593,6 @@ class LLMRails:
                 and isolated_llm.model_kwargs is not None
             ):
                 isolated_llm.model_kwargs = isolated_llm.model_kwargs.copy()
-            else:
-                isolated_llm.model_kwargs = {}
 
             log.debug(
                 "Successfully created isolated LLM copy for action: %s", action_name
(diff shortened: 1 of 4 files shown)


Pouyanpi:
> The fix looks good. Could you manually run tests for an Nvidia LLM and OpenAI LLM and copy the commands and output into the PR? Just to make sure there's no other exception text?

Just updated the PR descriptions 👍🏻

PR #1339, opened 2025-08-18
test(cli): add comprehensive CLI test suite and reorganize files

Pouyanpi:
tests for nemoguardrails CLI

requires #1343

### DIFF
--- a/tests/cli/test_chat.py
+++ b/tests/cli/test_chat.py
@@ -109,61 +109,67 @@ class TestChatState:
 
 
 class TestRunChat:
-    @patch("asyncio.run")
-    @patch.object(chat_module, "LLMRails")
-    @patch.object(chat_module, "RailsConfig")
-    def test_run_chat_v1_0(self, mock_rails_config, mock_llm_rails, mock_asyncio_run):
-        mock_config = MagicMock()
-        mock_config.colang_version = "1.0"
-        mock_rails_config.from_path.return_value = mock_config
+    def test_run_chat_v1_0(self):
+        with patch.object(
+            chat_module, "RailsConfig"
+        ) as mock_rails_config, patch.object(
+            chat_module, "LLMRails"
+        ) as mock_llm_rails, patch(
+            "asyncio.run"
+        ) as mock_asyncio_run:
+            mock_config = MagicMock()
+            mock_config.colang_version = "1.0"
+            mock_rails_config.from_path.return_value = mock_config
 
-        run_chat(config_path="test_config")
-
-        mock_rails_config.from_path.assert_called_once_with("test_config")
-        mock_asyncio_run.assert_called_once()
-
-    @patch.object(chat_module, "get_or_create_event_loop")
-    @patch.object(chat_module, "LLMRails")
-    @patch.object(chat_module, "RailsConfig")
-    def test_run_chat_v2_x(self, mock_rails_config, mock_llm_rails, mock_get_loop):
-        mock_config = MagicMock()
-        mock_config.colang_version = "2.x"
-        mock_rails_config.from_path.return_value = mock_config
-
-        mock_loop = MagicMock()
-        mock_get_loop.return_value = mock_loop
-
-        run_chat(config_path="test_config")
-
-        mock_rails_config.from_path.assert_called_once_with("test_config")
-        mock_llm_rails.assert_called_once_with(mock_config, verbose=False)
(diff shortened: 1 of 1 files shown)


tests/cli/test_chat.py
```diff
@@ -0,0 +1,315 @@
...
+    ChatState,
+    extract_scene_text_content,
+    parse_events_inputs,
+    run_chat,
+)
+
+chat_module = sys.modules["nemoguardrails.cli.chat"]
+
+
+class TestParseEventsInputs:
```
tgasser-nv, line 32: Is the full set of events and their formats documented anywhere? How do we know we covered all of them in this class?

tests/cli/test_chat.py
```diff
@@ -0,0 +1,315 @@
...
+    def test_parse_invalid_event(self):
+        result = parse_events_inputs("Invalid.Event.Format.TooMany")
+        assert result is None
+
+    def test_parse_event_missing_equals(self):
+        result = parse_events_inputs("UserAction(invalid_param)")
+        assert result is None
+
+
+class TestExtractSceneTextContent:
```
tgasser-nv, line 66: Same as `TestParseEventsInputs`, without some documentation it's hard to say whether we're testing correctly or missing coverage

tests/cli/test_chat.py
```diff
@@ -0,0 +1,315 @@
...
+        assert chat_state.input_events == []
+        assert chat_state.output_events == []
+        assert chat_state.output_state is None
+        assert chat_state.events_counter == 0
+        assert chat_state.first_time is False
+
+
+class TestRunChat:
+    @patch("asyncio.run")
+    @patch.object(chat_module, "LLMRails")
```
tgasser-nv: Is the same `chat_module` used throughout all these tests? Should we create it from scratch in each test to avoid mutations in a previous test changing the response?

tests/cli/test_cli_main.py
```diff
@@ -0,0 +1,338 @@
...
+    @patch("uvicorn.run")
+    @patch("nemoguardrails.server.api.app")
+    @patch("os.path.exists")
+    @patch("os.path.expanduser")
+    def test_server_with_config(
+        self, mock_expanduser, mock_exists, mock_app, mock_uvicorn
+    ):
+        mock_expanduser.return_value = "/path/to/config"
+        mock_exists.return_value = True
+        result = runner.invoke(app, ["server", "--config=/path/to/config"])
```
tgasser-nv, line 184: Shouldn't we pass in `~` as the directory here, to test the expanduser call expands this to "/path/to/config"?

tests/cli/test_debugger.py
```diff
@@ -0,0 +1,233 @@
...
+        mock_tree_instance = MagicMock()
+        mock_tree.return_value = mock_tree_instance
+
+        result = runner.invoke(debugger.app, ["tree"])
+
+        assert result.exit_code == 0
+        mock_tree.assert_called_with("main")
+
+
+# TODO: addd tests
```
tgasser-nv: Are there tests missing? Recommend removing this rather than commenting-out

tests/cli/test_migration.py
```diff
@@ -0,0 +1,686 @@
...
+    _revise_anonymous_flow,
+    _write_to_file,
+    _write_transformed_content_and_rename_original,
+    convert_colang_1_syntax,
+    convert_colang_2alpha_syntax,
+    migrate,
+)
+
+
+class TestColang2AlphaSyntaxConversion:
```
tgasser-nv, line 42: Is there a formal definition of Colang 2 syntax anywhere? Is there a Colang 2 Alpha subset of this? I can't figure out whether these tests are doing the right thing or not

tests/cli/test_cli_main.py
Pouyanpi: we are mocking it anyhow, so I don't think that's something that needs to be tested here.

tests/cli/test_chat.py
Pouyanpi: `parse_events_input` is a general utility and not tied to any particular event(s). we are testing its expected behavior for both valid and invalid cases.

tests/cli/test_chat.py
Pouyanpi: same; the utility works for any List[dict] and just combines the text field of each item.

PR #1340, opened 2025-08-18
feat: Modify endpoints for OpenAPI compatibility

christinaexyou:
## Description
This PR modifies the existing `/v1/chat/completions` and adds a `/v1/models` endpoint to be OpenAI compatible.

Changes: 
* `nemoguardrails/server/api.py` - Introduces new classes that correspond to OpenAI's standard chat completion and models API
* `tests/test_api.py`, `tests/test_server_calls_with_state.py` & `tests/test_threads` - Updates the message content path according to new `ResponseBody` class and adds new tests for the `/v1/models` API

## Checklist


christinaexyou:
Updated `nemoguardrails/colang/v2_x/runtime/runtime.py` and `tests/test_server_calls_with_state.py` because tests were failing due to Colang 2.x not supporting `assistant` messages as input. Fixed by adding logic to convert `State` object to dict and passing in state instead of assistant cc: @Pouyanpi @tgasser-nv

RobGeada:
please don't merge- we want to refactor this a little more

nemoguardrails/server/api.py
```diff
@@ -207,10 +208,53 @@ class RequestBody(BaseModel):
         default=None,
         description="A state object that should be used to continue the interaction.",
     )
+    # Standard OpenAI completion parameters
```
RobGeada: Let's bring the OpenAI schema into a separate file- perhaps `server/schemes/openai`

nemoguardrails/server/api.py
```diff
@@ -207,10 +208,53 @@ class RequestBody(BaseModel):
...
+    )
+    temperature: Optional[float] = Field(
+        default=None,
+        description="Sampling temperature to use.",
+    )
+    top_p: Optional[float] = Field(
+        default=None,
+        description="Top-p sampling parameter.",
+    )
+    stop: Optional[str] = Field(
```
RobGeada: `stop` needs to be `Optional[Union[str, List[str]]]`

nemoguardrails/server/api.py
```diff
@@ -231,25 +275,113 @@ def ensure_config_ids(cls, v, values):
         return v
 
 
+class Choice(BaseModel):
+    index: Optional[int] = Field(
+        default=None, description="The index of the choice in the list of choices."
+    )
+    messages: Optional[dict] = Field(
```
RobGeada: typo: `messages` needs to be `message` (no s)

RobGeada (approved):

Pouyanpi:
thank you @christinaexyou . I like the direction of making the api feel openai-compatible, but before we can call it fully compatible, a few foundational pieces would be helpful to add:


**version compatibility and testing**
* document which openai sdk versions this targets (eg: openai>=1.0.0,<2.0.0) and we need to figure out a way to test againts that matrix, e.g. adding that range to pyproject.toml dev deps so CI tests against it. without explicit version constraints, upstream sdk changes could introduce silent breakage.
* add integration tests using a real `openai.OpenAI(base_url=...)` client to confirm both non-streaming and streaming behave correctly end-to-end. this gives us direct proof of compatibility rather than relying on assumed formats.

**schema alignment**
* the PR currently re defines several schemas (Choice, ResponseBody, etc.). I was thinking if it is possible to extend existing types like `openai.types.chat.ChatCompletion` instead, and then adding Guardrails specific fields (ex: state, log). this reduces maintenance and keeps us aligned with openai’s evolving schema automatically. But probably there are some concerns that you have avoided it, I'd happy to learn about those.
* one concrete example: stop currently only supports a single string, but openai accepts both strings and lists. using their types would surface issues like this early.
* for ongoing validation, we could also verify responses against the openapi spec: https://github.com/openai/openai-openapi

**streaming format**
* openai streaming expects SSE with `ChatCompletionChunk` objects, prefixed with `data:`  and ending in `data: [DONE]`. reference: https://github.com/openai/openai-python/blob/main/src/openai/types/chat/chat_completion_chunk.py
* worth confirming the current streaming response follows this format exactly, since it determines whether client.chat.completions.create(..., stream=True) works out of the box. I know that the streaming functionality was broken on the server.

**documentation and scope**
* adding a feature support matrix would help clarify what’s currently supported (basic chat, streaming) vs not yet (**multimodal**, function calling, tools, response_format, etc.), and why. that sets clear expectations for users adopting this. (fyi, we support multimodal guardrails and models)


once we have explicit versioning, schema alignment, confirmed streaming format, integration tests, and a clear scope matrix, we’ll be in a solid position to confidently say “openai compatible.” happy to help with any of these pieces.

nemoguardrails/server/api.py
```diff
@@ -229,10 +228,53 @@ class RequestBody(BaseModel):
         default=None,
         description="A state object that should be used to continue the interaction.",
     )
+    # Standard OpenAI completion parameters
+    model: Optional[str] = Field(
+        default=None,
+        description="The model to use for chat completion. Maps to config_id for backward compatibility.",
```
tgasser-nv: I think it would be better to have the `model` field to match the `main` model from the Guardrails configuration, rather than a Guardrails config ID. Then the same /chat/completions request can be POSTed to either the main LLM directly or Guardrails without any changes. On build.nvidia.com the `guardrails` field is ignored, and Guardrails can set a default config ID if the `guardrails` field isn't included

nemoguardrails/server/api.py
```diff
@@ -253,25 +295,44 @@ def ensure_config_ids(cls, v, values):
...
-    )
-    log: Optional[GenerationLog] = Field(
-        default=None, description="Additional logging information."
-    )
-    state: Optional[dict] = Field(
-        default=None,
-        description="A state object that should be used to continue the interaction in the future.",
-    )
+@app.get(
+    "/v1/models",
```
tgasser-nv: ~~This looks good, to the comment above I recommend including the `main` model from each Guardrail configuration and adding an extra field to the list of `Model`s. So for example the [nemoguards configuration](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/examples/configs/nemoguards/config.yml) which currently looks like:~~

```
{
  "object": "list",
  "data": [
    {
      "id": "nemoguards",
      "created": 1764625695,
      "object": "model",
      "owned_by": "nemo-guardrails"
    },
...
}
```

~~You'd have:~~

```
{
  "object": "list",
  "data": [
    {
      "id": "meta/llama-3.3-70b-instruct",
      "guardrails_config_id": "nemoguards",
      "created": 1764625695,
      "object": "model",
      "owned_by": "nemo-guardrails"
    },
...
}
```

~~This can be used in a `/chat/completions` request, where the `id` is used in the `model` field, and `guardrails_config_id` is used in `guardrails.config_id`~~

EDITED I don't think it makes sense to combine guardrails configs to a specific LLM

tests/server/test_api.py
```diff
@@ -59,8 +82,14 @@ def test_chat_completion():
     )
     assert response.status_code == 200
     res = response.json()
-    assert len(res["messages"]) == 1
-    assert res["messages"][0]["content"]
+    # Check OpenAI-compatible response structure
```
tgasser-nv, line 76: Could you update the e2e tests to use the `LIVE_TEST_MODE` approach to be consistent with the [test_llm_params_e2e.py](https://github.com/NVIDIA-NeMo/Guardrails/blob/d3fb3d6da39a4d2cfbe5bf405c0842d4d271eda8/tests/test_llm_params_e2e.py#L72)

tests/test_openai_integration.py
```diff
@@ -0,0 +1,167 @@
...
+        http_client=test_client,
+    )
+    return client
+
+
+def test_openai_client_list_models(openai_client):
+    models = openai_client.models.list()
+
+    # Verify the response structure matches OpenAI's ModelList
+    assert models is not None
```
tgasser-nv: nit: Would it be simpler to check against an expected dict than each individual field ?

tests/test_openai_integration.py
```diff
@@ -0,0 +1,167 @@
...
+
+def test_openai_client_chat_completion(openai_client):
+    response = openai_client.chat.completions.create(
+        model="config_1",
+        messages=[{"role": "user", "content": "hi"}],
+        stream=False,
+    )
+
+    # Verify response structure matches OpenAI's ChatCompletion object
+    assert response is not None
```
tgasser-nv: nit: Would it be clearer to check against an expected dict rather than iterating over each of the attributes?

nemoguardrails/server/api.py
christinaexyou: Updated the `model` field to match the `main` model in the config. If there is no `main` model defined, it defaults to `config_id[0]`

tests/test_openai_integration.py
christinaexyou: Done - we check that the returned `Choice` response matches an expected dict and check the id and created timestamp separately since they are variable to change

nemoguardrails/server/api.py
christinaexyou: Done - updated `id` match the model field and extended the existing OpenAI Model API to have a field named `guardrail_config_id`. If there are no models defined in the config, the `id` field defaults to `config_id[0]`

christinaexyou:
@Pouyanpi @tgasser-nv thanks so much for your feedback !  

By default, the `nim` engine disables streaming by default so I added a method in `llmrails.py` called `_configure_main_llm_streaming` to enable streaming if it's set to `True` in the config. 

There's quite a lot of commits in this PR that don't necessarily relate to the original feature, so should they be kept as separate commits or do you still want me to squash and rebase ?

Pouyanpi:
Thank you @christinaexyou !

> By default, the `nim` engine disables streaming by default so I added a method in `llmrails.py` called `_configure_main_llm_streaming` to enable streaming if it's set to `True` in the config.

Unfortunately we had a bug which is fixed now. I add some detailed comments .

nemoguardrails/server/api.py
Pouyanpi: +1

@christinaex could you please apply the same approach you've already used for the response schemas to the request schemas as well?

nemoguardrails/server/schemas/openai.py
```diff
@@ -0,0 +1,51 @@
...
+"""OpenAI API schema definitions for the NeMo Guardrails server."""
+
+from typing import List, Optional
+
+from openai.types.chat.chat_completion import ChatCompletion
+from openai.types.model import Model
+from pydantic import BaseModel, Field
+
+
+class ResponseBody(ChatCompletion):
```
Pouyanpi:
```suggestion
class GuardrailsChatCompletion(ChatCompletion):
```

nemoguardrails/server/schemas/openai.py
```diff
@@ -0,0 +1,51 @@
...
+class GuardrailsModel(Model):
+    """OpenAI API model with NeMo-Guardrails extensions."""
+
+    guardrails_config_id: Optional[str] = Field(
+        default=None,
+        description="[NeMo Guardrails extension] The guardrails configuration ID associated with this model.",
+    )
+
+
+class ModelsResponse(BaseModel):
```
Pouyanpi: not sure about this, but let's be consistent:
```suggestion
class GuardrailsModels(BaseModel):
```

nemoguardrails/server/api.py
```diff
@@ -228,10 +233,53 @@ class RequestBody(BaseModel):
         default=None,
         description="A state object that should be used to continue the interaction.",
     )
+    # Standard OpenAI completion parameters
+    model: Optional[str] = Field(
```
Pouyanpi: model MUST be provided. It is a required field. It should be the model name that we currently use in the guardrails configuration. Using the request body, we set the main LLM; other types of models come from the config (config.yml).

Guardrails and OpenAI compatibility with respect to the model field brings some challenges:

1. One is that Guardrails requires an engine in addition to the model.

	* There are various possible solutions, like posting model metadata (including `engine` and `base_url`) to the models endpoint, where this becomes the model ID we use for lookup. This is possible but tricky.
	* Another option is to set a default engine and make it configurable via an environment variable. Cons: a user cannot experiment with multiple models from multiple providers while controlling for one `config_id`.
	* We could also settle on a global models file (I recommend avoiding this).
2. Another challenge is that such models may be hosted in different places, thus requiring different base_urls. For example, I can have 3 different NIMs with 3 different `base_url`s. How are we going to address that?

TL;DR: We need a mechanism to obtain <engine, base_url> from a given model name (ID).

What do you think?

nemoguardrails/server/api.py
```diff
@@ -228,10 +233,53 @@ class RequestBody(BaseModel):
...
+    )
+    log_probs: Optional[bool] = Field(
+        default=None,
+        description="Log probabilities parameter.",
+    )
 
     @root_validator(pre=True)
     def ensure_config_id(cls, data: Any) -> Any:
         if isinstance(data, dict):
+            if data.get("model") is not None and data.get("config_id") is None:
```
Pouyanpi: I'm not sure what is happening here.

model should never be None and we always override the main model defined in config.yml using the model provided via request body.

nemoguardrails/server/api.py
Pouyanpi: Can you help me understand the intent here? I’m not sure I follow, doesn’t this effectively bind a model to a config?

As a side note: config_id must always be set, potentially with a configurable default.

cc @tgasser-nv

nemoguardrails/server/schemas/openai.py
Pouyanpi, line 1: Thank you 🥇 

Let's do similar for Response schemas.

nemoguardrails/server/schemas/openai.py
```diff
@@ -0,0 +1,51 @@
...
+    """OpenAI API response body with NeMo-Guardrails extensions."""
+
+    guardrails_config_id: Optional[str] = Field(
+        default=None,
+        description="The guardrails configuration ID associated with this response.",
+    )
+    state: Optional[dict] = Field(default=None, description="State object for continuing the conversation.")
+    llm_output: Optional[dict] = Field(default=None, description="Additional LLM output data.")
+    output_data: Optional[dict] = Field(default=None, description="Additional output data.")
+    log: Optional[dict] = Field(default=None, description="Generation log data.")
```
Pouyanpi, line 37:
```python
    guardrails_config_ids: Optional[List[str]] = Field(default=None, description="The list of configuration ids that were used.")
    state: Optional[dict] = Field(default=None, description="State object for continuing the conversation.")
    llm_output: Optional[dict] = Field(default=None, description="Contains any additional output coming from the LLM.")
    output_data: Optional[dict] = Field(
        default=None,
        description="The output data, i.e. a dict with the values corresponding to the `output_vars`.",
    )

    log: Optional[GenerationLog] = Field(default=None, description="Additional logging information.")
```
nit:
re `config_ids` vs `guardrails_config_ids` is there any reason that we should explicitly mention `guardrails`? If not better to rename it to be consistent with the request schema. But I don't mind either name

nemoguardrails/server/api.py
```diff
@@ -383,19 +558,33 @@ async def chat_completion(body: RequestBody, request: Request):
...
         log.exception(ex)
         return ResponseBody(
-            messages=[
-                {
-                    "role": "assistant",
-                    "content": f"Could not load the {config_ids} guardrails configuration. "
-                    f"An internal error has occurred.",
-                }
-            ]
+            id=f"chatcmpl-{uuid.uuid4()}",
```
Pouyanpi: I think these sorts of transformation/conversions on request and response fits nicely to a utils module in schemas.

a simple function with
```
input:    response: GenerationResponse, config_ids=None
output: `GuardrailCompletionResponse`
```
same for streaming.

Also you can include [this logic](https://github.com/NVIDIA-NeMo/Guardrails/pull/1340/files#diff-e40b79fa06e87647988379efc0f1f3aa948eb153fdb73efc68989bdae0993d75R626-R644) there.


merely a suggestion, what do you think?

nemoguardrails/server/api.py
```diff
@@ -355,6 +462,73 @@ def _get_rails(config_ids: List[str]) -> LLMRails:
     return llm_rails
 
 
+async def _format_streaming_response(
```
Pouyanpi, line 345: for streaming it is better to use `stream_async` method of LLMRails. You can still iterate on it.

This allows you to support another feature. You can run ouput rails while streaming. To do so you can use/adapt following pieces:

```python
class ErrorDetails(BaseModel):
    message: str
    type: str
    param: str
    code: str


class ErrorData(BaseModel):
    error: ErrorDetails
```

for now define them here later I will include them in llmrails.py and will update this code.

```python
def process_chunk(chunk: str) -> Union[str, ErrorData]:
    """
    Processes a single chunk from the stream.

    Args:
        chunk (str): A single chunk from the stream.

    Returns:
        Union[str, ErrorData]: ErrorData instance for errors or the original chunk.
    """
    try:
        validated_data = ErrorData.model_validate_json(chunk)
        return validated_data  # Return the ErrorData instance directly
    except ValidationError:
        # Not an error, just a normal token
        pass
    except json.JSONDecodeError:
        # Invalid JSON format, treat as normal token
        pass
    except Exception as e:
        log.warning(f"Unexpected error processing stream chunk: {type(e).__name__}: {str(e)}", extra={"chunk": chunk})
```

then when you are iterating over the stream at the begining of your try/except block:
```python
                async for chunk in stream_iter:
                    processed_chunk = process_chunk(chunk)
                    if isinstance(processed_chunk, ErrorData):
                        # Yield the error and stop streaming
                        yield f"data: {processed_chunk.model_dump_json()}\n\n"
                        return
````

File: nemoguardrails/streaming.py
Comment (Pouyanpi) [inline nemoguardrails/streaming.py:1]: Is it the changes you made re broken streaming in NIM engine? or there are other reasons?

I think it won't be necessary if you use stream_async. You can revert it locally when you made the changes to the streaming and let me know.

File: pyproject.toml
Comment (Pouyanpi) [inline pyproject.toml:1]: we can do this at the end.

we should add an extra dependency called `server` with these two ( similar to tracing, nvidia etc.)

File: nemoguardrails/server/api.py (hunk 5 above)
Comment (tgasser-nv) [reply nemoguardrails/server/api.py]: @Pouyanpi That's a good point. We already have the option to set a default config so I don't think it makes sense to tie an application LLM to a specific Guardrails config.

File: docs/user-guides/community/openai.md [hunk 16]
```diff
@@ -0,0 +1,16 @@
+## OpenAI API Compatibility for NeMo Guardrails
+
+NeMo Guardrails provides server-side compatibility with OpenAI API endpoints, enabling applications that use OpenAI clients to seamlessly integrate with NeMo Guardrails for adding guardrails to LLM interactions. Point your OpenAI client to `http://localhost:8000` (or your server URL) and use the standard `/v1/chat/completions` endpoint.
```
tgasser-nv: Could you add some more information here on:

* The `/v1/models` endpoint.
* Example JSON requests for the new OpenAI endpoints (`/v1/models` and `/v1/chat/completions`)

Longer-term, we can auto-generate OpenAPI HTML pages for this (see http://0.0.0.0:8000/docs for example). But not needed for this PR.

nemoguardrails/server/schemas/openai.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class GuardrailsModel(Model):
+    """OpenAI API model with NeMo-Guardrails extensions."""
+
+    guardrails_config_id: Optional[str] = Field(
```
tgasser-nv: Sorry for the confusion, I don't think we need this in the `/v1/models` response and can remove it. Let's return ModelsResponse only from a GET to `/v1/models`

nemoguardrails/server/api.py
tgasser-nv: Echoing the discussion on Slack, I think having environment variables for the `base_url`, `engine`, and `api_key` would restrict us to using one LLM provider but is quicker to implement for v0.2

nemoguardrails/server/api.py
```diff
@@ -355,6 +462,73 @@ def _get_rails(config_ids: List[str]) -> LLMRails:
...
+    streaming_handler: StreamingHandler, model_name: Optional[str]
```
tgasser-nv: I'm not sure `model_name` should be Optional here. Are there any situations where we don't know the model's name?

nemoguardrails/server/schemas/openai.py
tgasser-nv: I agree we don't need the guardrails prefix, let's use `config_id` rather than `guardrails_config_id`

christinaexyou:
@Pouyanpi @tgasser-nv thanks for taking the time to re-review ! The main changes made in this commit (and I also directly responded to some of your comments to provide more clarity):
* Renamed OpenAI API schema definitions
* Added `base_url`, `engine`, and `api_key_env_var` fields to `GuardrailsModel` which are populated by environment variables at runtime
* Added a new file, `nemoguardrails/server/schemas/utils.py` to handle formatting for streaming chat generation and added unit tests in `tests/test_schema_utils.py`
* Updated the docs to include example requests to the endpoints

nemoguardrails/server/api.py
christinaexyou: Done - see [nemoguardrails/server/schemas/utils.py](https://github.com/NVIDIA-NeMo/Guardrails/blob/d0b047429bbde6acb1d4f052283a6d17675366f0/nemoguardrails/server/schemas/utils.py) and corresponding unit tests [tests/test_schema_utils.py](https://github.com/NVIDIA-NeMo/Guardrails/blob/d0b047429bbde6acb1d4f052283a6d17675366f0/tests/test_schema_utils.py)

nemoguardrails/server/api.py
christinaexyou: Done - see [lines 470-543](https://github.com/NVIDIA-NeMo/Guardrails/blob/d0b047429bbde6acb1d4f052283a6d17675366f0/nemoguardrails/server/api.py#L470C1-L543C17) in `nemoguardrails/server/api.py`

nemoguardrails/server/schemas/openai.py
christinaexyou: Updated fields to use `config_id`

docs/user-guides/community/openai.md
christinaexyou: Done - also added an example using the OpenAI Python client

Pouyanpi:
@christinaexyou I've included further suggestions in https://github.com/christinaexyou/NeMo-Guardrails/pull/1

Pouyanpi:
@christinaexyou thanks, here are the parts that need update:

- the default values for the environment variables (nim, openai?)
- for the docs we need to update https://github.com/NVIDIA-NeMo/Guardrails/tree/develop/docs/run-rails/using-fastapi-server and the changes do not belong to `community` section. @miyoungc would you please guide us on the doc changes?

PR #1343, opened 2025-08-19
fix(logging): handle missing id and task in verbose logs

Pouyanpi:
Update verbose logging to safely handle cases where log records may not have 'id' or 'task' attributes. Prevents potential `AttributeError` and improves robustness of LLM and prompt log output formatting.


Pouyanpi:
required by #1339

tgasser-nv (approved):
Looks good. Lines 57-8 and 71-2 are duplicated but it's probably not worth refactoring such a small code snippet

PR #1346, opened 2025-08-22
fix(jailbreak): handle URL joining with/without trailing slashes

Pouyanpi:
## Description

Fix URL joining in NIM jailbreak detection to handle base URLs with/without trailing slashes and classification paths with/without leading slashes.

  - Add `join_nim_url()` helper function
  - Update `jailbreak_nim_request()` to use normalized URL joining
  - Add comprehensive tests covering all slash combinations


erickgalinkin (approved):
LGTM

PR #1351, opened 2025-08-25
chore: use ruff linter and formatter

Pouyanpi:
Migrate linting and formatting from `isort` and `black` to the unified `ruff` tool via pre-commit configuration.

### DIFF
--- a/nemoguardrails/imports.py
+++ b/nemoguardrails/imports.py
@@ -119,11 +119,13 @@ def import_optional_dependency(
 
     # Version checking logic can be added here if needed
     if min_version:
-        try:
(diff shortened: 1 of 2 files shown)


tgasser-nv:
We already have Black in our pre-commits, but this would add a linter which would be really valuable. Can we merge this after the functional and type-check MRs?

Pouyanpi:
```
 poetry run ruff check tests
All checks passed!
```
```
 poetry run ruff check nemoguardrails/
All checks passed!

```

Pouyanpi:
@greptileai

PR #1355, opened 2025-08-26
feat: Add Trend Micro Vision One AI Application Security community integration

trend-willem-gooderham:
## Description

Adds a new community integration, Trend Micro Vision One AI Application Security, to NeMo Guardrails.

## Checklist

### DIFF
--- a/examples/configs/trend_micro_v2/rails.co
+++ b/examples/configs/trend_micro_v2/rails.co
@@ -2,11 +2,7 @@ import guardrails
 import nemoguardrails.library.trend_micro
 
 flow input rails $input_text
-    $result = await TrendAiGuardAction(text=$input_text)
-
-    if $result.action == "Block"
-       send AiGuardException(message="AI Guard detection: " + $result.reason)
-       abort
+    trend ai guard input $input_text
 
 flow output rails $output_text
-    trend ai guard $output_text
+    trend ai guard output $output_text
--- a/nemoguardrails/library/trend_micro/actions.py
+++ b/nemoguardrails/library/trend_micro/actions.py
@@ -14,10 +14,12 @@
 # limitations under the License.
 
 import logging
-from typing import Optional
+from typing import Literal, Optional
 
 import httpx
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
+from pydantic import field_validator as validator
+from pydantic import model_validator
 from pydantic_core import to_json
 from typing_extensions import cast
 
@@ -28,15 +30,60 @@ log = logging.getLogger(__name__)
 
 
 class Guard(BaseModel):
+    """
+    Represents a guard entity with a single string attribute.
+
+    Attributes:
+        guard (str): The input text for guard analysis.
+    """
+
     guard: str
 
 
 class GuardResult(BaseModel):
-    action: str
-    reason: str
+    """
+    Represents the result of a guard analysis, specifying the action to take and the reason.
+
+    Attributes:
+        action (Literal["Block", "Allow"]): The action to take based on guard analysis.
+        Must be either "Block" or "Allow".
+        reason (str): Explanation for the chosen action. Must be a non-empty string.
+    """
+
+    action: Literal["Block", "Allow"] = Field(
+        ..., description="Action to take based on " "guard analysis"
+    )
+    reason: str = Field(..., min_length=1, description="Explanation for the action")
+    blocked: bool = Field(
+        default=False, description="True if action is 'Block', else False"
+    )
+
+    @validator("action")
+    def validate_action(cls, v):
+        log.error(f"Validating action: {v}")
+        if v not in ["Block", "Allow"]:
+            return "Allow"
+        return v
+
+    @model_validator(mode="before")
+    def set_blocked(cls, values):
+        a = values.get("action")
+        values["blocked"] = a.lower() == "block"
+        return values
 
 
 def get_config(config: RailsConfig) -> TrendMicroRailConfig:
+    """
+    Retrieves the TrendMicroRailConfig from the provided RailsConfig object.
+
+    Args:
+        config (RailsConfig): The Rails configuration object containing possible
+        Trend Micro settings.
+
+    Returns:
+        TrendMicroRailConfig: The Trend Micro configuration, either from the provided
+        config or a default instance.
+    """
     if (
         not hasattr(config.rails.config, "trend_micro")
         or config.rails.config.trend_micro is None
@@ -46,7 +93,12 @@ def get_config(config: RailsConfig) -> TrendMicroRailConfig:
     return cast(TrendMicroRailConfig, config.rails.config.trend_micro)
 
 
-@action(is_system_action=True)
+def trend_ai_guard_mapping(result: GuardResult) -> bool:
+    """Convert Trend Micro result to boolean for flow logic."""
+    return result.action.lower() == "block"
+
+
+@action(is_system_action=True, output_mapping=trend_ai_guard_mapping)
 async def trend_ai_guard(config: RailsConfig, text: Optional[str] = None):
     """
     Custom action to invoke the Trend Ai Guard
@@ -59,10 +111,11 @@ async def trend_ai_guard(config: RailsConfig, text: Optional[str] = None):
 
     v1_api_key = trend_config.get_api_key()
     if not v1_api_key:
-        raise ValueError("Trend Micro Vision One API Key not found")
-
-    if text is None:
-        raise ValueError("No prompt/response found in the last event.")
+        log.error("Trend Micro Vision One API Key not found")
+        return GuardResult(
+            action="Block",
+            reason="Trend Micro Vision One API Key not found",
+        )
 
     async with httpx.AsyncClient() as client:
         data = Guard(guard=text).model_dump()
@@ -80,10 +133,10 @@ async def trend_ai_guard(config: RailsConfig, text: Optional[str] = None):
             response.raise_for_status()
             guard_result = GuardResult(**response.json())
             log.debug("Trend Micro AI Guard Result: %s", guard_result)
-        except Exception as e:
+        except httpx.HTTPStatusError as e:
             log.error("Error calling Trend Micro AI Guard API: %s", e)
             return GuardResult(
-                action="allow",
+                action="Allow",
                 reason="An error occurred while calling the Trend Micro AI Guard API.",
             )
         return guard_result
--- a/nemoguardrails/library/trend_micro/flows.co
+++ b/nemoguardrails/library/trend_micro/flows.co
@@ -1,10 +1,22 @@
 # INPUT AND/OR OUTPUT RAIL
-flow trend ai guard $text
+flow trend ai guard input $text
     $result = await TrendAiGuardAction(text=$text)
 
-    if $result.action == "Block" # Fails open if AI Guard service has an error
+    if $result.blocked # Fails open if AI Guard service has an error
         if $system.config.enable_rails_exceptions
-            send TrendAiGuardException(message="Blocked by the 'trend ai guard' flow: " + $result.reason)
+            send TrendAiGuardRailException(message="Blocked by the 'trend ai guard input' flow: " + $result.reason)
         else
             bot refuse to respond
-            abort
+        abort
+
+
+# OUTPUT RAIL
+flow trend ai guard output $text
+    $result = await TrendAiGuardAction(text=$text)
+
+    if $result.blocked # Fails open if AI Guard service has an error
+        if $system.config.enable_rails_exceptions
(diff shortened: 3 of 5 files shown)


tests/test_trend_ai_guard.py
```diff
@@ -0,0 +1,99 @@
...
+@pytest.mark.parametrize("status_code", frozenset({429, 500, 502, 503, 504}))
+def test_trend_ai_guard_error(
+    httpx_mock: HTTPXMock, monkeypatch: pytest.MonkeyPatch, status_code: int
+):
+    monkeypatch.setenv("V1_API_KEY", "test-token")
+    httpx_mock.add_response(
+        is_reusable=True, status_code=status_code, json={"result": {}}
+    )
+
+    chat = TestChat(output_rail_config, llm_completions=["  Hello!"])  ## ??
```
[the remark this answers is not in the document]
trend-willem-gooderham: Resolved

nemoguardrails/library/trend_micro/actions.py
```diff
@@ -0,0 +1,72 @@
...
+            },
+        )
+
+        try:
+            response.raise_for_status()
+            guard_result = GuardResult(**response.json())
+            log.debug("Trend Micro AI Guard Result: %s", guard_result)
+        except Exception as e:
+            log.error("Error calling Trend Micro AI Guard API: %s", e)
+            return GuardResult(action="allow", reason=str(e))
```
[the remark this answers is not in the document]
trend-willem-gooderham: Resolved

docs/user-guides/community/trend-micro.md
```diff
@@ -0,0 +1,51 @@
+# Trend Micro Vision One AI Application Security
+
+Trend Micro Vision One [AI Application Security's](https://docs.trendmicro.com/en-us/documentation/article/trend-vision-one-ai-scanner-ai-guard) AI Guard feature uses a configurable policy to identify risks in AI Applications, such as:
+
+- Prompt injection attacks
+- Toxicity, violent, and other harmful content
+- Sensitive Data
```
erickgalinkin, line 7: I've read the documents linked but they're extremely sparse. Are there any more detailed docs available?

docs/user-guides/community/trend-micro.md
```diff
@@ -0,0 +1,51 @@
...
+
+
+The following environment variable is required to use the integration:
+
+- `V1_API_KEY`: A Vision One API Token with AI Guard Permissions
+
+You can optionally set:
+
+- `V1_URL`: The URL for which instances of AI Guard should be invoked
+  Defaults to `https://api.xdr.trendmicro.com/beta/aiSecurity/guard` for Vision One's hosted US SaaS deployment
```
erickgalinkin: We may want to expose these in the colang config ala the `jailbreak_detection` methods.
See [the documentation](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/docs/user-guides/advanced/nemoguard-jailbreakdetect-deployment.md) for reference. 

`V1_API_KEY` and `V1_URL` are acceptable defaults, but are sufficiently generic that they may end up with naming collisions. Letting these be configurable is ideal, especially since the default url is only for the US deployment and according to the [Integrate AI Guard docs](https://docs.trendmicro.com/en-us/documentation/article/trend-vision-one-integrate-ai-guard) there appear to be 7 supported regions.

docs/user-guides/community/trend-micro.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Setup
```
erickgalinkin, line 10: Would like to see something mentioned about how to actually set it up on the Trend Micro side. See the [Fiddler docs](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/docs/user-guides/community/fiddler.md) for a reference.

docs/user-guides/guardrails-library.md
```diff
@@ -887,6 +888,24 @@ rails:
 
 For more details, check out the [Pangea AI Guard Integration](./community/pangea.md) page.
 
+### Trend Micro Vision One AI Application Security
+
+NeMo Guardrails supports using Trend Micro Vision One AI Guard for protecting input and output flows within AI-powered applications.
```
erickgalinkin: It would be good to mention the way the environment variables or colang configs get set up here.

nemoguardrails/library/trend_micro/actions.py
```diff
@@ -0,0 +1,75 @@
...
+                "Authorization": f"Bearer {v1_api_key}",
+                "Content-Type": "application/json",
+            },
+        )
+
+        try:
+            response.raise_for_status()
+            guard_result = GuardResult(**response.json())
+            log.debug("Trend Micro AI Guard Result: %s", guard_result)
+        except Exception as e:
```
erickgalinkin: This is a super broad exception clause, which I don't love, though that's a bit of a nit.

According to the [AI Guard API reference](https://docs.trendmicro.com/en-us/documentation/article/trend-vision-one-ai-guard-api-reference), there are only 4 status codes other than 200 that should be raised. 

For the 429 error code, should we consider exponential backoff? Probably not? Doing so would introduce latency, which can be a problem for these applications already. Ultimately, I suppose this is fine, though I would prefer to see it specifically catch `httpx.HTTPStatusError` unless we expect other errors?

trend-willem-gooderham: Our main reasoning to go with a broad Exception clause was to ensure that we always fail open. If your preference would be to switch to a `httpx.HTTPStatusServer` I have no problem making this change.

No other errors are expected as a result of making an AI Guard API call.

We didn't consider exponential backoff due to the increase in latency.

docs/user-guides/community/trend-micro.md
trend-willem-gooderham: Documentation is still a work in progress but additional content will be posted to the existing links

nemoguardrails/library/trend_micro/actions.py
```diff
@@ -0,0 +1,89 @@
...
+from typing_extensions import cast
+
+from nemoguardrails.actions import action
+from nemoguardrails.rails.llm.config import RailsConfig, TrendMicroRailConfig
+
+log = logging.getLogger(__name__)
+
+
+class Guard(BaseModel):
+    guard: str
```
Pouyanpi: I suggest adding field descriptions and more specific type hints. `action` could be only `Block` and `allow`? is it case sensitive?

can `reason` be `None`?

```python
  from pydantic import Field, validator

  class GuardResult(BaseModel):
      action: Literal["Block", "allow"] = Field(..., description="Action to take based on guard
  analysis")
      reason: str = Field(..., min_length=1, description="Explanation for the action")

      @validator('action')
      def validate_action(cls, v):
          if v not in ['Block', 'allow']:
              raise ValueError(f"Invalid action: {v}")
          return v

```

nemoguardrails/library/trend_micro/actions.py
Pouyanpi: It feels a bit too Broad: The code fails open for ALL errors and I think there is a risk: failing open on all type of errors feel dangerous. Any reason why that request should not get blocked or an exception being raised?

nemoguardrails/library/trend_micro/actions.py
Pouyanpi: on a different note I suggest using 

```python
trend_ai_guard_mapping(result: GuardResult) -> bool:
      """Convert Trend Micro result to boolean for flow logic."""

# the logic
# e.g.
# return result.action.lower() == "block"
```

please see https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/nemoguardrails/actions/output_mapping.py

this must be implemented unless the default behavior is expected which is not the case here.

I also think having a boolean in colang flows is more appropriate, "Block" is False and "allow" is True.

Flow becomes cleaner:
```
  # Before (fragile)
  if $result.action == "Block"

  # After (robust)
  if $result  # now it's just a boolean
```


this is necessary for streaming with output rails. Please have a look at the other integration in library.

Pouyanpi: but if you need to have access to "reason" in colang flows then you don't necessarily need to only return booleans. You can see output mapping as a semantic layer: what output should be considered block and what should be allowed.

examples/configs/trend_micro_v2/rails.co
Pouyanpi, line 1: the flow is being redefined here. How about using the flow defined in flows.co file?

nemoguardrails/library/trend_micro/flows.co
Pouyanpi, line 1: for a consistent user experience I recommend to use both input and output flows for Colang 2.0.

please update the exception messages with the type of rails similar to what you've done for Colang 1.0.

nemoguardrails/library/trend_micro/flows.co
```diff
@@ -0,0 +1,10 @@
+# INPUT AND/OR OUTPUT RAIL
+flow trend ai guard $text
+    $result = await TrendAiGuardAction(text=$text)
+
+    if $result.action == "Block" # Fails open if AI Guard service has an error
+        if $system.config.enable_rails_exceptions
+            send TrendAiGuardException(message="Blocked by the 'trend ai guard' flow: " + $result.reason)
```
Pouyanpi: the convention is to use `RailException` please modify it to `TrendAiGuardRailException` to be consistent. Please modify it in both co files

nemoguardrails/library/trend_micro/flows.co
Pouyanpi: also please make sure that the Colang 2.0 example that you have works, I am getting:

```
raise ColangValueError(f"Error evaluating '{expr}', {e}")
nemoguardrails.colang.v2_x.runtime.errors.ColangValueError: Error evaluating 'not($result.action ==
"Block")', Attribute 'action' does not exist in expression 'not(var_result.action == "Block")'
```

you can quickly test it with 
```
poetry run nemoguardrails chat --config="./examples/configs/trend_micro_v2"
```

This is expected because I don't have API keys etc. 
Please have a closer look at that exception handling and make sure that different failure scenarios are handled properly 👍🏻

nemoguardrails/library/trend_micro/actions.py
Pouyanpi: for example to handle missing API key the same way as API errors:

```python
    if not v1_api_key:
        return GuardResult(
            action="allow",
            reason="Trend Micro Vision One API Key not found - failing open"
        )

```
Is not a good approach because a user without api key would bypass the guardrails:

```
Specified a value for Trend Micro config api_key_env var at V1_API_KEY but the environment variable was not set!
WARNING:nemoguardrails.rails.llm.config:Specified a value for Trend Micro config api_key_env var at V1_API_KEY but the environment variable was not set!
Hello! How can I assist you today?
```

also please see https://github.com/NVIDIA/NeMo-Guardrails/pull/1355#discussion_r2337095905

Pouyanpi: but returning action="Block" leads to

```
Event: {'type': 'AiGuardException', 'uid': 'a540e95b-ca01-485b-8d68-5fe2a3dd86be', 'event_created_at':
'2025-09-10T15:27:50.560088+00:00', 'source_uid': 'NeMoGuardrails-Colang-2.x', 'message': 'AI Guard
detection: Trend Micro Vision One API Key not found - failing open'}
```

nemoguardrails/library/trend_micro/flows.co
trend-willem-gooderham: Changed how exceptions are handled for the API use cases and re-checked the example configs for the v1 and v2 examples

nemoguardrails/library/trend_micro/actions.py
trend-willem-gooderham: Implemented an ouput_mapping and added a boolean blocked filed to the GuardResponse. Following other integrations and the comments from output_mapping the value of true is assigned a request should be blocked.

nemoguardrails/library/trend_micro/actions.py
trend-willem-gooderham: The caught exception has been changed along with handling when API keys aren't found.

Pouyanpi (approved):
Thank you @trend-willem-gooderham for making the changes 👍🏻 It looks good to me.

trend-willem-gooderham:
Hello @erickgalinkin, Could I get another review? Hoping to get this PR merged soon.

Thanks

tgasser-nv (approved):
Looks good, thanks @erickgalinkin  and @Pouyanpi for the discussion and iterations

trend-willem-gooderham:
Hello @tgasser-nv  and @Pouyanpi 

I don't have permission to merge, so please merge this PR when you can.
Thanks again for all of the reviews

PR #1364, opened 2025-08-29
feat(tool-calling): add tool call passthrough support in LLMRails

Pouyanpi:
This PR implements tool call extraction and passthrough functionality in LLMRails, enabling the framework to capture and forward tool calls from LLM responses. Key changes include:

  - Add tool_calls_var context variable for storing LLM tool calls
  - Refactor llm_call utils to extract and store tool calls from responses
  - Support tool calls in both GenerationResponse and dict message formats
  - Add ToolMessage support for langchain message conversion
  - Comprehensive test coverage for tool calling integration scenarios

The implementation maintains backward compatibility while adding new tool calling capabilities to the passthrough mode.

### DIFF
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -20,11 +20,15 @@ from langchain.base_language import BaseLanguageModel
 from langchain.callbacks.base import AsyncCallbackHandler, BaseCallbackManager
 from langchain.prompts.base import StringPromptValue
 from langchain.prompts.chat import ChatPromptValue
-from langchain.schema import AIMessage, HumanMessage, SystemMessage
+from langchain_core.messages import AIMessage, HumanMessage, SystemMessage, ToolMessage
 
 from nemoguardrails.colang.v2_x.lang.colang_ast import Flow
 from nemoguardrails.colang.v2_x.runtime.flows import InternalEvent, InternalEvents
-from nemoguardrails.context import llm_call_info_var, reasoning_trace_var
+from nemoguardrails.context import (
+    llm_call_info_var,
+    reasoning_trace_var,
+    tool_calls_var,
+)
 from nemoguardrails.logging.callbacks import logging_callbacks
 from nemoguardrails.logging.explain import LLMCallInfo
 
@@ -72,7 +76,22 @@ async def llm_call(
     custom_callback_handlers: Optional[List[AsyncCallbackHandler]] = None,
 ) -> str:
     """Calls the LLM with a prompt and returns the generated text."""
-    # We initialize a new LLM call if we don't have one already
+    _setup_llm_call_info(llm, model_name, model_provider)
+    all_callbacks = _prepare_callbacks(custom_callback_handlers)
+
+    if isinstance(prompt, str):
+        response = await _invoke_with_string_prompt(llm, prompt, all_callbacks, stop)
+    else:
+        response = await _invoke_with_message_list(llm, prompt, all_callbacks, stop)
+
+    _store_tool_calls(response)
+    return _extract_content(response)
+
+
+def _setup_llm_call_info(
+    llm: BaseLanguageModel, model_name: Optional[str], model_provider: Optional[str]
+) -> None:
+    """Initialize or update LLM call info in context."""
     llm_call_info = llm_call_info_var.get()
     if llm_call_info is None:
         llm_call_info = LLMCallInfo()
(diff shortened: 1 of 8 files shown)


nemoguardrails/actions/llm/utils.py
```diff
@@ -81,52 +100,84 @@ async def llm_call(
     llm_call_info.llm_model_name = model_name or _infer_model_name(llm)
     llm_call_info.llm_provider_name = model_provider
 
+
+def _prepare_callbacks(
+    custom_callback_handlers: Optional[List[AsyncCallbackHandler]],
+) -> BaseCallbackManager:
+    """Prepare callback manager with custom handlers if provided."""
     if custom_callback_handlers and custom_callback_handlers != [None]:
```
tgasser-nv, line 108: How would we get an argument of `[None]` rather than just `None`?

nemoguardrails/actions/llm/utils.py
```diff
@@ -81,52 +100,84 @@ async def llm_call(
...
+    """Invoke LLM with string prompt."""
+    try:
+        return await llm.ainvoke(prompt, config={"callbacks": callbacks, "stop": stop})
+    except Exception as e:
+        raise LLMCallException(e)
+
+
+async def _invoke_with_message_list(
+    llm: BaseLanguageModel,
+    prompt: List[dict],
```
tgasser-nv, line 131: nit: Should this be list[dict[str, str]]? I'm guilty of using the old-style capital-first-letter types (List vs list). I found out recently these were deprecated as of Python 3.9 in [PEP585](https://peps.python.org/pep-0585/).

nemoguardrails/actions/llm/utils.py
```diff
@@ -81,52 +100,84 @@ async def llm_call(
...
+    """Convert message list to LangChain message format."""
+    messages = []
+    for msg in prompt:
+        msg_type = msg["type"] if "type" in msg else msg["role"]
+
+        if msg_type == "user":
+            messages.append(HumanMessage(content=msg["content"]))
+        elif msg_type in ["bot", "assistant"]:
+            messages.append(AIMessage(content=msg["content"]))
+        elif msg_type == "system":
```
tgasser-nv, line 155: I think we should include `developer` in here as well to support o1 and newer models. Checking on the [OpenAI API spec](https://platform.openai.com/docs/api-reference/chat/create):
```
Developer message
object
Developer-provided instructions that the model should follow, regardless of messages sent by the user. With o1 models and newer, developer messages replace the previous system messages.
```

tests/rails/llm/test_options.py
```diff
@@ -110,3 +114,82 @@ def test_generation_options_serialization():
...
+    assert response.response == "Hello, world!"
+    assert response.tool_calls is None
+    assert response.llm_output is None
+    assert response.state is None
+
+
+def test_generation_response_with_tool_calls():
+    test_tool_calls = [
+        {
+            "name": "get_weather",
```
tgasser-nv, line 131: Are the tool-calling fields standardized on [OpenAI's example here?](https://platform.openai.com/docs/guides/function-calling#function-tool-example) Looks like they have slightly different names ("type": "function", "parameters" rather than "args", etc)

tests/rails/llm/test_options.py
```diff
@@ -110,3 +114,82 @@ def test_generation_options_serialization():
...
+
+def test_generation_response_empty_tool_calls():
+    """Test GenerationResponse with empty tool calls list."""
+    response = GenerationResponse(response="No tools needed", tool_calls=[])
+
+    assert response.tool_calls == []
+    assert len(response.tool_calls) == 0
+
+
+def test_generation_response_serialization_with_tool_calls():
```
tgasser-nv, line 163: nit: Can you add a one-line docstring to tests? It helps when debugging

tests/test_tool_calling_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    result = _convert_messages_to_langchain_format(messages)
+
+    assert len(result) == 1
+    assert isinstance(result[0], AIMessage)
+    assert result[0].content == "Hello from bot"
+
+
+def test_convert_messages_to_langchain_format_system():
+    """Test converting system messages to LangChain format."""
+    messages = [{"role": "system", "content": "You are a helpful assistant"}]
```
tgasser-nv, line 88: Can you add a test here with `developer` as well (see earlier comment)

tests/test_tool_calling_utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+    result = _convert_messages_to_langchain_format(messages)
+
+    assert len(result) == 1
+    assert isinstance(result[0], ToolMessage)
+    assert result[0].content == "Tool result"
+    assert result[0].tool_call_id == ""
+
+
+def test_convert_messages_to_langchain_format_mixed():
```
tgasser-nv, line 121: Same as above, please include a `role` of `developer`

tgasser-nv (approved):
Looks good, left some comments to address before merging. This is a huge PR (4k LOC) could you split this type of change into multiple smaller PRs in future?

Also could you add some local integration testing to check this works when interfacing with real Nemoguard / LLMs which support tool-calling?

tests/rails/llm/test_options.py
Pouyanpi: Yes we don't support function calling but we do

message of type tool

image

as in https://platform.openai.com/docs/api-reference/chat/create

tests/test_tool_calling_utils.py
Pouyanpi: I have a separate PR where I will add developer type 👍🏻

nemoguardrails/actions/llm/utils.py
Pouyanpi: will include it in a separate PR

nemoguardrails/actions/llm/utils.py
Pouyanpi: good point, I think in many places [callback_hander] is passed to llm_call and that can be None. 
It is not within the scope of this PR though.

for example 

```python
                        text = await llm_call(
                            llm,
                            prompt,
                            custom_callback_handlers=[streaming_handler_var.get()],
                        )
```

if streaming_handler_var.get is None

PR #1367, opened 2025-09-02
chore(types): Top-level PR for Guardrails type-cleaning

tgasser-nv:
## Checklist


tgasser-nv:
Note to self: Need to replace `# pyright: ignore` with `# type: ignore` which is a PEP standard and generalizes to other type-checkers other than Pyright

PR #1374, opened 2025-09-04
Ensure providers can be switched on/off at build time

m-misiura:


m-misiura:
Accidentally created, please ignore

PR #1370, opened 2025-09-03
feat(runnable-rails): stream metadata in RunnableRails output

Pouyanpi:
Enhance streaming in RunnableRails to include generation metadata in streamed chunks. Skips END_OF_STREAM markers and updates chunk formatting to support metadata for AIMessageChunk outputs. This improves compatibility with consumers expecting metadata in streaming responses.

PR #1366, opened 2025-09-02
feat(runnable-rails): update RunnableRails with full LangChain Runnable protocol support

Pouyanpi:
## Description
  - Complete rewrite of RunnableRails implementing full LangChain Runnable protocol
  - Added comprehensive async/sync support for invoke, batch, and streaming operations
  - Enhanced input/output handling for all LangChain message formats and chaining scenarios
  - Extensive test suite with 14 test files covering batching, streaming, tool calling, and composition

requires #1364 and #1343 and #1289

PR #1382, opened 2025-09-09
feat(tool-rails): add support for tool output rails and validation

Pouyanpi:
Introduce tool output/input rails configuration and Colang flows for tool call validation and parameter security checks. Add support for BotToolCall event emission in passthrough mode, enabling tool call guardrails before execution.

must be merged after #1370 #1369 #1366

### DIFF
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -582,7 +582,21 @@ class LLMGenerationActions:
             if streaming_handler:
                 await streaming_handler.push_chunk(text)
 
-            output_events.append(new_event_dict("BotMessage", text=text))
+            if self.config.passthrough:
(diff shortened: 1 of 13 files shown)


nemoguardrails/actions/llm/generation.py
```diff
@@ -582,7 +582,21 @@ async def generate_user_intent(
...
+            if self.config.passthrough:
+                from nemoguardrails.actions.llm.utils import (
+                    get_and_clear_tool_calls_contextvar,
+                )
+
+                tool_calls = get_and_clear_tool_calls_contextvar()
+
+                if tool_calls:
+                    output_events.append(
+                        new_event_dict("BotToolCall", tool_calls=tool_calls)
```
tgasser-nv, line 594: Should this be an enum?

PR #1387, opened 2025-09-10
Feat/llm params

Pouyanpi:
## Checklist


tgasser-nv (approved):
Looks good, 4k LOC is too large for a single PR. 

I'm a little confused about a few things:

* Why did we use a context-manager to pass a dict of LLM parameters in the first place? Normally they're used to make sure we close files/DB connections so we don't forget. 
* Does a context-manager break some Langchain functionality?
* Can you add some local integration tests to make sure this works calling tools with production LLMs?

nemoguardrails/actions/llm/utils.py
```diff
@@ -73,11 +73,28 @@ async def llm_call(
...
+    """Calls the LLM with a prompt and returns the generated text.
+
+    Args:
+        llm: The language model instance to use
+        prompt: The prompt string or list of messages
+        model_name: Optional model name for tracking
+        model_provider: Optional model provider for tracking
+        stop: Optional list of stop tokens
+        custom_callback_handlers: Optional list of callback handlers
+        llm_params: Optional configuration dictionary to pass to the LLM (e.g., temperature, max_tokens)
```
trebedea:
```suggestion
        llm_params: Optional configuration dictionary to pass to the LLM (e.g., temperature, max_tokens) for the current generation / invocation
```

nemoguardrails/actions/v2_x/generation.py
```diff
@@ -450,19 +458,19 @@ async def passthrough_llm_action(
 
         generation_options: GenerationOptions = generation_options_var.get()
 
-        with llm_params(
+        generation_llm_params = (
+            generation_options and generation_options.llm_params
+        ) or {}
```
trebedea: Should we use `None` instead of `{}` as a default in all these cases?

PR #1388, opened 2025-09-10
chore(types): Type-clean tracing (10 errors)

tgasser-nv:
## Checklist


nemoguardrails/tracing/tracer.py
```diff
@@ -66,6 +66,10 @@ def generate_interaction_log(
         if generation_log is None:
             generation_log = self._generation_log
 
+        # At this point generation_log should not be None since it comes from self._generation_log
+        if generation_log is None:
```
trebedea, line 70: There is already an `RuntimeError` raised in `__init__ ` (L50).
Combined with L66, `generation_log` cannot be `None` here. Just of curiosity, this change is from a coding agent? 

```
if self._generation_log is None:
            raise RuntimeError("Generation log is missing.")
```

nemoguardrails/tracing/span_extractors.py
```diff
@@ -219,17 +219,11 @@ def extract_spans(
...
-                        type(action.error).__name__
-                        if hasattr(action, "error") and action.error
-                        else None
-                    ),
-                    error_message=(
-                        str(action.error)
-                        if hasattr(action, "error") and action.error
-                        else None
-                    ),
+                    # TODO: There is no error field in ExecutedAction. The fields below are defined on BaseSpan but
```
trebedea, line 222: Not sure here - maybe @Pouyanpi can provide some feedback quicker than me.

nemoguardrails/tracing/tracer.py
tgasser-nv: As far as the type-checker knows, `generation_log` can still be `None` at this point. On line 67 `self.generation_log` could be `None` in the assignment. The type of `self._generation_log` is set implicitly on line 48, and comes from the  `response` argument of type `GenerationResponse`. This field is defined as `log: Optional[GenerationLog]`. So without an explicit check it could be None


```python
class GenerationResponse(BaseModel):
    # TODO: add typing for the list of messages
    response: Union[str, List[dict]] = Field(
        description="The list of the generated messages."
    )
.....
```

nemoguardrails/tracing/interaction_types.py
```diff
@@ -78,6 +78,6 @@ def extract_interaction_log(
     return InteractionLog(
         id=interaction_output.id,
         activated_rails=generation_log.activated_rails,
-        events=generation_log.internal_events,
+        events=generation_log.internal_events or [],
```
Pouyanpi, line 81: I'm wondering why Pyright is complaining about this as events is default factory of List

nemoguardrails/tracing/tracer.py
trebedea: I understood the principle, and it's fine for me - but seems strange to have code which is never executed. Maybe in situations like this we should be explicit that `self._generation_log` cannot be `None`. This helps both developers and checkers.

nemoguardrails/tracing/span_extractors.py
tgasser-nv: @Pouyanpi could you check this?

PR #1396, opened 2025-09-15
chore(types): Type-clean rails (82 errors)

tgasser-nv:
## Checklist

### DIFF
--- a/nemoguardrails/context.py
+++ b/nemoguardrails/context.py
@@ -14,25 +14,42 @@
 # limitations under the License.
 
 import contextvars
-from typing import Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
-streaming_handler_var = contextvars.ContextVar("streaming_handler", default=None)
+if TYPE_CHECKING:
+    from nemoguardrails.logging.explain import ExplainInfo
+    from nemoguardrails.rails.llm.options import GenerationOptions, LLMStats
+    from nemoguardrails.streaming import StreamingHandler
+
+streaming_handler_var: contextvars.ContextVar[
+    Optional["StreamingHandler"]
+] = contextvars.ContextVar("streaming_handler", default=None)
 
 # The object that holds additional explanation information.
-explain_info_var = contextvars.ContextVar("explain_info", default=None)
+explain_info_var: contextvars.ContextVar[
+    Optional["ExplainInfo"]
+] = contextvars.ContextVar("explain_info", default=None)
 
 # The current LLM call.
-llm_call_info_var = contextvars.ContextVar("llm_call_info", default=None)
+llm_call_info_var: contextvars.ContextVar[Optional[str]] = contextvars.ContextVar(
+    "llm_call_info", default=None
+)
 
 # All the generation options applicable to the current context.
-generation_options_var = contextvars.ContextVar("generation_options", default=None)
+generation_options_var: contextvars.ContextVar[
+    Optional["GenerationOptions"]
+] = contextvars.ContextVar("generation_options", default=None)
 
 # The stats about the LLM calls.
-llm_stats_var = contextvars.ContextVar("llm_stats", default=None)
+llm_stats_var: contextvars.ContextVar[Optional["LLMStats"]] = contextvars.ContextVar(
+    "llm_stats", default=None
+)
 
 # The raw LLM request that comes from the user.
 # This is used in passthrough mode.
-raw_llm_request = contextvars.ContextVar("raw_llm_request", default=None)
+raw_llm_request: contextvars.ContextVar[Optional[Any]] = contextvars.ContextVar(
+    "raw_llm_request", default=None
+)
 
 reasoning_trace_var: contextvars.ContextVar[Optional[str]] = contextvars.ContextVar(
     "reasoning_trace", default=None
--- a/nemoguardrails/rails/llm/buffer.py
+++ b/nemoguardrails/rails/llm/buffer.py
@@ -14,7 +14,10 @@
 # limitations under the License.
 
 from abc import ABC, abstractmethod
-from typing import AsyncGenerator, List, NamedTuple
+from typing import TYPE_CHECKING, AsyncGenerator, List, NamedTuple
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncIterator
 
 from nemoguardrails.rails.llm.config import OutputRailsStreamingConfig
 
@@ -111,9 +114,7 @@ class BufferStrategy(ABC):
         ...
 
     @abstractmethod
-    async def process_stream(
-        self, streaming_handler
-    ) -> AsyncGenerator[ChunkBatch, None]:
+    async def process_stream(self, streaming_handler):
         """Process streaming chunks and yield chunk batches.
 
         This is the main method that concrete buffer strategies must implement.
@@ -138,9 +139,9 @@ class BufferStrategy(ABC):
             ...     print(f"Processing: {context_formatted}")
             ...     print(f"User: {user_formatted}")
         """
-        ...
+        yield ChunkBatch([], [])  # pragma: no cover
 
-    async def __call__(self, streaming_handler) -> AsyncGenerator[ChunkBatch, None]:
+    async def __call__(self, streaming_handler):
         """Callable interface that delegates to process_stream.
 
         It delegates to the `process_stream` method and can
@@ -256,9 +257,7 @@ class RollingBuffer(BufferStrategy):
             buffer_context_size=config.context_size, buffer_chunk_size=config.chunk_size
         )
 
-    async def process_stream(
-        self, streaming_handler
-    ) -> AsyncGenerator[ChunkBatch, None]:
+    async def process_stream(self, streaming_handler):
         """Process streaming chunks using rolling buffer strategy.
 
         This method implements the rolling buffer logic, accumulating chunks
--- a/nemoguardrails/rails/llm/config.py
+++ b/nemoguardrails/rails/llm/config.py
@@ -1048,7 +1048,9 @@ def _load_path(
 
     # the first .railsignore file found from cwd down to its subdirectories
     railsignore_path = utils.get_railsignore_path(config_path)
-    ignore_patterns = utils.get_railsignore_patterns(railsignore_path)
+    ignore_patterns = (
+        utils.get_railsignore_patterns(railsignore_path) if railsignore_path else set()
+    )
 
     if os.path.isdir(config_path):
         for root, _, files in os.walk(config_path, followlinks=True):
@@ -1165,8 +1167,8 @@ def _parse_colang_files_recursively(
         current_file, current_path = colang_files[len(parsed_colang_files)]
 
         with open(current_path, "r", encoding="utf-8") as f:
+            content = f.read()
             try:
-                content = f.read()
                 _parsed_config = parse_colang_file(
                     current_file, content=content, version=colang_version
                 )
@@ -1668,7 +1670,7 @@ class RailsConfig(BaseModel):
             # if we have output rails streaming enabled
             # we keep it in case it was needed when we have
             # support per rails
-            if self.rails.output.streaming.enabled:
+            if self.rails.output.streaming and self.rails.output.streaming.enabled:
                 return True
             return False
 
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -24,7 +24,18 @@ import re
 import threading
 import time
 from functools import partial
-from typing import Any, AsyncIterator, Dict, List, Optional, Tuple, Type, Union, cast
+from typing import (
+    Any,
+    AsyncIterator,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+    cast,
+)
 
 from langchain_core.language_models import BaseChatModel
 from langchain_core.language_models.llms import BaseLLM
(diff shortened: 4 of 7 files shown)


nemoguardrails/rails/llm/options.py
```diff
@@ -355,12 +355,12 @@ def print_summary(self):
...
             )
 
-        processing_overhead = self.stats.total_duration - duration
+        processing_overhead = (self.stats.total_duration or 0) - duration
         if processing_overhead >= 0.01:
             _pc = round(100 - pc, 2)
             print(f"  - [{processing_overhead:.2f}s][{_pc}%]: Processing overhead ")
 
-        if self.stats.llm_calls_count > 0:
+        if self.stats.llm_calls_count and self.stats.llm_calls_count > 0:
```
Pouyanpi:
```suggestion
        if self.stats.llm_calls_count:
```

it is not required for fixing the type errors. Better to be explicit about the LLM calls, otherwise should we consider this a breaking change?

nemoguardrails/rails/llm/options.py
Pouyanpi, line 1: ✅ 

I am not sure why `GenerationStats` fields are of Optional[float] type sometimes with default value of None and sometimes 0.

if we change it to float and default of 0

Then we don't need need `or 0`
```diff
-processing_overhead = (self.stats.total_duration or 0) - duration
+processing_overhead = self.stats.total_duration - duration
```

so I think it is better to use

```python
class GenerationStats(BaseModel):
    """General stats about the generation."""

    input_rails_duration: float = Field(
        default=0,
        description="The time in seconds spent in processing the input rails.",
    )
    dialog_rails_duration: float = Field(
        default=0,
        description="The time in seconds spent in processing the dialog rails.",
    )
    generation_rails_duration: float = Field(
        default=0,
        description="The time in seconds spent in generation rails.",
    )
    output_rails_duration: float = Field(
        default=0,
        description="The time in seconds spent in processing the output rails.",
    )
    total_duration: float = Field(default=0, description="The total time in seconds.")
    llm_calls_duration: float = Field(
        default=0, description="The time in seconds spent in LLM calls."
    )
    llm_calls_count: int = Field(
        default=0, description="The number of LLM calls in total."
    )
    llm_calls_total_prompt_tokens: int = Field(
        default=0, description="The total number of prompt tokens."
    )
    llm_calls_total_completion_tokens: int = Field(
        default=0, description="The total number of completion tokens."
    )
    llm_calls_total_tokens: int = Field(
        default=0, description="The total number of tokens."
    )
```

nemoguardrails/context.py
```diff
@@ -14,25 +14,42 @@
 # limitations under the License.
 
 import contextvars
-from typing import Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
-streaming_handler_var = contextvars.ContextVar("streaming_handler", default=None)
+if TYPE_CHECKING:
+    from nemoguardrails.logging.explain import ExplainInfo
+    from nemoguardrails.rails.llm.options import GenerationOptions, LLMStats
```
Pouyanpi: `LLMStats` is an unknown import symbol (does not exist)

Pouyanpi: should be

```python
nemoguardrails.logging.stats
```

nemoguardrails/context.py
```diff
@@ -14,25 +14,42 @@
...
 
 # The object that holds additional explanation information.
-explain_info_var = contextvars.ContextVar("explain_info", default=None)
+explain_info_var: contextvars.ContextVar[
+    Optional["ExplainInfo"]
+] = contextvars.ContextVar("explain_info", default=None)
 
 # The current LLM call.
-llm_call_info_var = contextvars.ContextVar("llm_call_info", default=None)
+llm_call_info_var: contextvars.ContextVar[Optional[str]] = contextvars.ContextVar(
```
Pouyanpi: to be consistent:

fix `llm_call_info_var` type to use `LLMCallInfo` from `nemoguardrails.logging.explain`

nemoguardrails/context.py
```diff
@@ -14,25 +14,42 @@
...
 # The stats about the LLM calls.
-llm_stats_var = contextvars.ContextVar("llm_stats", default=None)
+llm_stats_var: contextvars.ContextVar[Optional["LLMStats"]] = contextvars.ContextVar(
+    "llm_stats", default=None
+)
 
 # The raw LLM request that comes from the user.
 # This is used in passthrough mode.
-raw_llm_request = contextvars.ContextVar("raw_llm_request", default=None)
+raw_llm_request: contextvars.ContextVar[Optional[Any]] = contextvars.ContextVar(
```
Pouyanpi: maybe

` Optional[Union[str, List[Dict[str, Any]]]] `

tests/test_llm_options.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import Any, Dict, List, Optional
+from unittest.mock import Mock
+
+import pytest
+from langchain_core.language_models import BaseChatModel
+from langchain_core.messages import BaseMessage
+from langchain_core.outputs import ChatGeneration, ChatResult
+from pydantic import BaseModel, Field
+
+from nemoguardrails.rails.llm.config import RailsConfig
```
Pouyanpi: these are all unused imports.

Please move the test to `test_generation_options.py`

nemoguardrails/rails/llm/config.py
```diff
@@ -1668,7 +1670,7 @@ def streaming_supported(self):
             # if we have output rails streaming enabled
             # we keep it in case it was needed when we have
             # support per rails
-            if self.rails.output.streaming.enabled:
+            if self.rails.output.streaming and self.rails.output.streaming.enabled:
```
Pouyanpi, line 1753: nit: maybe better to make 

```python
    streaming: OutputRailsStreamingConfig = Field(
        default_factory=OutputRailsStreamingConfig,
        description="Configuration for streaming output rails.",
    )
```
in `OutputRails`

nemoguardrails/rails/llm/buffer.py
```diff
@@ -111,9 +114,7 @@ def format_chunks(self, chunks: List[str]) -> str:
         ...
 
     @abstractmethod
-    async def process_stream(
-        self, streaming_handler
-    ) -> AsyncGenerator[ChunkBatch, None]:
+    async def process_stream(self, streaming_handler):
```
Pouyanpi: I think it is better to revert all the changes made here and use

```python
@abstractmethod
    async def process_stream(
        self, streaming_handler
    ) -> AsyncGenerator[ChunkBatch, None]:
        """Process streaming chunks and yield chunk batches.

        This is the main method that concrete buffer strategies must implement.
        It defines how chunks from the streaming handler should be buffered,
        processed, and yielded as ChunkBatch objects.

        Args:
            streaming_handler: An async iterator that yields individual string
                chunks from the LLM stream.

        Yields:
            ChunkBatch: Named tuple containing processing_context and user_output_chunks.


        Example:
            >>> strategy = SomeBufferStrategy()
            >>> async for chunk_batch in strategy.process_stream(handler):
            ...     # for output rails processing (needs context):
            ...     context_formatted = strategy.format_chunks(chunk_batch.processing_context)
            ...     # for user output (new content only):
            ...     user_formatted = strategy.format_chunks(chunk_batch.user_output_chunks)
            ...     print(f"Processing: {context_formatted}")
            ...     print(f"User: {user_formatted}")
        """
        raise NotImplementedError
        yield

```

the yield is never reached but helps pyright to infer type.
we can also add # type: ignore  to the yield but should not cause any issue,

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -889,16 +910,25 @@ async def generate_async(
...
         # we move that to the `$bot_message` variable. This is to enable a more
         # convenient interface. (only when dialog rails are disabled)
         if (
-            messages[-1]["role"] == "assistant"
+            messages
+            and messages[-1]["role"] == "assistant"
             and options
-            and options.rails.dialog is False
+            and hasattr(options, "rails")
+            and getattr(getattr(options, "rails", None), "dialog", None) is False
```
trebedea:
```suggestion
            and getattr(getattr(options, "rails"), "dialog", None) is False
```

trebedea: We already have the check for `hasattr(options, "rails")` in the conjunction.

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -915,7 +945,7 @@ async def generate_async(
         processing_log = []
 
         # The array of events corresponding to the provided sequence of messages.
-        events = self._get_events_for_messages(messages, state)
+        events = self._get_events_for_messages(messages or [], state)
```
trebedea: It's better to not have `messages` as `Optional` in the function signature?

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1094,82 +1124,90 @@ async def generate_async(
...
+                        res.output_data = {k: context.get(k) for k in output_vars}
                     else:
                         # Otherwise, we return the full context
                         res.output_data = context
 
                 _log = compute_generation_log(processing_log)
 
                 # Include information about activated rails and LLM calls if requested
-                if options.log.activated_rails or options.log.llm_calls:
+                log_options = getattr(options, "log", None)
```
trebedea: `GenerationLogOptions` cannot be `None` as per its definition in `GenerationOptions`

```
    log: GenerationLogOptions = Field(
        default_factory=GenerationLogOptions,
        description="Options about what to include in the log. By default, nothing is included. ",
    )
```

The same is valid for most of its attributes: they are not `Optional`, not sure why `getattr` provides a default value in many places in this function.

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1094,82 +1124,90 @@ async def generate_async(
...
+                if getattr(options, "llm_output", False):
                     # Currently, we include the output from the generation LLM calls.
                     for activated_rail in _log.activated_rails:
                         if activated_rail.type == "generation":
                             for executed_action in activated_rail.executed_actions:
                                 for llm_call in executed_action.llm_calls:
                                     res.llm_output = llm_call.raw_response
             else:
-                if options.output_vars:
+                if getattr(options, "output_vars", None):
```
Pouyanpi: prefer options.output_vars in case of Pydantic models, also easier for refactoring.

nemoguardrails/rails/llm/llmrails.py
tgasser-nv: It's likely because `options` is of type `Optional[Union[dict, GenerationOptions]]` and if it's a dict we don't know which keys will be present. So the code has to be very defensive and cope with either.

The best way to deal with this is to cast to GenerationOptions at the start, but there are so many paths through the code it's hard to tell if there will be unintended consequences.

These `Optional[Union[dict, ActualType]]` fields are scattered through the code and make type-checking much more complex.

nemoguardrails/rails/llm/llmrails.py
tgasser-nv: This code has to work for either GenerationOptions or a dict without a code block to convert dicts to GenerationOptions at the start of the function. There are so many paths through this 500+ line function and we also mess with the generation options on lines 996-1018. I'd rather leave this and come back to clean up when refactoring when we can make stronger assumptions about this

nemoguardrails/rails/llm/options.py
tgasser-nv: The advantage is that None represents no data recorded (i.e. that type of rails didn't actually run). While 0 would show a rail executed and completed in 0 seconds.

nemoguardrails/rails/llm/options.py
tgasser-nv: Yes I agree, we should report these values even if we had no LLM calls for consistency

tests/test_llm_options.py
tgasser-nv: Will do!

nemoguardrails/context.py
tgasser-nv: Yep added this in latest commit

nemoguardrails/context.py
tgasser-nv: Done!

nemoguardrails/context.py
tgasser-nv: Updated

tests/test_llm_options.py
tgasser-nv: Moved the test and deleted this file

nemoguardrails/rails/llm/config.py
tgasser-nv: Agree, made this change

nemoguardrails/rails/llm/buffer.py
tgasser-nv: Done!

nemoguardrails/rails/llm/llmrails.py
tgasser-nv: Done!

nemoguardrails/rails/llm/buffer.py
Pouyanpi: It isn't changed yet, would you please restore the previous version and then just make that minor change?

tests/test_llm_options.py
Pouyanpi: thanks, just make sure this file is deleted

nemoguardrails/rails/llm/options.py
Pouyanpi: I understand your point but I think 0 seconds duration could mean no execution at all. But I'm OK with both approach 👍🏻

Pouyanpi (approved):
Thank you for adding it to pre-commits, look good to merge 👍🏻

PR #1405, opened 2025-09-19
fix(runnable-rails): preserve message metadata in RunnableRails tool calling

Pouyanpi:
## Summary
- Fixes tool calling failures when using RunnableRails with passthrough=True
- Refactors message conversion logic into reusable utilities
- Ensures all message metadata is preserved during LangChain format conversions

## Changes
- Created `message_utils.py` with centralized message conversion functions
- Updated RunnableRails to use message_utils instead of hardcoded conversions
- Modified utils.py to dynamically preserve all message fields

### DIFF
--- a/tests/runnable_rails/test_transform_input.py
+++ b/tests/runnable_rails/test_transform_input.py
@@ -63,9 +63,26 @@ def test_transform_chat_prompt_value(rails):
 
     result = rails._transform_input_to_rails_format(chat_prompt)
     expected = [
-        {"role": "system", "content": "You are helpful"},
-        {"role": "user", "content": "Hello"},
-        {"role": "assistant", "content": "Hi there"},
+        {
+            "role": "system",
+            "content": "You are helpful",
+            "additional_kwargs": {},
+            "response_metadata": {},
+        },
+        {
+            "role": "user",
+            "content": "Hello",
+            "additional_kwargs": {},
+            "response_metadata": {},
+        },
+        {
+            "role": "assistant",
+            "content": "Hi there",
+            "additional_kwargs": {},
+            "response_metadata": {},
+            "tool_calls": [],
+            "invalid_tool_calls": [],
+        },
     ]
     assert result == expected
 
@@ -139,8 +156,20 @@ def test_transform_list_of_base_messages(rails):
 
     result = rails._transform_input_to_rails_format(messages)
     expected = [
-        {"role": "user", "content": "What is Python?"},
-        {"role": "assistant", "content": "Python is a programming language"},
+        {
+            "role": "user",
+            "content": "What is Python?",
+            "additional_kwargs": {},
+            "response_metadata": {},
+        },
+        {
+            "role": "assistant",
+            "content": "Python is a programming language",
+            "additional_kwargs": {},
+            "response_metadata": {},
+            "tool_calls": [],
+            "invalid_tool_calls": [],
+        },
     ]
     assert result == expected
 
@@ -150,7 +179,14 @@ def test_transform_single_human_message(rails):
     message = HumanMessage(content="Hello there")
 
     result = rails._transform_input_to_rails_format(message)
-    expected = [{"role": "user", "content": "Hello there"}]
+    expected = [
+        {
+            "role": "user",
+            "content": "Hello there",
+            "additional_kwargs": {},
+            "response_metadata": {},
+        }
+    ]
     assert result == expected
 
 
@@ -159,7 +195,16 @@ def test_transform_single_ai_message(rails):
     message = AIMessage(content="Hello back")
 
     result = rails._transform_input_to_rails_format(message)
-    expected = [{"role": "assistant", "content": "Hello back"}]
+    expected = [
+        {
+            "role": "assistant",
+            "content": "Hello back",
+            "additional_kwargs": {},
+            "response_metadata": {},
+            "tool_calls": [],
+            "invalid_tool_calls": [],
+        }
+    ]
     assert result == expected
 
 


nemoguardrails/integrations/langchain/message_utils.py
```diff
@@ -0,0 +1,292 @@
...
+)
+
+
+def get_message_role(msg: BaseMessage) -> str:
+    """Get the role string for a BaseMessage."""
+    if isinstance(msg, AIMessage):
+        return "assistant"
+    elif isinstance(msg, HumanMessage):
+        return "user"
+    elif isinstance(msg, SystemMessage):
```
tgasser-nv, line 36: Recommend including `developer` as well as `system`. Looks like this used from o1 onwards (https://platform.openai.com/docs/api-reference/chat/create)

nemoguardrails/integrations/langchain/message_utils.py
```diff
@@ -0,0 +1,292 @@
...
+        return getattr(msg, "type", "user")
+
+
+def get_message_class(msg_type: str) -> Type[BaseMessage]:
+    """Get the appropriate message class for a given type/role."""
+    if msg_type == "user":
+        return HumanMessage
+    elif msg_type in ["bot", "assistant"]:
+        return AIMessage
+    elif msg_type == "system":
```
tgasser-nv: Same as previous comment, looks like `developer` is used for OpenAI models after o1 rather than system

nemoguardrails/integrations/langchain/message_utils.py
```diff
@@ -0,0 +1,292 @@
...
+    if isinstance(msg, ToolMessage):
+        result["tool_call_id"] = msg.tool_call_id
+
+    exclude_fields = {"type", "content", "example"}
+
+    if hasattr(msg, "model_fields"):
+        for field_name in msg.model_fields:
+            if field_name not in exclude_fields and field_name not in result:
+                value = getattr(msg, field_name, None)
+                if value is not None and (not isinstance(value, (list, dict)) or value):
```
tgasser-nv: nit: Should we include falsey variables too (`if value` rather than `if value is not None`)? Then if a string value is "" or list [] then we won't create a dict key with the empty value?

nemoguardrails/integrations/langchain/runnable_rails.py
```diff
@@ -129,7 +121,7 @@ async def passthrough_fn(context: dict, events: List[dict]):
             # If the output is a string, we consider it to be the output text
             if isinstance(_output, str):
                 text = _output
-            elif isinstance(_output, BaseMessage):
+            elif message_utils.is_base_message(_output):
```
tgasser-nv: nit: Could you import each of these functions (i.e. `from message utils import is_base_message, ...` ) rather than having the `message_utils.x()` in each line?

tests/runnable_rails/test_message_utils.py
```diff
@@ -0,0 +1,419 @@
...
+    def test_get_message_class_unknown(self):
+        with pytest.raises(ValueError, match="Unknown message type"):
+            get_message_class("unknown")
+
+
+class TestMessageConversion:
+    def test_ai_message_with_tool_calls(self):
+        original = AIMessage(
+            content="",
+            tool_calls=[
```
tgasser-nv, line 93: For completeness, could you add `[invalid_tool_calls](https://python.langchain.com/api_reference/core/messages/langchain_core.messages.ai.AIMessage.html#langchain_core.messages.ai.AIMessage.invalid_tool_calls)` too and tests to cover this?

tests/runnable_rails/test_message_utils.py
```diff
@@ -0,0 +1,419 @@
...
+    def test_human_message_basic(self):
+        original = HumanMessage(content="Hello", id="human-1")
+        msg_dict = message_to_dict(original)
+        recreated = dict_to_message(msg_dict)
+
+        assert isinstance(recreated, HumanMessage)
+        assert recreated.content == original.content
+        assert recreated.id == original.id
+
+    def test_system_message_basic(self):
```
tgasser-nv, line 177: As mentioned above, need a test for `developer` which seems to be thenew replacement for `system` role post-o1 models

tests/runnable_rails/test_message_utils.py
```diff
@@ -0,0 +1,419 @@
...
+            assert type(orig) is type(recr)
+            assert orig.content == recr.content
+            if hasattr(orig, "id") and orig.id:
+                assert orig.id == recr.id
+            if hasattr(orig, "name") and orig.name:
+                assert orig.name == recr.name
+
+
+class TestTypeChecking:
+    def test_is_message_type(self):
```
tgasser-nv, line 277: nit: Should this be two tests, one for AIMessage and another for HumanMessage?

tests/runnable_rails/test_message_utils.py
```diff
@@ -0,0 +1,419 @@
...
+        assert msg.content == "tool result"
+        assert msg.tool_call_id == "call-123"
+        assert msg.name == "calculator"
+        assert msg.additional_kwargs == {"result": "success"}
+        assert msg.response_metadata == {"tool": "calc"}
+        assert msg.id == "tool-1"
+        assert msg.status == "success"
+
+
+class TestEdgeCases:
```
tgasser-nv, line 419: Could you add some test-cases for falsey-type values (empty lists, `""` strings)? What is the expected behaviour for these?

tgasser-nv (approved):
Looks good. The only issue I can see is that Langchain doesn't support OpenAI's new [`developer` role](https://platform.openai.com/docs/api-reference/chat/create). We can handle this by treating `developer` turns the same as `system` ones (that they're replacing).

Some other nits to address before merging.

nemoguardrails/integrations/langchain/message_utils.py
Pouyanpi: Made the change 👍🏻 thanks for mentioning this.

nemoguardrails/integrations/langchain/message_utils.py
Pouyanpi: YES, thanks for catching this 👍🏻

PR #1412, opened 2025-09-24
fix(logging): add "Tool" type to message sender labeling

Pouyanpi:
## Description

Previously, messages of type "tool" were not distinctly labeled in the LoggingCallbackHandler output, causing them to be grouped under "System".
This change adds explicit handling for "tool" messages, labeling them as "Tool" in the logs for improved clarity and debugging.

### DIFF
--- a/nemoguardrails/logging/callbacks.py
+++ b/nemoguardrails/logging/callbacks.py
@@ -109,6 +109,8 @@ class LoggingCallbackHandler(AsyncCallbackHandler, StdOutCallbackHandler):
                     if msg.type == "human"
                     else "Bot"
                     if msg.type == "ai"
+                    else "Tool"
+                    if msg.type == "tool"
                     else "System"
                 )
                 + "[/]"
--- a/tests/test_callbacks.py
+++ b/tests/test_callbacks.py
@@ -13,11 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from unittest.mock import patch
 from uuid import uuid4
 
 import pytest
 from langchain.schema import Generation, LLMResult
-from langchain_core.messages import AIMessage
+from langchain_core.messages import AIMessage, HumanMessage, SystemMessage, ToolMessage
 from langchain_core.outputs import ChatGeneration
 
 from nemoguardrails.context import explain_info_var, llm_call_info_var, llm_stats_var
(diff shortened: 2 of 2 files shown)


tests/test_callbacks.py
```diff
@@ -168,3 +169,46 @@ async def test_multiple_generations_token_accumulation():
     assert llm_stats.get_stat("total_tokens") == 19
     assert llm_stats.get_stat("total_prompt_tokens") == 12
     assert llm_stats.get_stat("total_completion_tokens") == 7
+
+
+@pytest.mark.asyncio
+async def test_tool_message_labeling_in_logging():
```
tgasser-nv, line 175: Would it be simpler to use Pytest's [caplog fixture](https://docs.pytest.org/en/stable/how-to/logging.html#caplog-fixture) rather than patching a higher-level guardrails specific file?

nemoguardrails/logging/callbacks.py
```diff
@@ -109,6 +109,8 @@ async def on_chat_model_start(
                     if msg.type == "human"
                     else "Bot"
                     if msg.type == "ai"
+                    else "Tool"
```
tgasser-nv, line 112: Nit: Would be cleaner to have a dict of possible msg.type values to their logging equivalent (e.g. `{"human": "User", ..}`) and do a lookup here.

nemoguardrails/logging/callbacks.py
```diff
@@ -109,6 +109,8 @@ async def on_chat_model_start(
...
+                    if msg.type == "tool"
                     else "System"
```
tgasser-nv, line 114: Why do we assume anything dropping through to the `else:` is a System message? Could you check for membership in `{"system", "developer"}` and return "System". Then use the value of `msg.type` in the `else:` clause?

tgasser-nv (approved):
Looks good, just a couple of cleanups and suggestion on using caplog instead of patching logging ourselves before merging.

tests/test_callbacks.py
Pouyanpi: I think not here, as the test validates the LoggingCallbackHandler's message formatting logic directly, not just logging output. Using caplog would test the wrong abstraction (logging system vs handler behavior)

nemoguardrails/logging/callbacks.py
Pouyanpi: Yes will do in a refactoring PR 👍🏻

nemoguardrails/logging/callbacks.py
Pouyanpi: done in https://github.com/NVIDIA-NeMo/Guardrails/pull/1416

tgasser-nv: Just approved #1416

PR #1416, opened 2025-09-25
fix(logging): correct message type formatting in logs

Pouyanpi:
## Description

[correct message type formatting in logs](https://github.com/NVIDIA-NeMo/Guardrails/commit/5b43c7eaf9678be9b0af28d90c124988cfdd474f)


tgasser-nv (approved):
Looks good, the dict lookup is cleaner than the nested if .. else

PR #1419, opened 2025-09-25
docs(langchain): update with full support and add tool calling guide …

Pouyanpi:
update with full support and add tool calling guide …with examples


docs/user-guides/langchain/runnable-rails.md
```diff
@@ -64,105 +73,285 @@ branch = RunnableBranch(
...
+    input_key="question",    # Default: "input"
+    output_key="answer"      # Default: "output"
+)
+
+# Usage with RAG chain
 rag_chain_with_guardrails = guardrails | rag_chain
 ```
 
-When a guardrail is triggered, and predefined messages must be returned, instead of the output from the LLM, only a dict with the output key is returned:
+<!-- Q. Why is the following suddenly documented? I can't see how this is related to the custom keys. -->
```
miyoungc:
```suggestion
```

PR #1424, opened 2025-09-26
minor doc edits

miyoungc:
## Checklist


docs/user-guides/langchain/runnable-rails.md
```diff
@@ -198,21 +218,24 @@ guardrails = RunnableRails(
 rag_chain_with_guardrails = guardrails | rag_chain
 ```
 
-When a guardrail is triggered and predefined messages must be returned instead of the output from the LLM, only a dict with the output key is returned:
+<!-- Q. Why is the following suddenly documented? I can't see how this is related to the custom keys. -->
```
miyoungc, line 221: @Pouyanpi I have a question for you here.

Pouyanpi: https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/docs/user-guides/langchain/runnable-rails.md?plain=1#L103-L117

This is from the current version

PR #1426, opened 2025-09-29
chore: drop Python 3.9 support ahead of October 2025 EOL

Pouyanpi:
Summary
  - Remove Python 3.9 support ahead of its October 2025 EOL date
  - Update minimum Python requirement to 3.10
  - Clean up all references across CI/CD, testing, and documentation

https://peps.python.org/pep-0596/

  Files Changed
  - CI workflows, tox.ini, devcontainer configs, and docs updated
  - Python version ranges now support 3.10-3.13


tgasser-nv (approved):
Looks good! I did a quick grep and there aren't any references to 3.9 remaining. 

Not needed for this PR, but FYI did a check with pyupgrade to remove deprecated syntax from Python 3.9 in the codebase:

```
$ find . -name '*.py' -exec pyupgrade --py310-plus {} +
```

There were 177 modified files. I spot-checked a few files and the updates were all to type syntax, for example using `list` instead of `List`, `dict` instead of `Dict`, and `typeA  | typeB`  rather than `Optional[TypeA, TypeB]`. I can address these in the ongoing [type-cleaning PRs](https://github.com/NVIDIA-NeMo/Guardrails/pulls?q=is%3Apr+is%3Aopen+chore%28types%29+sort%3Aupdated-desc) after the release.

PR #1428, opened 2025-09-30
ci(workflows): add PyPI publishing workflow with manual approval

Pouyanpi:
Add publish-pypi-approval.yml workflow that triggers after successful builds on version tags. Includes manual approval gate, version validation, PyPI publishing via trusted publishing, draft GitHub release creation, and installation verification.


.github/workflows/publish-pypi-approval.yml
```diff
@@ -0,0 +1,158 @@
...
+            || echo "Release already exists or failed to create"
+
+          rm -f release_notes.md
+
+  verify-pypi-publish:
+    needs: publish-pypi
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.13"]
```
tgasser-nv: Do we need all the versions in-between as well (3.11, 3.12)?

.github/workflows/publish-pypi-approval.yml
```diff
@@ -0,0 +1,158 @@
...
+          echo "version=${VERSION}" >> $GITHUB_OUTPUT
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Wait for PyPI to process package
+        run: |
+          echo "Waiting 120 seconds for PyPI to process the package..."
```
tgasser-nv: Do you have any data on how long this normally takes?

.github/workflows/publish-pypi-approval.yml
```diff
@@ -0,0 +1,158 @@
...
+          sleep 120
+
+      - name: Install from PyPI
+        run: |
+          pip install --upgrade pip
+          pip install "nemoguardrails==${{ steps.version.outputs.version }}" --no-cache-dir
+
+      - name: Start server in the background
```
tgasser-nv: What's the purpose of this? Is it to validate the artifact we're about to upload to PyPI ?

tgasser-nv (approved):
Looks good, left some comments below. How did you test this? Can you add the commands and output to the PR description.

.github/workflows/publish-pypi-approval.yml
Pouyanpi: It tests the published version (what was just published). we actually do similar test in `test-and-build-wheel` workflow, I expect it to pass most of the time. We might remove it later.

.github/workflows/publish-pypi-approval.yml
Pouyanpi: From my limited observation, PyPI seems to process in under 10s, but we use a longer wait to be safe. We can adjust later if needed.

.github/workflows/publish-pypi-approval.yml
Pouyanpi: not really, even one is ok. I am thinking maybe we are better of dropping this last test entirely.

PR #1427, opened 2025-09-30
refactor!: drop reasoning trace extraction logic

Pouyanpi:
## Stack Info
  This PR is part of a stack:
  1. **#1427 (this PR)** 
  2. #1431
  3. #1432
  4. #1434

  **⚠️ Review order: Start here, then proceed through #1431 → #1432 → #1434**

  ## Description
  Replaces custom reasoning trace extraction with LangChain's built-in `additional_kwargs` approach, removing ~1,500 lines of code.

  ## Breaking Changes
  - `reasoning_config` (including `start_token`, `end_token`, `remove_reasoning_traces`) removed from model configuration
  - `apply_to_reasoning_traces` removed from output rails configuration

  ## Migration
  Remove `reasoning_config` sections from config files. Reasoning traces are now automatically extracted by LangChain for supported models (e.g., DeepSeek-R1).
  
  - TODO: 
  - [ ] deprecate reasoning_config from config gracefully

  ## Changes
  - Removed: `ReasoningModelConfig`, `ParsedTaskOutput`, token extraction logic
  - Added: Automatic extraction from `response.additional_kwargs['reasoning_content']`
  - Deleted: 4 test files (~1,085 lines)

PR #1431, opened 2025-10-01
feat: add reasoning trace extraction from llm calls

Pouyanpi:
## Stack Info
  This PR is part of a stack:
  1. #1427
  2. #1431 **(this PR)**
  3. #1432
  4. #1434

  **⚠️ Depends on: #1427**

## Description
 add reasoning trace extraction from llm calls as we are using invoke and ainvoke in llm_call

### DIFF
--- a/tests/test_reasoning_trace_extraction.py
+++ b/tests/test_reasoning_trace_extraction.py
@@ -1,0 +1,306 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


nemoguardrails/actions/llm/utils.py
```diff
@@ -192,15 +205,6 @@ def _store_response_metadata(response) -> None:
                 metadata[field_name] = getattr(response, field_name)
         llm_response_metadata_var.set(metadata)
```
tgasser-nv, line 207: I'm not sure I understand if you're refactoring here to clean things up, shouldn't we call `_store_reasoning_trace()` to maintain the same functionality ?

nemoguardrails/actions/llm/utils.py
```diff
@@ -704,6 +708,12 @@ def extract_tool_calls_from_events(events: list) -> Optional[list]:
     return None
 
 
+def extract_bot_thinking_from_events(events: list):
```
tgasser-nv, line 711: nit: Should this be called 'extract_first_bot_thinking_from_events()' or similar? If the events are ordered with the newest at the start of the list this returns the first one

nemoguardrails/actions/llm/utils.py
Pouyanpi: this function is refactored into two: `store_response_metadata` and `_store_reasoning_traces` to have clear single responsibility and as we are calling both the functionality is preserved.

nemoguardrails/actions/llm/utils.py
Pouyanpi: tried to be consistent with `extract_tool_calls_from_events`. I'll refactor these in a separate PR 👍🏻

PR #1432, opened 2025-10-01
feat: emit BotThinking events with reasoning traces

Pouyanpi:
## Stack Info
  This PR is part of a stack:
  1. #1427
  2. #1431 
  3. #1432 **(this PR)**
  4. #1434
  
  
⚠️ Depends on: 
 
 #1427 → #1431

## Description

Extracts reasoning traces from LLM responses and emits BotThinking
events before BotMessage events. Adds bot_thinking_content to context
for use in output rails.


nemoguardrails/library/self_check/output_check/actions.py
```diff
@@ -61,6 +62,7 @@ async def self_check_output(
             context={
                 "user_input": user_input,
                 "bot_response": bot_response,
+                "bot_thinking": bot_thinking,
```
tgasser-nv, line 65: Are there any issues if we don't use a reasoning model and `bot_thinking` is empty?

examples/configs/self_check_thinking/prompts.yml
```diff
@@ -0,0 +1,42 @@
+prompts:
```
tgasser-nv, line 1: Would a better config name be `self_check_thinking` instead?

examples/configs/self_check_thinking/prompts.yml
```diff
@@ -0,0 +1,42 @@
...
+  - task: self_check_input
```
tgasser-nv, line 2: Why do we need to make any changes to the prompt for self-checking mode for reasoning vs non-reasoning LLMs, now we're using Langchain's parser to split off reasoning tokens?

examples/configs/self_check_thinking/prompts.yml
Pouyanpi: Yes 👍🏻 and I need to mv it to `examples/configs`

nemoguardrails/library/self_check/output_check/actions.py
Pouyanpi: no there won't be any issue.

examples/configs/self_check_thinking/prompts.yml
Pouyanpi: This is an example to show how one can use output rails to check the reasoning content. The user might want to process reasoning content differently which the main purpose of this PR.

PR #1434, opened 2025-10-02
feat(logging): improve event logging and add bot thinking display

Pouyanpi:
## Description

  Enhanced event logging in the runtime and callbacks to better display bot thinking events with reasoning traces. Moved event logging to occur after event generation for proper ordering, added reasoning content extraction from LLM completions, and improved console output formatting for different event types.

## Changes

  - nemoguardrails/colang/v1_0/runtime/runtime.py: Moved event logging to occur after events are generated to ensure proper logging order
  - nemoguardrails/logging/callbacks.py: Extract and include reasoning_content from ChatGeneration messages in completion logs
  - nemoguardrails/logging/verbose.py: Add special formatting for BotThinking events and improve action event display
  - tests/test_runtime_event_logging.py: Add comprehensive tests for event logging behavior including BotThinking, BotMessage, ContextUpdate events and ordering


tests/test_runtime_event_logging.py
```diff
@@ -0,0 +1,160 @@
...
+        chat = TestChat(config, llm_completions=["The answer is 42"])
+
+        await chat.app.generate_events_async(
+            [{"type": "UserMessage", "text": "What is the answer?"}]
+        )
+
+        bot_thinking_logs = [
+            record
+            for record in caplog.records
+            if "Event :: BotThinking" in record.message
```
tgasser-nv, line 46: nit: Should we define an enum with these strings if there are a static set of events? It would be more robust to typos

nemoguardrails/colang/v1_0/runtime/runtime.py
```diff
@@ -186,16 +179,23 @@ async def generate_events(
...
 
-            # Otherwise, we append the event and continue the processing.
-            events.extend(next_events)
-            new_events.extend(next_events)
-
+            # Log all generated events and add them to processing log
             for event in next_events:
                 if event["type"] != "EventHistoryUpdate":
+                    event_type = event["type"]
+                    log.info(
```
tgasser-nv, line 186: nit: Should this be a log.debug() rather than log.info()? How much noise would this make in logs?

nemoguardrails/logging/verbose.py
```diff
@@ -92,7 +92,17 @@ def emit(self, record) -> None:
             elif title == "Event":
                 # For events, we also color differently the type of event.
                 event_name, body = body.split(" ", 1)
-                title = f"[blue]{title}[/] [bold]{event_name}[/]"
+
+                if event_name == "BotThinking":
+                    title = f"[blue]{title}[/] [yellow bold]{event_name}[/]"
+                elif event_name.startswith("Start") and event_name.endswith("Action"):
+                    title = f"[magenta][bold]Start[/]{event_name[5:]}[/]"
```
tgasser-nv: Where does the magic number 5 come from here?

nemoguardrails/logging/verbose.py
```diff
@@ -92,7 +92,17 @@ def emit(self, record) -> None:
...
+                elif event_name.endswith("ActionFinished"):
+                    title = f"[magenta]{event_name[:-8]}[bold]Finished[/][/]"
```
tgasser-nv: Another magic number here, is there a danger we add longer-named events and this breaks?

tgasser-nv (approved):
Looks good, just a few small comments

nemoguardrails/colang/v1_0/runtime/runtime.py
Pouyanpi: it is used in verbose mode (https://github.com/NVIDIA-NeMo/Guardrails/pull/1434/files#diff-d6d2d3f3cdd002a939b4e7841164b14ef3454524eaf0701ca5763ffd7c7d4106L163)

nemoguardrails/logging/verbose.py
Pouyanpi: we actually have these numbers after line 120.  I'm going to revert these changes as it just helped me to debug while looking at the events having different colors.

nemoguardrails/logging/verbose.py
Pouyanpi: we actually have these numbers after line 120.  I'm going to revert these changes as it just helped me to debug while looking at the events having different colors.

tests/test_runtime_event_logging.py
Pouyanpi: I think for the tests it makes sense to use the actual value, but for the implementation I agree, we must refactor all the events that we are using as these are mostly strings scattered all over the codebase.

nemoguardrails/logging/verbose.py
Pouyanpi: reverted

PR #1435, opened 2025-10-02
feat(llm): add provider-agnostic parameter mapping system

Pouyanpi:
Implements flexible LLM parameter transformation to support provider-specific naming conventions (e.g., max_tokens -> max_new_tokens for HuggingFace).

  Key features:
  - Automatic provider inference from LangChain module names
  - Built-in mappings for HuggingFace and Google Vertex AI
  - Custom parameter mapping support via config
  - Parameter dropping capability (map to null)
  - Integration with llm_call function
  
  
 ## TODO:
 - [ ] see if we can reduce the parameter transformation overhead by only doing it when parameter_mapping is given or provider is in the default ones. (nothing critical)

### DIFF
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -30,6 +30,7 @@ from nemoguardrails.context import (
     tool_calls_var,
 )
 from nemoguardrails.integrations.langchain.message_utils import dicts_to_messages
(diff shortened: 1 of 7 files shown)


Pouyanpi:
we won't support this as recent version of langchain providers are using alias for the problematic fields

PR #1436, opened 2025-10-06
feat(cache): Add LFU caching system for models (currently applied to content safety checks)

hazai:
This PR replaces previous (closed) PR #1404

The PR contains a new nemoguardrails/cache folder with an LFU cache implementation (and interface).
The cache can be configured for any model.
It supports:

configuration
stats tracking
logging
thread-safety
(a very minimal) normalization of cache key
@Pouyanpi @tgasser-nv

This PR replaces previous (closed) PR #1404

## Readme

# Content Safety LLM Call Caching

## Overview

The content safety checks in `actions.py` now use an LFU (Least Frequently Used) cache to improve performance by avoiding redundant LLM calls for identical safety checks.

### Cache Configuration

- Per-model caches: Each model gets its own LFU cache instance
- Default capacity: 50,000 entries per model
- Eviction policy: LFU with LRU tiebreaker
- Statistics tracking: Disabled by default (configurable)
- Tracks timestamps: `created_at` and `accessed_at` for each entry
- Cache creation: Automatic when a model is initialized with cache enabled
- Supported model types: Any non-`main` and non-`embeddings` model type (typically content safety models)

### Cached Functions

`content_safety_check_input()` - Caches safety checks for user inputs

### Cache Key Components

The cache key is generated from:

- The rendered prompt (normalized for whitespace)

Since temperature is fixed (1e-20) and stop/max_tokens are derived from the model configuration, they don't need to be part of the cache key.

### How It Works

1. **Before LLM Call**:
   - Generate cache key from request parameters
   - Check if result exists in cache
   - If found, return cached result (cache hit)

2. **After LLM Call**:
   - If not in cache, make the actual LLM call
   - Store the result in cache for future use

### Cache Management

The caching system automatically creates and manages separate caches for each model. Key features:

- **Automatic Creation**: Caches are created when the model is initialized with cache configuration
- **Isolated Storage**: Each model maintains its own cache, preventing cross-model interference
- **Default Settings**: Each cache has 50,000 entry capacity (configurable)
- **Per-Model Configuration**: Cache is configured per model in the YAML configuration

### Statistics and Monitoring

The cache supports detailed statistics tracking and periodic logging for monitoring cache performance:

```yaml
models:
  - type: content_safety
    engine: nim
    model: nvidia/llama-3.1-nemoguard-8b-content-safety
    cache:
      enabled: true
      capacity_per_model: 10000
      store: memory  # Currently only 'memory' is supported
      stats:
        enabled: true      # Enable stats tracking
        log_interval: 60.0 # Log stats every minute
```

**Statistics Features:**

1. **Tracking Only**: Set `stats.enabled: true` with no `log_interval` to track stats without logging
2. **Automatic Logging**: Set both `stats.enabled: true` and `log_interval` for periodic logging

**Statistics Tracked:**

- **Hits**: Number of cache hits (successful lookups)
- **Misses**: Number of cache misses (failed lookups)
- **Hit Rate**: Percentage of requests served from cache
- **Evictions**: Number of items removed due to capacity
- **Puts**: Number of new items added to cache
- **Updates**: Number of existing items updated
- **Current Size**: Number of items currently in cache

**Log Format:**

```
LFU Cache Statistics - Size: 2456/10000 | Hits: 15234 | Misses: 2456 | Hit Rate: 86.11% | Evictions: 0 | Puts: 2456 | Updates: 0
```

**Usage Examples:**

The cache is managed internally by the NeMo Guardrails framework. When you configure a model with caching enabled, the framework automatically:

1. Creates an LFU cache instance for that model
2. Passes the cache to content safety actions via kwargs
3. Tracks statistics if configured
4. Logs statistics at the specified interval

**Configuration Options:**

- `stats.enabled`: Enable/disable statistics tracking (default: false)
- `stats.log_interval`: Seconds between automatic stats logs (None = no logging)

**Notes:**

- Stats logging requires stats tracking to be enabled
- Logs appear at INFO level in the `nemoguardrails.cache.lfu` logger
- Stats are reset when cache is cleared or when `reset_stats()` is called
- Each model maintains independent statistics

### Example Configuration

```yaml
# config.yml
models:
  - type: main
    engine: openai
    model: gpt-4

  - type: content_safety
    engine: nim
    model: nvidia/llama-3.1-nemoguard-8b-content-safety
    cache:
      enabled: true
      capacity_per_model: 50000
      store: memory
      stats:
        enabled: true
        log_interval: 300.0  # Log stats every 5 minutes

rails:
  input:
    flows:
      - content safety check input model="content_safety"
```

### Example Usage

```python

# The cache is automatically configured based on your YAML config
config = RailsConfig.from_path("./config.yml")
rails = LLMRails(config)

# Content safety checks will be cached automatically
response = await rails.generate_async(
    messages=[{"role": "user", "content": "Hello, how are you?"}]
)
```

### Thread Safety

The content safety caching system is **thread-safe** for single-node deployments:

1. **LFUCache Implementation**:
   - Uses `threading.RLock` for all operations
   - All public methods (`get`, `put`, `size`, `clear`, etc.) are protected by locks
   - Supports atomic `get_or_compute()` operations that prevent duplicate computations

2. **LLMRails Model Initialization**:
   - Thread-safe cache creation during model initialization
   - Ensures only one cache instance per model across all threads

3. **Key Features**:
   - **No Data Corruption**: Concurrent operations maintain data integrity
   - **No Race Conditions**: Proper locking prevents race conditions
   - **Atomic Operations**: `get_or_compute()` ensures expensive computations happen only once
   - **Minimal Lock Contention**: Efficient locking patterns minimize performance impact

4. **Usage in Web Servers**:
   - Safe for use in multi-threaded web servers (FastAPI, Flask, etc.)
   - Handles concurrent requests without issues
   - Each thread sees consistent cache state

**Note**: This implementation is designed for single-node deployments. For distributed systems, consider using external caching solutions like Redis.

### Benefits

1. **Performance**: Eliminates redundant LLM calls for identical inputs
2. **Cost Savings**: Reduces API calls to LLM services
3. **Consistency**: Ensures identical inputs always produce identical outputs
4. **Smart Eviction**: LFU policy keeps frequently checked content in cache
5. **Model Isolation**: Each model has its own cache, preventing interference between different safety models
6. **Statistics Tracking**: Monitor cache performance with hit rates, evictions, and more per model
7. **Timestamp Tracking**: Track when entries were created and last accessed
8. **Efficiency**: LFU eviction algorithm ensures the most useful entries remain in cache
9. **Thread Safety**: Safe for concurrent access in multi-threaded environments

### Example Usage Pattern

```python
# First call - takes ~500ms (LLM API call)
result = await content_safety_check_input(
    llms=llms,
    llm_task_manager=task_manager,
    model_name="content_safety",
    context={"user_message": "Hello world"}
)

# Subsequent identical calls - takes ~1ms (cache hit)
result = await content_safety_check_input(
    llms=llms,
    llm_task_manager=task_manager,
    model_name="content_safety",
    context={"user_message": "Hello world"}
)
```

### Logging

The implementation includes debug logging:

- Cache creation: `"Created cache for model '{model_name}' with capacity {capacity}"`
- Cache hits: `"Content safety cache hit for model '{model_name}'"`
- Cache stores: `"Content safety result cached for model '{model_name}'"`

Enable debug logging to monitor cache behavior:

```python
import logging
logging.getLogger("nemoguardrails.library.content_safety.actions").setLevel(logging.DEBUG)
```

### DIFF
--- a/tests/test_cache_interface.py
+++ b/tests/test_cache_interface.py
@@ -1,0 +1,104 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+from typing import Any
+
+import pytest
+
+from nemoguardrails.llm.cache.interface import CacheInterface
+
+
+class MinimalCache(CacheInterface):
+    def __init__(self, maxsize: int = 10):
+        self._cache = {}
+        self._maxsize = maxsize
+
+    def get(self, key: Any, default: Any = None) -> Any:
+        return self._cache.get(key, default)
+
+    def put(self, key: Any, value: Any) -> None:
+        self._cache[key] = value
+
+    def size(self) -> int:
+        return len(self._cache)
+
+    def is_empty(self) -> bool:
+        return len(self._cache) == 0
+
+    def clear(self) -> None:
+        self._cache.clear()
+
+    @property
+    def maxsize(self) -> int:
+        return self._maxsize
+
+
+@pytest.fixture
+def cache():
+    return MinimalCache()
+
+
+def test_contains(cache):
+    cache.put("key1", "value1")
+    assert cache.contains("key1")
+
+
+def test_get_stats(cache):
+    stats = cache.get_stats()
+    assert isinstance(stats, dict)
+    assert "message" in stats
+
+
+def test_reset_stats(cache):
+    cache.reset_stats()
+
+
+def test_log_stats_now(cache):
+    cache.log_stats_now()
+
+
+def test_supports_stats_logging(cache):
+    assert cache.supports_stats_logging() is False
+
+
+@pytest.mark.asyncio
+async def test_get_or_compute_cache_hit(cache):
+    cache.put("key1", "cached_value")
+
+    async def compute_fn():
+        return "computed_value"
+
+    result = await cache.get_or_compute("key1", compute_fn)
+    assert result == "cached_value"
+
+
+@pytest.mark.asyncio
+async def test_get_or_compute_cache_miss(cache):
+    async def compute_fn():
+        return "computed_value"
+
+    result = await cache.get_or_compute("key1", compute_fn)
+    assert result == "computed_value"
+    assert cache.get("key1") == "computed_value"
+
+
+@pytest.mark.asyncio
+async def test_get_or_compute_exception(cache):
+    async def failing_compute():
+        raise ValueError("Computation failed")
+
+    result = await cache.get_or_compute("key1", failing_compute, default="fallback")
+    assert result == "fallback"


nemoguardrails/cache/README.md
```diff
@@ -0,0 +1,218 @@
...
+```python
+from nemoguardrails import RailsConfig, LLMRails
+
+# Method 1: Using context manager
+config = RailsConfig.from_path("./config.yml")
+with LLMRails(config) as rails:
+    # Content safety checks will be cached automatically
+    response = await rails.generate_async(
+        messages=[{"role": "user", "content": "Hello, how are you?"}]
+    )
```
Pouyanpi: I think this is from the old PR

nemoguardrails/cache/README.md
```diff
@@ -0,0 +1,218 @@
...
+### Thread Safety
+
+The content safety caching system is **thread-safe** for single-node deployments:
+
+1. **LFUCache Implementation**:
+   - Uses `threading.RLock` for all operations
+   - All public methods (`get`, `put`, `size`, `clear`, etc.) are protected by locks
+   - Supports atomic `get_or_compute()` operations that prevent duplicate computations
+
+2. **ContentSafetyManager**:
```
Pouyanpi: same here

nemoguardrails/cache/README.md
Pouyanpi, line 1: seems that part of this doc belongs to the previous implementation. Also we don't include README.md in package but this is very helpful for us to review. We might consider removing it when merging.

nemoguardrails/library/content_safety/actions.py
```diff
@@ -27,6 +29,31 @@
...
+PROMPT_PATTERN_WHITESPACES = re.compile(r"\s+")
+
+
+def _create_cache_key(prompt: Union[str, List[str]]) -> str:
+    """Create a cache key from the prompt."""
+    # can the prompt really be a list?
+    if isinstance(prompt, list):
+        prompt_str = json.dumps(prompt)
+    else:
+        prompt_str = prompt
```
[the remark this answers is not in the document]
Pouyanpi: besides the copilot command.
I suggest moving it to a `utils.py` in `cache` package.

nemoguardrails/rails/llm/config.py
```diff
@@ -15,6 +15,8 @@
 
 """Module for the configuration of rails."""
 
+from __future__ import annotations
```
Pouyanpi: would we need this if we move `CacheStatsConfig` and `ModelCacheConfig` above `Model` ?

nemoguardrails/rails/llm/config.py
```diff
@@ -897,6 +905,41 @@ def get_api_key(self) -> Optional[str]:
...
+    )
+    capacity_per_model: int = Field(
+        default=50000, description="Maximum number of entries in the cache per model"
+    )
+    store: str = Field(
+        default="memory", description="Cache store: 'memory', 'filesystem', 'redis'"
+    )
+    store_config: Dict[str, Any] = Field(
+        default_factory=dict, description="Backend-specific configuration"
+    )
```
Pouyanpi: please remove it as it is not used right now. later we will include it if needed.

nemoguardrails/rails/llm/config.py
```diff
@@ -897,6 +905,41 @@ def get_api_key(self) -> Optional[str]:
...
+    enabled: bool = Field(
+        default=False,
+        description="Whether caching is enabled (default: False - no caching)",
+    )
+    capacity_per_model: int = Field(
+        default=50000, description="Maximum number of entries in the cache per model"
+    )
+    store: str = Field(
+        default="memory", description="Cache store: 'memory', 'filesystem', 'redis'"
+    )
```
Pouyanpi: If I recall correctly we mentioned to keep this field and we might extend it in future as currently we support only one store type I think removing it is better ?

nemoguardrails/library/content_safety/actions.py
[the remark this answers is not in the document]
Pouyanpi: also  better to create a deterministic cache key from the prompt.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  For example use SHA256 hashing to create a consistent key from the normalized prompt ...

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -525,6 +527,39 @@ def _init_llms(self):
...
+                        cache = LFUCache(
+                            capacity=model.cache.capacity_per_model,
+                            track_stats=model.cache.stats.enabled,
+                            stats_logging_interval=stats_logging_interval,
+                        )
+
+                        log.info(
+                            f"Created cache for model '{model.type}' with capacity {model.cache.capacity_per_model}"
+                        )
+
```
Pouyanpi: move it to a method like 

```python
    def _init_model_caches(self):
        """
        Initialize caches for models that have caching configured.

        creates per-model cache instances and registers them as action parameters.
        only models that are not 'main' or 'embeddings' types are eligible for caching.
        """
```

nemoguardrails/library/content_safety/actions.py
```diff
@@ -75,6 +102,21 @@ async def content_safety_check_input(
 
     max_tokens = max_tokens or _MAX_TOKENS
 
+    # Check cache if available for this model
+    cached_result = None
+    cache_key = None
+
+    # Try to get the model-specific cache
+    cache = kwargs.get(f"model_cache_{model_name}")
```
Pouyanpi: nit: can we have a util?

```python
def _get_model_cache(model_name: str, kwargs: Dict[str, Any]) -> Optional[Any]:
    """
    Retrieve the cache instance for a specific model.

    Args:
        model_name: The name of the model to get the cache for.
        kwargs: Action parameters dictionary containing cache instances.

    Returns:
        The cache instance if available, None otherwise.
    """
```

nemoguardrails/llm/cache/lfu.py
Pouyanpi, line 1: still see file and persistence related implementation. Should we remove them for now?

nemoguardrails/llm/cache/lfu.py
```diff
@@ -0,0 +1,677 @@
...
+                # Create new node and add to cache
+                new_node = LFUNode(key, value)
+                self.key_map[key] = new_node
+
+                # Add to frequency 1 list
+                if 1 not in self.freq_map:
+                    self.freq_map[1] = DoublyLinkedList()
+
+                self.freq_map[1].append(new_node)
+                self.min_freq = 1
```
Pouyanpi, line 216: Following test would fail (is it expected?):

```python
    def test_eviction_clears_frequency_map(self):
        cache = LFUCache(2)
        cache.put("a", 1)
        cache.put("b", 2)

        cache.put("c", 3)

        assert 1 not in cache.freq_map or cache.freq_map[1].size == 0
```

when items are evicted, the frequency 1 list still contains 2 items instead of being empty. This shows that when putting new items at frequency 1, the old evicted nodes from frequency 1 are not properly removed from the doubly linked list.

I think the root cause is in put() method at 237-238, when adding a new item at frequency 1, the code doesn't check if items were just evicted from that same frequency list. The eviction removes the node from key_map but the frequency list cleanup only happens in _update_node_freq() when old_list.size == 0.

this is a subtle memory leak where the frequency map's linked lists can contain stale node references after evictions.

nemoguardrails/llm/cache/lfu.py
Pouyanpi, line 1: could find 3 bugs:

persistence related:

Bug 1.
cache exceeds configured capacity, violates invariants this is a good reason to remove the persistence for now:

```python
    def test_load_exceeds_capacity():
        temp_file = tempfile.mktemp()

        cache1 = LFUCache(10, persistence_interval=1.0, persistence_path=temp_file)
        for i in range(10):
            cache1.put(f"key_{i}", i)
        cache1.persist_now()

        cache2 = LFUCache(3, persistence_interval=1.0, persistence_path=temp_file)

        assert cache2.size() <= cache2.capacity, \
            f"Cache size {cache2.size()} exceeds capacity {cache2.capacity}"

        os.remove(temp_file)

    def test_load_respects_new_capacity():
        temp_file = tempfile.mktemp()

        cache1 = LFUCache(10, persistence_interval=1.0, persistence_path=temp_file)
        for i in range(10):
            cache1.put(f"key_{i}", i)
        cache1.persist_now()

        cache2 = LFUCache(5, persistence_interval=1.0, persistence_path=temp_file)
        cache2.put("new_key", "new_value")

        assert cache2.size() <= cache2.capacity

        os.remove(temp_file)
```

also both `_persist_to_disk()` and `_load_from_disk()` silently swallow ALL exceptions, giving no indication when persistence fails.

Bug 2.
I think there is also anoter potential issue: when nodes are evicted via `DoublyLinkedList.pop()`, their `prev` and `next` pointers are not cleared which means evicted nodes still reference active nodes in the list. It could prevent garbage collection.

Bug 3.

memory leak:
as commented later in the code:

when evicting items from a frequency list that contains multiple nodes, the evicted node is removed from `key_map` but the frequency list itself is not cleaned up properly

nemoguardrails/library/content_safety/actions.py
```diff
@@ -27,6 +29,31 @@
...
+def _create_cache_key(prompt: Union[str, List[str]]) -> str:
+    """Create a cache key from the prompt."""
+    # can the prompt really be a list?
+    if isinstance(prompt, list):
+        prompt_str = json.dumps(prompt)
+    else:
+        prompt_str = prompt
+
+    # normalize the prompt to a string
+    # should we do more normalizations?
```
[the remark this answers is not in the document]
hazai: Changed the comment.
However, my question still holds: do we want to do more normalizations now?

nemoguardrails/library/content_safety/actions.py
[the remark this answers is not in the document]
hazai: @Pouyanpi I intentionally put the comment "an the prompt really be a list?" here, as I am not sure. Can you answer it.
Re key hashing, I did it intentionally too: wanted the key to be interpretable. OTOH, it can be a large key, so  hashing can be a good idea in this case.
Re moving to `utils.py`: you mean the _create_cache_key function, right? The normalization here is content_safety specific, so if we do that, I believe we should pass the string normalized, or pass a normalization function.
Waiting for feedback here

nemoguardrails/llm/cache/lfu.py
```diff
@@ -0,0 +1,677 @@
...
+            persistence_path: Path to persistence file (defaults to 'lfu_cache.json' if persistence enabled)
+            stats_logging_interval: Seconds between periodic stats logging (None disables logging)
+        """
+        if capacity < 0:
+            raise ValueError("Capacity must be non-negative")
+
+        self._capacity = capacity
+        self.track_stats = track_stats
+        self._lock = threading.RLock()  # Thread-safe access
+        self._computing: dict[Any, asyncio.Future] = {}  # Track keys being computed
```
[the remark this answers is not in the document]
hazai: We are Python 3.9+ compatible, so it shouldn't be an issue

nemoguardrails/cache/README.md
hazai: Fixed

nemoguardrails/cache/README.md
hazai: Fixed

nemoguardrails/cache/README.md
hazai: Fixed

nemoguardrails/rails/llm/config.py
hazai: Fixed

nemoguardrails/rails/llm/config.py
hazai: I'm not sure... I think we perhaps discussed another field that we wanted to add?

nemoguardrails/library/content_safety/actions.py
hazai: You mean moving it into `utils.py`? not sure it should be there...

nemoguardrails/llm/cache/lfu.py
hazai: Not sure what you mean here:
At the point the assert is made, the freq_map should hold a single key (1) with a linkedlist holding the items "b" and "c".
Thus, 1 is in cache.freq_map AND cache.freq_map[1].size > 0.
So it certainly fails (I also checked that).
The removal from both key_map and freq_map happens in `_evict_lfu`

nemoguardrails/llm/cache/lfu.py
hazai: Bug 1:
Good catch. Although persistence is not used anymore, and it's not that hard to fix, I will also remove its implementation.

Bug 2:
I see no bug here... check out `DoublyLinkedList::pop()`(line 63 at `lfu.py`) -- it is redirecting the pointers (also note the head and tail dummy pointers)

Bug 3:
If you're referring to your previous comment, then not a bug as well (see above)

.vscode/settings.json
Pouyanpi, line 1: let's remove this change 👍🏻

nemoguardrails/library/content_safety/actions.py
```diff
@@ -27,6 +29,31 @@
...
+    return PROMPT_PATTERN_WHITESPACES.sub(" ", prompt_str).strip()
```
Pouyanpi: @tgasser-nv what do you think about normalization here?

Pouyanpi:
@hazai for telemetry and logging, we can get `llm_stats_var` after `llm_call`:


```python
    result = await llm_call(
        llm,
        check_input_prompt,
        stop=stop,
        llm_params={"temperature": 1e-20, "max_tokens": max_tokens},
    )
    print("llm_stats_var:", llm_stats_var.get())
```
    
  so similar to the result it should be cached, but the second time when it is read from the cache we should set it
  
  `llm_stat_var.set(value_from_cache)`
  
  
  langchain for example returns the token usage metrics as is but change the duration to the cache read duration.

nemoguardrails/library/content_safety/actions.py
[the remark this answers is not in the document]
Pouyanpi: @tgasser-nv this is the question about normalization as it affects hit rate and correctness.

nemoguardrails/llm/cache/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""General-purpose caching utilities for NeMo Guardrails."""
+
+from nemoguardrails.llm.cache.interface import CacheInterface
+from nemoguardrails.llm.cache.lfu import LFUCache
+from nemoguardrails.llm.cache.utils import create_normalized_cache_key
```
Pouyanpi:
```suggestion
```

PR #1451, opened 2025-10-16
review: PR #1436

Pouyanpi:
Cache:
  - Added create_normalized_cache_key() utility for consistent cache key generation with whitespace normalization
  - Renamed capacity_per_model → maxsize in ModelCacheConfig for clarity
  - Implemented _create_model_cache() and _initialize_model_caches() in LLMRails for dynamic cache creation
  - Registered caches as model_caches dict in runtime action parameters

  Content Safety Integration:
  - Modified content_safety_check_input() to use centralized cache registry (model_caches)
  - Cache entries now store both result and LLM stats for accurate metrics restoration
  - Added _restore_llm_stats_from_cache() and _extract_llm_stats_for_cache() helpers
  - Cache hits restore duration, token counts, and set from_cache=True flag

  Observability:
  - Added from_cache field to LLMCallInfo for cache hit tracking
  - Enhanced span extraction to include from_cache in custom attributes
  - Enabled tracing in content safety example config

  Testing:
  - Added test_cache_utils.py with comprehensive normalization tests
  - Added test_content_safety_cache.py with cache storage/retrieval tests
  - Added test_integration_cache.py with end-to-end integration tests
  - Added test_llmrails.py tests for cache initialization and config validation
  - Updated existing cache tests to use new maxsize parameter

Test Plan:

run https://gist.github.com/Pouyanpi/3461b5310a17da17097280b5efc9cebe

TODO: 
- [ ] push the restructuring commit that moves `cache` to `llm` package once review is done.

### DIFF
--- a/nemoguardrails/library/content_safety/actions.py
+++ b/nemoguardrails/library/content_safety/actions.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 import logging
-from time import time
 from typing import Dict, Optional
 
 from langchain_core.language_models.llms import BaseLLM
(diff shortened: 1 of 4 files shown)


nemoguardrails/library/content_safety/actions.py
```diff
@@ -13,46 +13,59 @@
...
+
+def _extract_llm_stats_for_cache() -> Optional[dict]:
+    llm_call_info = llm_call_info_var.get()
+    if llm_call_info:
+        return {
+            "total_tokens": llm_call_info.total_tokens or 0,
+            "prompt_tokens": llm_call_info.prompt_tokens or 0,
+            "completion_tokens": llm_call_info.completion_tokens or 0,
+        }
+    return None
```
Pouyanpi: move these to utils.py

nemoguardrails/library/content_safety/actions.py
```diff
@@ -103,21 +117,36 @@ async def content_safety_check_input(
...
+            if processing_log:
+                llm_call_info = llm_call_info_var.get()
+                if llm_call_info:
+                    processing_log.append(
+                        {
+                            "type": "llm_call_info",
+                            "timestamp": time(),
+                            "data": llm_call_info,
+                        }
+                    )
```
Pouyanpi: should be moved to utils

nemoguardrails/library/content_safety/actions.py
```diff
@@ -131,9 +160,12 @@ async def content_safety_check_input(
...
     final_result = {"allowed": is_safe, "policy_violations": violated_policies}
 
-    # Store in cache if available
-    if cache_key and cache:
-        cache.put(cache_key, final_result)
+    if cache:
+        cache_entry = {
+            "result": final_result,
+            "llm_stats": _extract_llm_stats_for_cache(),
+        }
```
Pouyanpi, line 113: make cache entry struct

examples/configs/content_safety/config.yml
```diff
@@ -22,3 +22,5 @@ rails:
   output:
     flows:
       - content safety check output $model=content_safety
+tracing:
```
hazai, line 25: Should it be at this level? or per model?

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -527,47 +527,60 @@ def _init_llms(self):
...
+    def _initialize_model_caches(self) -> None:
+        """Initialize caches for configured models."""
+        model_caches: Optional[Dict[str, CacheInterface]] = dict()
+        for model in self.config.models:
+            if model.type in ["main", "embeddings"]:
+                continue
+
+            if model.cache and model.cache.enabled:
+                cache = self._create_model_cache(model)
+                model_caches[model.type] = cache
```
hazai, line 574: Shouldn't the cache be assigned to a specific model (not by type, but by name)?
Why not keep using `register_action_param`? (asking mainly for understanding)

hazai (approved):
Approved with some comments

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: this is intentional to only allow it for content safety or topic control. Also `model_name` is not always available where we are doing caching.

regarding the `register_action_param`. we are still doing it but the registered param name should be directly defined in actions signature: so if it was f"model_cache_{model_name}" then we should define the actual value in actions signature, e.g. `content_safety_check_input(..., model_cache_nemoguard-...)` which is not practical. This is similar to what we are doing for llms : `self.runtime.register_action_param("llms", llms)`.

examples/configs/content_safety/config.yml
Pouyanpi: yes, tracing is available for the complete guardrails instance

PR #1455, opened 2025-10-17
style(cache): replace pass with ellipsis in abstract methods

Pouyanpi:
Updates abstract method implementations in CacheInterface to use ellipsis (...) instead of pass, following modern Python conventions for abstract method stubs.

## Part of Stack
This is PR 1/5 in the NeMoGuards caching feature stack.
- ✅ **PR 1: Cache interface cleanup** (this PR)
- ⬜ PR 2: #1456 
- ⬜ PR 3: #1457 
- ⬜ PR 4: #1458 
- ⬜ PR 5: #1459

### DIFF
--- a/nemoguardrails/llm/cache/interface.py
+++ b/nemoguardrails/llm/cache/interface.py
@@ -44,7 +44,7 @@ class CacheInterface(ABC):
         Returns:
             The value associated with the key, or default if not found.
         """
-        pass
+        ...
 
     @abstractmethod
     def put(self, key: Any, value: Any) -> None:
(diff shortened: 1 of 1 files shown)


Pouyanpi:
this change brings coverage to 100% from 87% (see https://github.com/NVIDIA-NeMo/Guardrails/pull/1436#issuecomment-3371671795)

tgasser-nv (approved):
Looks good, definitely prefer `...` to `pass`. We could also use raise a `NotImplementedError` but this is fine

PR #1461, opened 2025-10-20
feat(llm): Add custom HTTP headers support to ChatNVIDIA provider

Pouyanpi:
## Description

Add custom HTTP headers support to the ChatNVIDIA class patch, enabling users to pass custom headers
(authentication tokens, request IDs, billing information, etc.) with all requests to NVIDIA AI endpoints.

example config

```yaml
models:
  - type: main
    engine: nvidia_ai_endpoints
    model: meta/llama-3.1-8b-instruct
    parameters:
      base_url: http://localhost:8000/v1
      custom_headers:
        X-Model-Auth: test-bearer-token
        X-Request-ID: "12345"
        X-Billing-Project: project-abc
        X-Trace-ID: trace-xyz
```

The feat is verified on microservice side.

## Implementation

- Added custom_headers optional field to ChatNVIDIA class with Pydantic v2 compatibility
- Implemented runtime method wrapping that intercepts _client.get_req() and _client.get_req_stream()
to merge custom headers with existing headers
- Included automatic version detection to ensure compatibility with langchain-nvidia-ai-endpoints >=
0.3.0, with clear error messages for older versions
- Works with both synchronous invoke() and streaming requests, fully compatible with VLM (Vision
Language Models)
## Checklist

### DIFF
--- a/nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
+++ b/nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
@@ -63,22 +63,20 @@ class ChatNVIDIA(ChatNVIDIAOriginal):  # pragma: no cover
     def __init__(self, **kwargs: Any):
         super().__init__(**kwargs)
         if self.custom_headers:
-            super().__init__(**kwargs)
             custom_headers_error = (
                 "custom_headers requires langchain-nvidia-ai-endpoints >= 0.3.0. "
                 "Your version does not support the required client structure or "
                 "extra_headers parameter. Please upgrade: "
                 "pip install --upgrade langchain-nvidia-ai-endpoints>=0.3.0"
             )
-            if self.custom_headers:
-                if not hasattr(self._client, "get_req"):
-                    raise RuntimeError(custom_headers_error)
+            if not hasattr(self._client, "get_req"):
+                raise RuntimeError(custom_headers_error)
 
-                sig = inspect.signature(self._client.get_req)
-                if "extra_headers" not in sig.parameters:
-                    raise RuntimeError(custom_headers_error)
+            sig = inspect.signature(self._client.get_req)
+            if "extra_headers" not in sig.parameters:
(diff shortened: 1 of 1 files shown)


nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
```diff
@@ -51,10 +52,50 @@ def wrapper(
...
+        if self.custom_headers:
+            if not hasattr(self._client, "get_req"):
+                raise RuntimeError(
+                    "custom_headers requires langchain-nvidia-ai-endpoints >= 0.2.1. "
+                    "Your version uses a nested client structure that is not supported. "
+                    "Please upgrade: pip install --upgrade langchain-nvidia-ai-endpoints>=0.3.0"
+                )
+
+            sig = inspect.signature(self._client.get_req)
+            if "extra_headers" not in sig.parameters:
```
trebedea: Why do we have two conditions for the same RuntimeError message? 
And it's unclear why in the first one it says both versions `0.2.1` and `0.3.0`.

If it's the same message can you add a bool that merges the two conditions?

nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
```diff
@@ -51,10 +52,50 @@ def wrapper(
...
+
+    def _wrap_client_methods(self):
+        original_get_req = self._client.get_req
+        original_get_req_stream = self._client.get_req_stream
+
+        def wrapped_get_req(payload: dict = {}, extra_headers: dict = {}):
+            merged_headers = {**extra_headers, **self.custom_headers}
+            return original_get_req(payload=payload, extra_headers=merged_headers)
+
+        def wrapped_get_req_stream(payload: dict, extra_headers: dict = {}):
```
trebedea:
```suggestion
        def wrapped_get_req_stream(payload: dict = {}, extra_headers: dict = {}):
```

PR #1463, opened 2025-10-20
test: fix flaky stats logging interval timing test

Pouyanpi:
## Summary
Fixes a flaky test in `test_cache_lfu.py` that was intermittently failing on CI due to timing dependencies.

## Problem
The test relied on `time.sleep(0.05)` with an assumption that 10 iterations would take < 1.0 seconds. On CI, execution overhead from cache operations and system load caused cumulative delays that would occasionally push total time over 1.0 seconds, triggering unexpected stats logging and causing the test to fail with `AssertionError: 1 != 0`.

## Solution
Mock `time.time()` with controlled values that advance exactly 0.05 seconds per iteration, removing all timing dependencies.

## Changes
- Replaced real `time.sleep()` calls with mocked `time.time()` in `test_stats_logging_interval_timing`
- Eliminates execution overhead and scheduling delays that caused test flakiness
- Test now runs deterministically across all environments

## Benefits
- Eliminates flakiness on CI
- Faster test execution (~0.02s vs ~1.16s, 58x speedup)
- Deterministic behavior across all environments

## Test plan
- [x] Test passes locally (5/5 runs)
- [x] Pre-commit hooks pass
- [ ] CI validates fix resolves flakiness

### DIFF
--- a/tests/test_cache_lfu.py
+++ b/tests/test_cache_lfu.py
@@ -553,21 +553,24 @@ class TestLFUCacheStatsLogging(unittest.TestCase):
 
         with patch.object(
             logging.getLogger("nemoguardrails.llm.cache.lfu"), "info"
-        ) as mock_log:
-            # Multiple operations within interval
+        ) as mock_log, patch("time.time") as mock_time:
+            current_time = [0.0]
+
+            def time_side_effect():
+                return current_time[0]
+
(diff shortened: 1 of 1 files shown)


tgasser-nv (approved):
Looks much better, thanks for fixing this

PR #1465, opened 2025-10-20
chore(logging): Log model name and base URL before invoking LLMs

JashG:
## Description
This PR logs the model name and base URL before an LLM is invoked.

Some context: From the [NeMo Guardrails microservice](https://docs.nvidia.com/nemo/microservices/latest/guardrails/index.html), we've run into different errors that can be root caused to the incorrect base URL being used for a model. This can be difficult to root cause, since neither the URL nor model gets logged at any point in the call chain.

## Checklist


JashG:
@Pouyanpi thanks for that pointer! I agree, centralizing the logic in `LoggingCallbackHandler` makes more sense. The one minor drawback: `repr` is the string representation of the provider class, rather than an instance of the class, so my logic parses that string for the model and base URL properties. Let me know if you have any concerns with this approach.

PR #1467, opened 2025-10-22
fix(runtime): ensure stop flag is set for policy violations in parallel rails

Pouyanpi:
When rails ran in parallel, the stopped rail's processing log was excluded from aggregation, causing `compute_generation_log()` to never create an `ActivatedRail` entry for the stopped rail. Additionally,
  the `post_event` (`OutputRailFinished`/`InputRailFinished`) was incorrectly added even when a rail stopped.

### DIFF
--- a/nemoguardrails/colang/v1_0/runtime/runtime.py
+++ b/nemoguardrails/colang/v1_0/runtime/runtime.py
@@ -367,7 +367,7 @@ class RuntimeV1_0(Runtime):
             unique_flow_ids[flow_uid] = task
 
         stopped_task_results: List[dict] = []
-        stopped_task_processing_log: List[dict] = []
+        stopped_task_processing_logs: List[dict] = []
 
         # Process tasks as they complete using as_completed
         try:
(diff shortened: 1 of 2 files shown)


Pouyanpi:
@greptileai

nemoguardrails/colang/v1_0/runtime/runtime.py
```diff
@@ -433,6 +443,15 @@ async def task_call_helper(flow_uid, post_event, func, *args, **kwargs):
             finished_task_processing_logs.extend(task_processing_logs[flow_id])
 
         if processing_log:
+            for plog in stopped_task_processing_log:
```
trebedea: Should we create a helper function not to duplicate this code? Same logic is for `finished_task_processing_logs`.

trebedea (approved):
👍 LGTM

nemoguardrails/colang/v1_0/runtime/runtime.py
Pouyanpi: done 👍🏻

PR #1470, opened 2025-10-24
fix(streaming)!: raise error when stream_async used with disabled output rails streaming

Pouyanpi:
## Description

Fixes a bug where `stream_async()` would hang or behave unpredictably when output rails are configured but `output.streaming.enabled` is `False` or not set.

## Problem

When users configured output rails but either:
1. Set `rails.output.streaming.enabled = False`, or
2. Didn't configure `rails.output.streaming` at all (defaults to disabled)

And then called `stream_async()`, the system would enter an incompatible state:
- The streaming handler expected real-time token streaming
- But output rails tried to run in blocking/non-streaming mode on the complete response
- This resulted in hangs, undefined behavior, or silent failures

## Fix

Added validation at the start of `stream_async()` to detect this misconfiguration and raise a clear, actionable `ValueError` with guidance on how to fix it.

The error message tells users to either:
- Set `rails.output.streaming.enabled = True` in their configuration, or
- Use `generate_async()` instead of `stream_async()`

### DIFF
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -1259,6 +1259,16 @@ class LLMRails:
     ) -> AsyncIterator[str]:
         """Simplified interface for getting directly the streamed tokens from the LLM."""
 
+        if len(self.config.rails.output.flows) > 0 and (
+            not self.config.rails.output.streaming
+            or not self.config.rails.output.streaming.enabled
+        ):
+            raise ValueError(
+                "stream_async() cannot be used when output rails are configured but "
+                "output.streaming.enabled is False. Either set "
+                "rails.output.streaming.enabled to True in your configuration, or use "
+                "generate_async() instead of stream_async()."
+            )
         # if an external generator is provided, use it directly
         if generator:
             if (
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -1255,7 +1255,7 @@ class LLMRails:
         ):
             raise ValueError(
                 "stream_async() cannot be used when output rails are configured but "
-                "output.streaming.enabled is False. Either set "
+                "rails.output.streaming.enabled is False. Either set "
                 "rails.output.streaming.enabled to True in your configuration, or use "
                 "generate_async() instead of stream_async()."
             )
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -1248,6 +1248,18 @@ class LLMRails:
                     new_message["tool_calls"] = tool_calls
                 return new_message
 
+    def _validate_streaming_with_output_rails(self) -> None:
+        if len(self.config.rails.output.flows) > 0 and (
+            not self.config.rails.output.streaming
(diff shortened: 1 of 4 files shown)


nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1248,6 +1248,18 @@ async def generate_async(
...
                 return new_message
 
+    def _validate_streaming_with_output_rails(self) -> None:
+        if len(self.config.rails.output.flows) > 0 and (
+            not self.config.rails.output.streaming
+            or not self.config.rails.output.streaming.enabled
+        ):
+            raise ValueError(
+                "stream_async() cannot be used when output rails are configured but "
+                "output.streaming.enabled is False. Either set "
```
tgasser-nv: typo: Should this be `rails.output.streaming.enabled` to match the line below?

tgasser-nv (commented):
Looks good, just a couple of items before merging:

* Could you add more tests to cover the 8 permutations of {stream_async, generate_async}, output rail streaming enabled, and RailsConfig.streaming? I expect some of these are already covered in other tests. For example (generate_async, output rail streaming.enabled False, Railsconfig.streaming False). 
* I understand we need to wait until inference-time for the call to either `generate_async` or `stream_async` to figure out whether the client actually wants to stream output or not. Are there any cases we can detect statically from Pydantic validators that we know wouldn't work regardless of which method the client calls?

Pouyanpi:
> Looks good, just a couple of items before merging:
> 
> * Could you add more tests to cover the 8 permutations of {stream_async, generate_async}, output rail streaming enabled, and RailsConfig.streaming? I expect some of these are already covered in other tests. For example (generate_async, output rail streaming.enabled False, Railsconfig.streaming False).

I checked and currently we are testing all permutations


1. stream_async + enabled=True + config.streaming=True → test_streaming_output_rails_allowed
2. stream_async + enabled=True + config.streaming=False → test_stream_async_with_output_rails_streaming_enabled_rails_config_streaming_false
3. stream_async + enabled=False + config.streaming=True → test_streaming_with_output_rails_disabled_raises_error 
4. stream_async + enabled=False + config.streaming=False → test_streaming_with_output_rails_no_streaming_config_raises_error
generate_async tests (4 permutations):

5. generate_async + enabled=True + config.streaming=True → test_generate_async_with_output_rails_streaming_enabled_rails_config_streaming_true 
6. generate_async + enabled=True + config.streaming=False → test_generate_async_with_output_rails_streaming_enabled_rails_config_streaming_false 
7. generate_async + enabled=False + config.streaming=True → test_generate_async_with_output_rails_streaming_disabled_rails_config_streaming_true 
8. generate_async + enabled=False + config.streaming=False → test_generate_async_with_output_rails_streaming_disabled_rails_config_streaming_false 

> *I understand we need to wait until inference-time for the call to either generate_async or stream_async to figure out whether the client actually wants to stream output or not. Are there any cases we can detect statically from Pydantic validators that we know wouldn't work regardless of which method the client calls

No, that's all we can do for now. Later when we drop top level streaming field makes it much cleaner.

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: Done 👍🏻

PR #1471, opened 2025-10-24
chore(llm): deprecate nemoguardrails.llm.params module in favor of direct parameter passing

Pouyanpi:
This commit adds deprecation warnings to the `nemoguardrails.llm.params` module, which will be removed in version 0.19.0. Users should migrate to passing parameters directly to `llm_call()` using the `llm_params` argument instead of using the context manager approach. 
#1387 

### Migration Guide

Users should replace the deprecated context manager approach with direct parameter passing:

**Old way (deprecated):**
```python
from nemoguardrails.llm.params import llm_params
with llm_params(llm, temperature=0.7):
    result = await llm_call(llm, prompt)
```

New way (recommended):
```python
result = await llm_call(llm, prompt, llm_params={"temperature": 0.7})
```

### DIFF
--- a/nemoguardrails/llm/params.py
+++ b/nemoguardrails/llm/params.py
@@ -17,20 +17,48 @@
 Module for providing a context manager to temporarily adjust parameters of a language model.
 
 Also allows registration of custom parameter managers for different language model types.
+
+.. deprecated:: 0.17.0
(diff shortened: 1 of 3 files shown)


tgasser-nv (approved):
Looks good. I think we need an action on the docs-side as well to show this will be deprecated, and use the new syntax in our example code

PR #1474, opened 2025-10-27
fix(llm): add fallback extraction for reasoning traces from <think> tags

Pouyanpi:
## Description

Adds a compatibility layer for LLM providers that don't properly populate `reasoning_content` in `additional_kwargs`. When `reasoning_content` is missing, the system now falls back to extracting reasoning traces from `<think>...</think>` tags in the response content and removes the tags from the final output.

This fixes compatibility with certain NVIDIA models (e.g., `nvidia/llama-3.3-nemotron-super-49b-v1.5`) in `langchain-nvidia-ai-endpoints` that include reasoning traces in `<think>` tags but fail to populate the `reasoning_content` field. 

All reasoning models using ChatNVIDIA should expose reasoning content consistently through the same interface

## Changes
- **primary path**: extracts from `additional_kwargs["reasoning_content"]` (properly implemented models)
- **fallback path**: extracts from `<think>...</think>` tags when `reasoning_content` is missing
- **side effect**: removes `<think>` tags from response content when using fallback
- **overhead**: negligible (~0.001-0.01ms per invocation, only when needed)

```
    ======================================================================
     OVERHEAD BREAKDOWN (per LLM call)
     ======================================================================

     1. FAST PATH (no <think> tags - 99% of models):
        0.028 µs  (0.0000 ms)
        └─ Just 2 string containment checks

     2. SLOW PATH (models with <think> tags - buggy NVIDIA models):
        String check:      0.070 µs
        Regex search:      0.866 µs
        Regex substitute:  0.818 µs
        String strip:      ~0.050 µs
        ────────────────────────────
        TOTAL:            1.936 µs  (0.000 ms)

     ======================================================================
     CONTEXT: TYPICAL LLM CALL LATENCY
     ======================================================================

     Network latency (API call)...................    50.00 ms  ( 6.67%)
     LLM inference (forward pass).................   500.00 ms  (66.66%)
     Token generation.............................   200.00 ms  (26.66%)
     Response parsing.............................     0.10 ms  ( 0.01%)
     Our overhead (worst case)....................     0.00 ms  ( 0.00%)
     ───────────────────────────────────────────── 
     TOTAL........................................   750.10 ms  (100.00%)

```

### DIFF
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -238,15 +238,64 @@ def _convert_messages_to_langchain_format(prompt: List[dict]) -> List:
 
 
 def _store_reasoning_traces(response) -> None:
+    """Store reasoning traces from response in context variable.
+
+    Extracts reasoning content from response.additional_kwargs["reasoning_content"]
+    if available. Otherwise, falls back to extracting from <think> tags in the
+    response content (and removes the tags from content).
+
+    Args:
+        response: The LLM response object
+    """
+
+    reasoning_content = _extract_reasoning_content(response)
+
+    if not reasoning_content:
+        reasoning_content = _extract_and_remove_think_tags(response)
+
+    if reasoning_content:
+        reasoning_trace_var.set(reasoning_content)
+
+
+def _extract_reasoning_content(response):
     if hasattr(response, "additional_kwargs"):
         additional_kwargs = response.additional_kwargs
         if (
             isinstance(additional_kwargs, dict)
             and "reasoning_content" in additional_kwargs
         ):
-            reasoning_content = additional_kwargs["reasoning_content"]
-            if reasoning_content:
-                reasoning_trace_var.set(reasoning_content)
+            return additional_kwargs["reasoning_content"]
+    return None
+
+
+def _extract_and_remove_think_tags(response) -> Optional[str]:
+    """Extract reasoning from <think> tags and remove them from `response.content`.
+
+    This function looks for <think>...</think> tags in the response content,
+    and if found, extracts the reasoning content inside the tags, it has a side-effect
+    it removes the <think> tags and their content from response.content.
+
+    Args:
+        response: The LLM response object
+
+    Returns:
+        The extracted reasoning content, or None if no <think> tags found
+    """
+    if not hasattr(response, "content"):
+        return None
+
+    content = response.content
+    if "<think>" not in content or "</think>" not in content:
+        return None
+
+    match = re.search(r"<think>(.*?)</think>", content, re.DOTALL)
+    if match:
+        reasoning_content = match.group(1).strip()
+        response.content = re.sub(
+            r"<think>.*?</think>", "", content, flags=re.DOTALL
+        ).strip()
+        return reasoning_content
+    return None
 
 
 def _store_tool_calls(response) -> None:
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -13,8 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import re
 from typing import Any, Dict, List, Optional, Sequence, Union
+
+logger = logging.getLogger(__name__)
 
 from langchain.base_language import BaseLanguageModel
 from langchain.callbacks.base import AsyncCallbackHandler, BaseCallbackManager
@@ -251,6 +254,9 @@ def _store_reasoning_traces(response) -> None:
     reasoning_content = _extract_reasoning_content(response)
 
     if not reasoning_content:
+        # Some LLM providers (e.g., certain NVIDIA models) embed reasoning in <think> tags
+        # instead of properly populating reasoning_content in additional_kwargs, so we need
+        # both extraction methods to support different provider implementations.
         reasoning_content = _extract_and_remove_think_tags(response)
 
     if reasoning_content:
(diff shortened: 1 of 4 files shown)


tests/test_reasoning_trace_extraction.py
```diff
@@ -304,3 +304,94 @@ async def test_reasoning_content_with_other_additional_kwargs(self):
...
+            additional_kwargs={"reasoning_content": reasoning_from_kwargs},
+        )
+        mock_llm.ainvoke = AsyncMock(return_value=mock_response)
+
+        from nemoguardrails.actions.llm.utils import llm_call
+
+        reasoning_trace_var.set(None)
+        result = await llm_call(mock_llm, "Query")
+
+        assert result == f"<think>{reasoning_from_tags}</think>Response"
```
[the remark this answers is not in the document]
Pouyanpi: it is already removed and content field is free from <think> tag

nemoguardrails/actions/llm/utils.py
```diff
@@ -238,15 +238,64 @@ def _convert_messages_to_langchain_format(prompt: List[dict]) -> List:
...
+    content = response.content
+    if "<think>" not in content or "</think>" not in content:
+        return None
+
+    match = re.search(r"<think>(.*?)</think>", content, re.DOTALL)
+    if match:
+        reasoning_content = match.group(1).strip()
+        response.content = re.sub(
+            r"<think>.*?</think>", "", content, flags=re.DOTALL
+        ).strip()
```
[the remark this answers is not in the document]
Pouyanpi: nested tags are not expected

nemoguardrails/actions/llm/utils.py
```diff
@@ -238,15 +238,64 @@ def _convert_messages_to_langchain_format(prompt: List[dict]) -> List:
...
+
+    Returns:
+        The extracted reasoning content, or None if no <think> tags found
+    """
+    if not hasattr(response, "content"):
+        return None
+
+    content = response.content
+    if "<think>" not in content or "</think>" not in content:
+        return None
```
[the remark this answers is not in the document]
Pouyanpi: we won't have nested tags, would you update your score? this is Ok like this

nemoguardrails/actions/llm/utils.py
```diff
@@ -238,15 +238,64 @@ def _convert_messages_to_langchain_format(prompt: List[dict]) -> List:
...
+
+    Extracts reasoning content from response.additional_kwargs["reasoning_content"]
+    if available. Otherwise, falls back to extracting from <think> tags in the
+    response content (and removes the tags from content).
+
+    Args:
+        response: The LLM response object
+    """
+
+    reasoning_content = _extract_reasoning_content(response)
```
trebedea, line 254: I would add a short comment why we need both methods to extract reasoning traces.

nemoguardrails/actions/llm/utils.py
```diff
@@ -238,15 +238,64 @@ def _convert_messages_to_langchain_format(prompt: List[dict]) -> List:
...
+            return additional_kwargs["reasoning_content"]
+    return None
+
+
+def _extract_and_remove_think_tags(response) -> Optional[str]:
+    """Extract reasoning from <think> tags and remove them from `response.content`.
+
+    This function looks for <think>...</think> tags in the response content,
+    and if found, extracts the reasoning content inside the tags, it has a side-effect
+    it removes the <think> tags and their content from response.content.
```
trebedea:
```suggestion
    it removes the full reasoning trace and tags from response.content.
```

nemoguardrails/actions/llm/utils.py
```diff
@@ -238,15 +238,64 @@ def _convert_messages_to_langchain_format(prompt: List[dict]) -> List:
...
+        response: The LLM response object
+
+    Returns:
+        The extracted reasoning content, or None if no <think> tags found
+    """
+    if not hasattr(response, "content"):
+        return None
+
+    content = response.content
+    if "<think>" not in content or "</think>" not in content:
```
trebedea: Should we at least output a warning then we have a `<think>` tag and no enclosing one? 

We could even strip all the tokens (as they are reasoning tokens) and have an empty response in this case - this would most probably fail the underlying app but on my side it is cleaner. Otherwise, the app would process the incomplete reasoning traces and would provide an incorrect behavior, e.g. an incomplete reasoning trace for input rails instead of just "safe/unsafe".

trebedea (approved):
Looks good, some comments.

nemoguardrails/actions/llm/utils.py
Pouyanpi: sure 👍🏻

nemoguardrails/actions/llm/utils.py
Pouyanpi: Great point, I'll push the change 👍🏻

nemoguardrails/actions/llm/utils.py
```diff
@@ -238,15 +241,78 @@ def _convert_messages_to_langchain_format(prompt: List[dict]) -> List:
...
+    if not has_opening_tag and not has_closing_tag:
+        return None
+
+    if has_opening_tag != has_closing_tag:
+        logger.warning(
+            "Malformed <think> tags detected: missing %s tag. "
+            "Skipping reasoning extraction to prevent corrupted content.",
+            "closing" if has_opening_tag else "opening",
+        )
+        return None
```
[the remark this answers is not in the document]
Pouyanpi: the first guard condition (if not has_opening_tag and not has_closing_tag) catches the case where both tags are missing and returns early before the XOR check is evaluated. Therefore, the XOR check only handles cases where exactly one tag is present, making the warning message correct.

PR #1487, opened 2025-10-31
fix(runtime): set stop flag for exception-based rails in parallel mode

Pouyanpi:
## Description

Fixes the `stop` flag not being set to `True` for exception-based rails when running in parallel mode. Previously, when `parallel: True` and `enable_rails_exceptions: True`, rails that raised exceptions (like `ContentSafetyCheckInputException`) had `stop=False` instead of `stop=True`, creating inconsistency with sequential mode.

### Root Cause

The parallel execution code only checked for `BotIntent` with `intent == "stop"` to determine if a rail stopped. Exception events were not recognized as stop conditions, causing:
- `InputRailFinished` post-event to be added even when rail raised exception
- `stop` flag to remain `False`
- Inconsistent behavior between parallel and sequential modes

### DIFF
--- a/nemoguardrails/colang/v1_0/runtime/runtime.py
+++ b/nemoguardrails/colang/v1_0/runtime/runtime.py
@@ -313,7 +313,8 @@ class RuntimeV1_0(Runtime):
             result = await func(*args, **kwargs)
 
             has_stop = any(
-                event["type"] == "BotIntent" and event["intent"] == "stop"
+                (event["type"] == "BotIntent" and event["intent"] == "stop")
+                or event["type"].endswith("Exception")
                 for event in result
             )
 
@@ -377,7 +378,8 @@ class RuntimeV1_0(Runtime):
 
                     # Check if this rail requested to stop
                     has_stop = any(
-                        event["type"] == "BotIntent" and event["intent"] == "stop"
+                        (event["type"] == "BotIntent" and event["intent"] == "stop")
+                        or event["type"].endswith("Exception")
                         for event in result
                     )
 
@@ -402,6 +404,7 @@ class RuntimeV1_0(Runtime):
                                     if v == pending_task:
                                         del unique_flow_ids[k]
                                         break
+                        # Remove the stopped flow from unique_flow_ids so it's not in finished_task_results
                         del unique_flow_ids[flow_id]
                         break
                     else:
@@ -448,13 +451,13 @@ class RuntimeV1_0(Runtime):
                 for plog in logs:
                     # Filter out "Listen" and "start_flow" events from task processing log
                     if plog["type"] == "event" and (
-                        plog["data"]["type"] == "Listen"
-                        or plog["data"]["type"] == "start_flow"
+                        plog["data"]["type"] == "start_flow"
                     ):
                         continue
                     target_log.append(plog)
 
-            filter_and_append(stopped_task_processing_logs, processing_log)
+            # Only append finished rails logs. Stopped rail logs should not be appended
+            # again since they're already in the processing log from when they started
             filter_and_append(finished_task_processing_logs, processing_log)
 
         # We pack all events into a single event to add it to the event history.
@@ -463,6 +466,7 @@ class RuntimeV1_0(Runtime):
             data={"events": finished_task_results},
         )
 
+        # Return stopped_task_results separately so the caller knows to stop processing
         return ActionResult(
             events=[history_events] + stopped_task_results,
             context_updates=context_updates,
--- a/nemoguardrails/colang/v1_0/runtime/runtime.py
+++ b/nemoguardrails/colang/v1_0/runtime/runtime.py
@@ -449,7 +449,6 @@ class RuntimeV1_0(Runtime):
 
             def filter_and_append(logs, target_log):
                 for plog in logs:
-                    # Filter out "Listen" and "start_flow" events from task processing log
                     if plog["type"] == "event" and (
                         plog["data"]["type"] == "start_flow"
                     ):
--- a/nemoguardrails/logging/processing_log.py
+++ b/nemoguardrails/logging/processing_log.py
@@ -191,6 +191,15 @@ def compute_generation_log(processing_log: List[dict]) -> GenerationLog:
             elif event_type == "OutputRailsFinished":
                 output_rails_finished_at = event["timestamp"]
 
+            elif event_type.endswith("Exception"):
+                if activated_rail is not None and activated_rail.type in [
+                    "input",
+                    "output",
+                ]:
+                    activated_rail.stop = True
+                    if "stop" not in activated_rail.decisions:
+                        activated_rail.decisions.append("stop")
+
         elif event["type"] == "llm_call_info":
             if executed_action is not None:
                 executed_action.llm_calls.append(event["data"])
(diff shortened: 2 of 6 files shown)


nemoguardrails/colang/v1_0/runtime/runtime.py
```diff
@@ -448,13 +451,13 @@ def filter_and_append(logs, target_log):
                 for plog in logs:
                     # Filter out "Listen" and "start_flow" events from task processing log
```
tgasser-nv: nit: please update the comment now we don't filter "Listen" data

nemoguardrails/colang/v1_0/runtime/runtime.py
```diff
@@ -313,7 +313,8 @@ async def task_call_helper(flow_uid, post_event, func, *args, **kwargs):
             result = await func(*args, **kwargs)
 
             has_stop = any(
-                event["type"] == "BotIntent" and event["intent"] == "stop"
+                (event["type"] == "BotIntent" and event["intent"] == "stop")
+                or event["type"].endswith("Exception")
```
tgasser-nv, line 317: Few questions on the matching:
* Are there any other conditions that can stop us returning the response to the client?
* Also should we do any normalization (strip trailing whitespace or convert to lower-case for example)?

nemoguardrails/colang/v1_0/runtime/runtime.py
```diff
@@ -402,6 +404,7 @@ async def task_call_helper(flow_uid, post_event, func, *args, **kwargs):
                                     if v == pending_task:
                                         del unique_flow_ids[k]
                                         break
+                        # Remove the stopped flow from unique_flow_ids so it's not in finished_task_results
```
tgasser-nv, line 407: Do we need to add a lock here (line 405 (`del unique_flow_ids[k]`)) or is it thread-safe? Couple of possibilities

* Is this the only place we delete an item in `unique_flow_ids`? 
* Are there any if statements that check the `unique_flow_ids` contents and then access one of them which could have been deleted in-between?

nemoguardrails/logging/processing_log.py
```diff
@@ -191,6 +191,15 @@ def compute_generation_log(processing_log: List[dict]) -> GenerationLog:
             elif event_type == "OutputRailsFinished":
                 output_rails_finished_at = event["timestamp"]
 
+            elif event_type.endswith("Exception"):
```
tgasser-nv, line 194: We're doing the same check in three places, maybe worth refactoring into a helper?

tests/test_parallel_rails_exceptions.py
```diff
@@ -0,0 +1,432 @@
...
+@pytest.mark.asyncio
+async def test_parallel_rails_exception_consistency_with_sequential():
+    """Test that parallel mode with exceptions behaves like sequential mode.
+
+    This test verifies that the fix makes parallel and sequential modes consistent.
+    """
+    config_parallel = RailsConfig.from_path(
+        os.path.join(CONFIGS_FOLDER, "parallel_rails_with_exceptions")
+    )
+    config_parallel.rails.input.parallel = True
```
tgasser-nv: Isn't this already set in the config?

tests/test_parallel_rails_exceptions.py
```diff
@@ -0,0 +1,432 @@
...
+    )
+    chat_parallel >> "This message has an unsafe word"
+    result_parallel = await chat_parallel.app.generate_async(
+        messages=chat_parallel.history, options=OPTIONS
+    )
+
+    config_sequential = RailsConfig.from_path(
+        os.path.join(CONFIGS_FOLDER, "parallel_rails_with_exceptions")
+    )
+    config_sequential.rails.input.parallel = False
```
tgasser-nv, line 166: Do you need to set the output flows to False also (`config_sequential.rails.output.parallel = False`)?

tests/test_parallel_rails_exceptions.py
```diff
@@ -0,0 +1,432 @@
...
+    )
+    chat_sequential >> "This message has an unsafe word"
+    result_sequential = await chat_sequential.app.generate_async(
+        messages=chat_sequential.history, options=OPTIONS
+    )
+
+    parallel_stopped = [r for r in result_parallel.log.activated_rails if r.stop]
+    sequential_stopped = [r for r in result_sequential.log.activated_rails if r.stop]
+
+    assert len(parallel_stopped) > 0, "Parallel mode should have stopped rails"
```
tgasser-nv, line 181: Can we make a stronger check here that the number of stopped rails are equal in parallel and sequential? Or do we short-circuit if a parallel-rail fails?

tests/test_parallel_rails_exceptions.py
```diff
@@ -0,0 +1,432 @@
...
+    )
+
+    # this message triggers safety check but not topic check
+    chat >> "This message is unsafe but on topic"
+    result = await chat.app.generate_async(messages=chat.history, options=OPTIONS)
+
+    stopped_rails = [r for r in result.log.activated_rails if r.stop]
+    passed_rails = [r for r in result.log.activated_rails if not r.stop]
+
+    assert len(stopped_rails) > 0, "Should have at least one stopped rail"
```
tgasser-nv, line 219: Should we check the `check_safety_action` fired but not any of the others here to be more specific?

tests/test_parallel_rails_exceptions.py
```diff
@@ -0,0 +1,432 @@
...
+        llm_completions=["Hi there!"],
+    )
+
+    # this input violates BOTH safety (has "unsafe") AND topic (has "stupid")
+    chat >> "This is an unsafe and stupid message"
+    result = await chat.app.generate_async(messages=chat.history, options=OPTIONS)
+
+    stopped_rails = [r for r in result.log.activated_rails if r.stop]
+
+    assert (
```
tgasser-nv, line 267: Same as above, we know which rails should trigger. Why not check for them specifically and not make a weaker check?

tests/test_parallel_rails_exceptions.py
```diff
@@ -0,0 +1,432 @@
...
+
+    chat >> "Hello"
+    result = await chat.app.generate_async(messages=chat.history, options=OPTIONS)
+
+    output_rails = [r for r in result.log.activated_rails if r.type == "output"]
+    assert len(output_rails) > 0, "Expected at least one output rail to be activated"
+
+    stopped_output_rails = [r for r in output_rails if r.stop]
+
+    assert (
```
tgasser-nv, line 316: Same as above, should we check which actions fired?

tgasser-nv (approved):
Looks good, left some feedback to address before merging. Can you run a simple local integration test to reproduce the issue in our previous release, and then check it works after this commit?

Similarly can you work with Zhenming to add a test in QA which fails for the previous release and passes with an RC including this commit?

nemoguardrails/colang/v1_0/runtime/runtime.py
Pouyanpi: No these are the only ones

nemoguardrails/colang/v1_0/runtime/runtime.py
Pouyanpi: in asyncio, code between await points runs atomically wrt  other tasks in the same event loop. No lock is needed for shared data accessed only within the event loop

PR #1472, opened 2025-10-24
feat: support langchain v1

Pouyanpi:
# PR Description

## Summary

Adds full LangChain 1.x support while maintaining backward compatibility with 0.x versions. This PR migrates all imports to use canonical `langchain-core` paths, removes deprecated features (Chain support, LLMParams, SummarizeDocument), and provides documentation updates.

BREAKING CHANGES:
- Chain support removed from action dispatcher (use Runnable instead)
- SummarizeDocument built-in action removed (implement custom version)
- langchain-nvidia-ai-endpoints removed from optional dependencies

### Dependency Updates
- Extended langchain constraint from `<0.4.0` to `<2.0.0`
- Extended langchain-core constraint from `<0.4.0` to `<2.0.0`
- Extended langchain-community constraint from `<0.4.0` to `<2.0.0`

### Import Standardization (28 files)
Migrated all imports to canonical paths:
```python
# before (deprecated)
from langchain.chat_models.base import BaseChatModel
from langchain_core.language_models.llms import BaseLLM

# after (correct)
from langchain_core.language_models import BaseChatModel, BaseLLM
```

### Removed Deprecated Features
- **Chain support** : Removed from action_dispatcher.py (use Runnable instead)
- **LLMParams** : Removed context manager (use native LLM params)
- **SummarizeDocument** : Removed built-in action (implement custom)
- **langchain-nvidia-ai-endpoints** : Removed from optional deps (install manually untill it supports v1)

### Documentation Updates
- Completely rewrote custom LLM provider guide
- Added BaseChatModel documentation (was missing!)
- Added migration guide for Chain → Runnable
- [ ] update CHANGELOG.md with breaking changes

### Backward Compatibility
Examples include fallback pattern for LangChain 1.x users:
```python
try:
    from langchain.chains import RetrievalQA
except ImportError:
    from langchain_classic.chains import RetrievalQA
```

PR #1496, opened 2025-11-06
feat(llm): add LangChain 1.x content blocks support for reasoning and tool calls

Pouyanpi:
## Description

Added support for extracting reasoning traces and tool calls from LangChain 1.x content blocks format, with automatic fallback to the legacy format for backward compatibility.

The implementation:
- extracts reasoning from `content_blocks` with `type: "reasoning"`
- extracts tool calls from `content_blocks` with `type: "tool_call"`
- falls back to `additional_kwargs` and `tool_calls` attribute for providers using the legacy format

references: 

- https://docs.langchain.com/oss/python/migrate/langchain-v1#standard-content
- https://docs.langchain.com/oss/python/langchain/messages#standard-content-blocks

requires:

#1472

### DIFF
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -243,15 +243,18 @@ def _convert_messages_to_langchain_format(prompt: List[dict]) -> List:
 def _store_reasoning_traces(response) -> None:
     """Store reasoning traces from response in context variable.
 
-    Extracts reasoning content from response.additional_kwargs["reasoning_content"]
-    if available. Otherwise, falls back to extracting from <think> tags in the
-    response content (and removes the tags from content).
+    Tries multiple extraction methods in order of preference:
+    1. content_blocks with type="reasoning" (LangChain v1 standard)
+    2. additional_kwargs["reasoning_content"] (provider-specific)
+    3. <think> tags in content (legacy fallback)
 
     Args:
         response: The LLM response object
     """
+    reasoning_content = _extract_reasoning_from_content_blocks(response)
 
-    reasoning_content = _extract_reasoning_content(response)
+    if not reasoning_content:
+        reasoning_content = _extract_reasoning_from_additional_kwargs(response)
 
     if not reasoning_content:
         # Some LLM providers (e.g., certain NVIDIA models) embed reasoning in <think> tags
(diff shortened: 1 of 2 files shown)


nemoguardrails/actions/llm/utils.py
```diff
@@ -238,23 +238,52 @@ def _convert_messages_to_langchain_format(prompt: List[dict]) -> List:
...
+
+    if reasoning_content:
+        reasoning_trace_var.set(reasoning_content)
+
+
+def _extract_reasoning_from_content_blocks(response) -> str | None:
+    if hasattr(response, "content_blocks"):
+        for block in response.content_blocks:
+            if block.get("type") == "reasoning":
+                return block["reasoning"]
```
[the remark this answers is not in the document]
Pouyanpi: this must be guaranteed by downstream (langchain content block) interface, no? otherwise we silently ignore such bug in downstream package.

tests/test_actions_llm_utils.py
```diff
@@ -123,3 +134,452 @@ class Wrapper3(Wrapper2):
...
+
+def test_extract_reasoning_from_additional_kwargs_not_dict():
+    response = MockResponse(additional_kwargs="not a dict")
+    reasoning = _extract_reasoning_from_additional_kwargs(response)
+    assert reasoning is None
+
+
+def test_extract_tool_calls_from_content_blocks_single_tool_call():
+    response = MockResponse(
+        content_blocks=[
```
tgasser-nv: nit: Pull out the dict into a single variable and put it into the `content_blocks` and assert it matches below?

tests/test_actions_llm_utils.py
```diff
@@ -123,3 +134,452 @@ class Wrapper3(Wrapper2):
...
+        ]
+    )
+    _store_reasoning_traces(response)
+
+    reasoning = reasoning_trace_var.get()
+    assert reasoning is None
+
+
+def test_store_tool_calls_from_content_blocks():
+    tool_calls_var.set(None)
```
tgasser-nv: There are ~13 tests which use the `tool_calls_var` and set it to None at the start and end of the tests. Consider using a pytest fixture like this:

```
@pytest.fixture(autouse=True)
def reset_tool_calls_var():
    """
    Automatically reset tool_calls_var before and after every test.
    """
    # token has the previous state
    token = tool_calls_var.set(None)
    
    # This is where the test itself will run
    yield
    
    # We use the token to reset the var to its previous state
    tool_calls_var.reset(token)
```

tgasser-nv (approved):
Looks good, please address the comments on the tests before merging. These auto-generated tests tend to bloat the line-count of tests by checking individual contents of dicts. 

These tests basically store and get the same data back in each test via different permutations. Recommend re-writing test to define a dict which is both stored and checked in the assert

PR #1502, opened 2025-11-17
feat: propagate model and base URL in LLMCallException; improve error handling

JashG:
## Description
An area of improvement we'd like to address in the [NeMo Guardrails microservice](https://docs.nvidia.com/nemo/microservices/latest/guardrails/index.html) is error messages. Specifically, they tend to lack enough context and are not actionable.

This PR contains a few changes around error handling:
- Propagate relevant context in error messages. Specifically, since a config can invoke multiple models hosted at different URLs, it'd be helpful for the error message to include the model, base URL, and most relevant error.
- Add a few custom exceptions to enable consumers to better handle specific errors.

## Checklist

### DIFF
--- a/nemoguardrails/llm/models/langchain_initializer.py
+++ b/nemoguardrails/llm/models/langchain_initializer.py
@@ -168,10 +168,10 @@ def init_langchain_model(
             if first_import_error is None:
                 first_import_error = e
             last_exception = e
-            log.error(f"Initialization import‐failure in {initializer}: {e}")
+            log.debug(f"Initialization import‐failure in {initializer}: {e}")
         except Exception as e:
             last_exception = e
-            log.error(f"Initialization failed with {initializer}: {e}")
+            log.debug(f"Initialization failed with {initializer}: {e}")
     # build the final message, preferring that first ImportError if we saw one
     base = f"Failed to initialize model {model_name!r} with provider {provider_name!r} in {mode!r} mode"
 
@@ -222,9 +222,6 @@ def _init_chat_completion_model(model_name: str, provider_name: str, kwargs: Dic
             **kwargs,
         )
     except ValueError:
-        raise
-    except Exception as e:
-        log.error(e, exc_info=True)
         raise
 
 
--- a/nemoguardrails/llm/models/langchain_initializer.py
+++ b/nemoguardrails/llm/models/langchain_initializer.py
@@ -225,9 +225,7 @@ def _init_chat_completion_model(model_name: str, provider_name: str, kwargs: Dic
         raise
 
 
-def _init_text_completion_model(
-    model_name: str, provider_name: str, kwargs: Dict[str, Any]
-) -> BaseLLM:
+def _init_text_completion_model(model_name: str, provider_name: str, kwargs: Dict[str, Any]) -> BaseLLM:
     """Initialize a text completion model.
 
     Args:
@@ -251,9 +249,7 @@ def _init_text_completion_model(
     return provider_cls(**kwargs)
 
 
-def _init_community_chat_models(
-    model_name: str, provider_name: str, kwargs: Dict[str, Any]
-) -> BaseChatModel:
+def _init_community_chat_models(model_name: str, provider_name: str, kwargs: Dict[str, Any]) -> BaseChatModel:
     """Initialize community chat models.
 
     Args:
@@ -275,9 +271,7 @@ def _init_community_chat_models(
     return provider_cls(**kwargs)
 
 
-def _init_gpt35_turbo_instruct(
-    model_name: str, provider_name: str, kwargs: Dict[str, Any]
-) -> BaseLLM:
+def _init_gpt35_turbo_instruct(model_name: str, provider_name: str, kwargs: Dict[str, Any]) -> BaseLLM:
     """Initialize GPT-3.5 Turbo Instruct model.
 
     Currently init_chat_model from langchain infers this as a chat model.
--- a/nemoguardrails/llm/models/langchain_initializer.py
+++ b/nemoguardrails/llm/models/langchain_initializer.py
@@ -227,7 +227,7 @@ def _init_chat_completion_model(model_name: str, provider_name: str, kwargs: Dic
 
 def _init_text_completion_model(
     model_name: str, provider_name: str, kwargs: Dict[str, Any]
-) -> BaseLLM | None:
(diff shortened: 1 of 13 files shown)


nemoguardrails/llm/models/langchain_initializer.py
```diff
@@ -142,13 +142,13 @@ def init_langchain_model(
...
+        # FIXME: is text and chat a good idea?
+        # For text mode, use text completion, we are using both text and chat as the last resort
+        ModelInitializer(_init_text_completion_model, ["text", "chat"]),
         # For chat mode, first try the standard chat completion API
         ModelInitializer(_init_chat_completion_model, ["chat"]),
         # For chat mode, fall back to community chat models
         ModelInitializer(_init_community_chat_models, ["chat"]),
-        # FIXME: is text and chat a good idea?
-        # For text mode, use text completion, we are using both text and chat as the last resort
-        ModelInitializer(_init_text_completion_model, ["text", "chat"]),
```
JashG, line 147: The NeMo Guardrails MS uses a custom chat client for `main` models. We often face this generic error:
```
nemoguardrails.llm.models.langchain_initializer.ModelInitializationError: Failed to initialize model 'meta/llama-3.3-70b-instruct' with provider 'nimchat' in 'chat' mode: Could not find LLM provider 'nimchat'
```

However, the relevant exception was thrown by our custom class, but gets swallowed here by the exception thrown by `_init_text_completion_model`.

I repositioned `_init_text_completion_model` to come before `ModelInitializer(_init_community_chat_models, ["chat"])`, to ensure the relevant exception is bubbled up. Let me know if there are any issues with this.

Pouyanpi: Alright, I see the issue 👍🏻  And yes the re ordering causes issues. How about?

```python
initializers: list[ModelInitializer] = [
    ModelInitializer(_handle_model_special_cases, ["chat", "text"]), 
    ModelInitializer(_init_chat_completion_model, ["chat"]),
    ModelInitializer(_init_community_chat_models, ["chat"]), 
    ModelInitializer(_init_text_completion_model, ["text"]), # change to text mode only
]
```

JashG: That causes some test failures: https://github.com/NVIDIA-NeMo/Guardrails/actions/runs/19683828515/job/56384278450?pr=1502

I'm not entirely sure why. Looking further

JashG: @Pouyanpi Just to better understand, what's the issue with re-ordering?

JashG: Ok, a solution that seems to work: I updated `_init_text_completion_model` and `_init_community_chat_models` to return `None` if a provider isn't found, instead of throwing an exception. This way, the exception thrown by the provider that was actually invoked should be the last exception that gets bubbled up.

Pouyanpi: the order is critical because:

  1. init_chat_model (from langchain's standard API) supports only ~20 major providers like openai, anthropic, cohere, etc., and is actively maintained.
  2. _init_community_chat_models supports ~60+ providers from langchain_community, including legacy/niche providers like baichuan, maritalk, yuan2, etc., but these are older and less maintained.
  3. overlap exists: providers like openai, anthropic, cohere appear in BOTH. We want init_chat_model to handle them
  4. moving _init_text_completion_model earlier breaks this preference, it becomes a catch all that prevents
  init_chat_model from running for chat mode, forcing all models through either text completion or community
  implementations instead of the preferred standard path.

  The original order ensures: special cases → modern standard API → legacy community → desperate fallback.
  
 **and I like very much the changes you made to the exceptions and messages in this PR, I'd suggest that you move the initializer related to changes and open a separate PR. It also makes the scope of each PR cleaner and does not unnecessarily block this. Would it be OK with you?**
  
  as a side note:
  

  
Why _init_text_completion_model exists:

this is due to text completion and chat completion API types. llm vs chat models in langchain. Similar to what we have in MS.

Users can explicitly set mode: "text" in their config.yml when they need models that only support text completion
APIs. The default is "chat", but text mode is necessary for certain models.

LangChain has 95+ text completion providers versus ~60 chat providers, so some legacy or niche providers only have
text completion implementations available. Special cases like gpt-3.5-turbo-instruct require text completion handling because LangChain incorrectly identifies
it as a chat model when it's actually text-only.

in practice, most usage is chat mode (the default). Text mode exists for legacy models, explicit user configuration,
and as a safety net.

JashG: Got it, this definitely makes sense - thanks for the detailed explanation.

Sure, I removed the initializer changes and will open a separate PR to unblock moving this one forward.

nemoguardrails/llm/models/langchain_initializer.py
```diff
@@ -142,13 +142,13 @@ def init_langchain_model(
     initializers: list[ModelInitializer] = [
         # Try special case handlers first (handles both chat and text)
         ModelInitializer(_handle_model_special_cases, ["chat", "text"]),
+        # FIXME: is text and chat a good idea?
+        # For text mode, use text completion, we are using both text and chat as the last resort
+        ModelInitializer(_init_text_completion_model, ["text", "chat"]),
```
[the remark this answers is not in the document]
JashG: Reverted changes to this file.

nemoguardrails/llm/models/langchain_initializer.py
```diff
@@ -142,13 +142,13 @@ def init_langchain_model(
...
         # For chat mode, first try the standard chat completion API
         ModelInitializer(_init_chat_completion_model, ["chat"]),
         # For chat mode, fall back to community chat models
         ModelInitializer(_init_community_chat_models, ["chat"]),
```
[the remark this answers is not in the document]
JashG: Reverted changes to this file.

nemoguardrails/llm/models/langchain_initializer.py
JashG: Separate PR for initializer changes: https://github.com/NVIDIA-NeMo/Guardrails/pull/1516

Pouyanpi (approved):
Thank you @JashG . Looks great, the changes are super helpful.

Let's merge when you are ready 👍🏻

PR #1504, opened 2025-11-18
fix(llm): Add async streaming support to ChatNVIDIA provider patch

Pouyanpi:
## Description

Adds async streaming support to the patched ChatNVIDIA provider, enabling `stream_async()` to properly stream with NIM models. Prior to this fix, `stream_async()` with the NIM engine would block for the complete generation and yield all content at once, rather than streaming incrementally.

## Problem

Before this fix, async streaming tests would **fail** because chunks weren't streamed incrementally:

```python
async def test_ttft(engine, model):
    config = RailsConfig.from_content(yaml_content=f"""
models:
  - type: main
    engine: nim
    model: {model}
streaming: True
""")
    rails = LLMRails(config)
    chunk_times = [time.time()]
    async for chunk in rails.stream_async(messages=[...]):
        chunk_times.append(time.time())

    ttft = chunk_times[1] - chunk_times[0]  # Time to first token
    total_time = chunk_times[-1] - chunk_times[0]

    # this assertion fails without the fix:
    assert ttft < (total_time / 2), "TTFT should be less than half of total time"
```

for example:
```python
await test_ttft(model="meta/llama-3.3-70b-instruct", engine="nim")
 ```
 
 Tested it on all the supported versions:
 
 ```markdown
   Version         TTFT (s)     Total Time (s)
  ------------------------------------------
  0.3.1           0.473        2.875
  0.3.2           0.498        2.797
  0.3.3           0.476        2.745
  0.3.4           0.450        2.507
  0.3.5           0.489        2.760
  0.3.6           0.461        2.787
  0.3.7           0.444        2.886
  0.3.8           0.432        2.831
  0.3.9           0.479        2.794
  0.3.10          0.440        2.426
  0.3.11          0.433        2.909
  0.3.12          0.471        2.692
  0.3.13          0.455        2.603
  0.3.14          0.474        2.825
  0.3.15          0.444        2.488
  0.3.16          0.473        2.645
  0.3.17          0.432        2.857
  0.3.18          0.473        3.438
  0.3.19          0.458        2.824
```

prior to the fix

```markdown
  Version         Error Type                Error Message
  --------------------------------------------------------------------------------
  Version         Error Type                Error Message
  --------------------------------------------------------------------------------
  0.3.0           AssertionError            TTFT (3.1635499000549316) should be l...
  0.3.1           AssertionError            TTFT (2.7374980449676514) should be l...
  0.3.2           AssertionError            TTFT (2.588848829269409) should be le...
  0.3.3           AssertionError            TTFT (2.7798001766204834) should be l...
  0.3.4           AssertionError            TTFT (2.870666980743408) should be le...
  0.3.5           AssertionError            TTFT (2.900015115737915) should be le...
  0.3.6           AssertionError            TTFT (3.027787923812866) should be le...
  0.3.7           AssertionError            TTFT (2.741342782974243) should be le...
  0.3.8           AssertionError            TTFT (2.8514771461486816) should be l...
  0.3.9           AssertionError            TTFT (3.0388290882110596) should be l...
  0.3.10          AssertionError            TTFT (3.1536900997161865) should be l...
  0.3.11          AssertionError            TTFT (2.754746913909912) should be le...
  0.3.12          AssertionError            TTFT (2.7870500087738037) should be l...
  0.3.13          AssertionError            TTFT (2.575707197189331) should be le..
  0.3.14          AssertionError            TTFT (2.9357969760894775) should be l...
  0.3.15          AssertionError            TTFT (2.972874879837036) should be le...
  0.3.16          AssertionError            TTFT (2.6601181030273438) should be l...
  0.3.17          AssertionError            TTFT (3.1566638946533203) should be l...
  0.3.18          AssertionError            TTFT (3.070240020751953) should be le...
  0.3.19          RuntimeError              list index out of range
```

in 0.3.19 async api support is added to ChatNVIDIA, that's why we are getting a different error. This is fixed too.

```
LIVE_TEST_MODE=1 poetry run pytest tests/llm_providers/test_langchain_nvidia_ai_endpoints_patch.py -v
```
ensure you have `langchain-nvidia-ai-endpoints` installed and NVIDIA_API_KEY set.

### DIFF
--- a/nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
+++ b/nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
@@ -57,7 +57,7 @@ def stream_decorator(func):  # pragma: no cover
     return wrapper
 
 
-def async_stream_decorator(func):
+def async_stream_decorator(func):  # pragma: no cover
     @wraps(func)
     async def wrapper(
         self,
--- a/nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
+++ b/nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
@@ -18,7 +18,6 @@ import logging
 from functools import wraps
 from typing import Any, Dict, List, Optional
 
-from langchain_core.callbacks import Callbacks
 from langchain_core.callbacks.manager import (
     AsyncCallbackManagerForLLMRun,
     CallbackManagerForLLMRun,
@@ -134,14 +133,12 @@ class ChatNVIDIA(ChatNVIDIAOriginal):  # pragma: no cover
         messages: List[BaseMessage],
         stop: Optional[List[str]] = None,
         run_manager: Optional[CallbackManagerForLLMRun] = None,
-        callbacks: Callbacks = None,
         **kwargs: Any,
     ) -> ChatResult:
         return super()._generate(
             messages=messages,
             stop=stop,
             run_manager=run_manager,
-            callbacks=callbacks,
             **kwargs,
         )
 
--- a/nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
+++ b/nemoguardrails/llm/providers/_langchain_nvidia_ai_endpoints_patch.py
@@ -18,8 +18,15 @@ import logging
 from functools import wraps
 from typing import Any, Dict, List, Optional
 
-from langchain_core.callbacks.manager import CallbackManagerForLLMRun
-from langchain_core.language_models.chat_models import generate_from_stream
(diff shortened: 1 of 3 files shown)


tgasser-nv (approved):
Thanks for fixing this. I'm really surprised streaming with the `nim` engine type has been broken the whole time (unless I misunderstood?). here are a couple of followup actions:

1. Before merging can you run this yourself locally (using curl or similar) to make sure the chunks stream correctly after the fix?
2. How was this not caught in QA in a streaming test up until now? We need a test in QA that fails prior to this fix, and passes after the fix. That will prevent regressions.
3. What is the guidance on `LIVE_TEST_MODE`? Should we run this as part of a test plan for each PR? I usually just run `pytest tests -q` but don't enable LIVE_TEST_MODE

PR #1516, opened 2025-11-26
fix: Surface relevant exception when initializing langchain model

JashG:
## Description
[init_langchain_model()](https://github.com/JashG/NeMo-Guardrails/blob/92162028a641a90d53a02fb394aa5c46c63a9923/nemoguardrails/llm/models/langchain_initializer.py#L138) attempts to initialize a model in a specific order. If no methods succeed, it returns the last exception thrown. This can mask the actual exception that is most relevant to the consumer.

For example, the [NeMo Guardrails MS](https://docs.nvidia.com/nemo/microservices/latest/guardrails/index.html) internally uses a custom chat client for `main` models. If there's an initialization error thrown by `_init_community_chat_models`, it always gets masked by `_init_text_completion_model`, which throws an exception when a provider can't be found.

This PR updates the `init_*` functions to instead return `None` if a provider can't be found. This way, if a previous `init_*` function threw an exception as a result of actually invoking a provider class, it gets preserved.

## Checklist

- [x] @mentions of the person or team responsible for reviewing proposed changes. @Pouyanpi

### DIFF
--- a/nemoguardrails/llm/models/langchain_initializer.py
+++ b/nemoguardrails/llm/models/langchain_initializer.py
@@ -386,6 +386,8 @@ def _handle_model_special_cases(
         return None
 
     result = initializer(model_name, provider_name, kwargs)
+    if result is None:
+        return None
     if not isinstance(result, (BaseChatModel, BaseLLM)):
         raise TypeError("Initializer returned an invalid type")
     return result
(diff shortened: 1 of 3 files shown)


Pouyanpi (approved):
Thank you @JashG for fixing this critical bug! Looks great! I'll follow up with tests in a separate PR.

PR #1526, opened 2025-12-01
fix(llm): filter temperature parameter for OpenAI reasoning models

### LINKED ISSUE #992: bug: calling o1 or o3 causes NeMo to use deprecated chat completion API

Python 3.10.11

### Operating system/version

Microsoft Windows 11 Enterprise

0.11.1

### Describe the bug

Trying to use Azure OpenAI o3-mini (or any o1 model) causes a bad request error.

### Steps To Reproduce

1. Put this model into config.yml

```
type: main
    engine: azure
    model: o3-mini
    api_version: 2024-12-01-preview
    ...
```

2. Start a chat with
`nemoguardrails chat --config your/path --verbose --debug-level=DEBUG`

3. Ask anything that uses the main model.

### Expected Behavior

Receive a chat completion without error.

### Actual Behavior

```
HTTPStatusError: Client error '400 Bad Request' for url
                             'https://...openai.azure.com/openai/.../o3-mini/completions?api-version=...'
```

Stack trace ends with
`nemoguardrails.actions.llm.utils.LLMCallException: LLM Call Exception: Error code: 400 - {'error': {'code': 'OperationNotSupported', 'message': 'The completion operation does not work with the specified model, o3-mini. Please choose different model and try again. You can learn more about which models can be used with each operation here: https://go.microsoft.com/fwlink/?linkid=2197993.'}}`

### LINKED ISSUE #1512: bug: BadRequestError as gpt-5 only supports temperature=1
### Did you check the docs?

Hi I am trying to use the latest model for the guardrail I assume better LLM means better performance.
I am trying to use the `openai` engine with `gpt5-mini` 
```
DEFAULT_CONFIG_DICT = {"models": [{"type": "main", "engine": "openai", "model": "gpt-5-mini"}]}
```

but I keep getting error:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Unsupported value:                  
                             'temperature' does not support 0.001 with this model. Only the default (1) value is                  
                             supported.", 'type': 'invalid_request_error', 'param': 'temperature', 'code':                        
                             'unsupported_value
```

Since I cannot make it works using the built-in engine I try to use a custom engine that calling the needed model behind but the performance is really slow (in average 30 seconds), compared to using the build in engine(under 5 seconds).

Is there any way to use GPT-5 mini or any other latest model with Nemo Guardrail built-in engine?
If it is not, is the custom engine expected to be slow? How to optimize it?

Thank you.

To be able to use GPT5-mini without customization

Custom model engine

### Additional context

-

Pouyanpi:
OpenAI reasoning models (o1, o3, gpt-5 series excluding gpt-5-chat) only support `temperature=1`. When NeMo Guardrails uses `.bind(temperature=0.001)` for deterministic tasks like self-check input/output, the API returns an error:

```
Unsupported value: 'temperature' does not support 0.001 with this model.
Only the default (1) value is supported.
```

This happens because LangChain's `ChatOpenAI` handles temperature restrictions at initialization time (setting `temperature=None` for reasoning models), but `.bind()` bypasses this protection and passes the temperature directly to the API.

- Added `_filter_params_for_openai_reasoning_models()` function that:
- Detects reasoning models by name (o1*, o3*, gpt-5* excluding gpt-5-chat)
- Removes the `temperature` parameter before binding for these models
- Follows the same pattern as LangChain's `validate_temperature` validator

to repro

```python
from langchain_openai import ChatOpenAI

# LangChain handles this at init by setting temperature=None
llm = ChatOpenAI(model='gpt-5-mini', temperature=0.1)
print(llm.temperature)  # None

# but .bind() bypasses this protection
bound = llm.bind(temperature=0.001)
bound.invoke(messages)  # Error!
```
or

```python
```python
import asyncio
from langchain_openai import ChatOpenAI
from nemoguardrails.actions.llm.utils import llm_call

llm = ChatOpenAI(model="gpt-5-mini")

async def test():
    # This would fail before the fix because llm_call binds temperature=0.001
    result = await llm_call(
        llm,
        "Say hello",
        llm_params={"temperature": 0.001, "max_tokens": 100}
    )
    print(result)

asyncio.run(test())
```

Fixes: #1512 #992

### DIFF
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -138,6 +138,35 @@ def _infer_model_name(llm: BaseLanguageModel):
     return "unknown"
 
 
+def _filter_params_for_openai_reasoning_models(llm: BaseLanguageModel, llm_params: Optional[dict]) -> Optional[dict]:
+    """Filter out unsupported parameters for OpenAI reasoning models.
+
+    OpenAI reasoning models (o1, o3, gpt-5 excluding gpt-5-chat) only support
+    temperature=1. When using .bind() with other temperature values, the API
+    returns an error. This function removes the temperature parameter for these
+    models to allow the API default to apply.
+
+    See: https://github.com/langchain-ai/langchain/blob/master/libs/partners/openai/langchain_openai/chat_models/base.py
+    """
+    if not llm_params or "temperature" not in llm_params:
+        return llm_params
+
+    model_name = _infer_model_name(llm).lower()
+
+    is_openai_reasoning_model = (
+        model_name.startswith("o1")
+        or model_name.startswith("o3")
+        or (model_name.startswith("gpt-5") and "chat" not in model_name)
+    )
+
+    if is_openai_reasoning_model:
+        filtered = llm_params.copy()
+        filtered.pop("temperature", None)
+        return filtered
+
+    return llm_params
+
+
 async def llm_call(
     llm: Optional[BaseLanguageModel],
     prompt: Union[str, List[dict]],
@@ -166,7 +195,8 @@ async def llm_call(
     _setup_llm_call_info(llm, model_name, model_provider)
     all_callbacks = _prepare_callbacks(custom_callback_handlers)
 
-    generation_llm: Union[BaseLanguageModel, Runnable] = llm.bind(**llm_params) if llm_params else llm
+    filtered_params = _filter_params_for_openai_reasoning_models(llm, llm_params)
+    generation_llm: Union[BaseLanguageModel, Runnable] = llm.bind(**filtered_params) if filtered_params else llm
 
     if isinstance(prompt, str):
         response = await _invoke_with_string_prompt(generation_llm, prompt, all_callbacks, stop)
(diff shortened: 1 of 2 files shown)


nemoguardrails/actions/llm/utils.py
```diff
@@ -136,6 +136,35 @@ def _infer_model_name(llm: BaseLanguageModel):
...
+    models to allow the API default to apply.
+
+    See: https://github.com/langchain-ai/langchain/blob/master/libs/partners/openai/langchain_openai/chat_models/base.py
+    """
+    if not llm_params or "temperature" not in llm_params:
+        return llm_params
+
+    model_name = _infer_model_name(llm).lower()
+
+    is_openai_reasoning_model = (
```
tgasser-nv, line 156: Agree with the Greptile comment, is there a less brittle way to check for an OpenAI reasoning model? Maybe any model name starting with `o`

tgasser-nv (approved):
LGTM, just a small comment on how to pick up OpenAI reasoning model names

nemoguardrails/actions/llm/utils.py
Pouyanpi: I've followed the original implementation from langchain_openai:

https://github.com/langchain-ai/langchain/blob/878f033ed72d16e54783549d9fe77899708eec07/libs/partners/openai/langchain_openai/chat_models/base.py#L822-L844

Probably there is a good reason for not doing this. We can update it later when new models are introduced.

PR #1529, opened 2025-12-02
fix: ensure that stop token is not ignored if llm_params is None

Pouyanpi:
## Description
please refer to the original PR: #1524 opened by @RobGeada


Pouyanpi:
Thank you @RobGeada 👍🏻

PR #1530, opened 2025-12-02
feat(content_safety): add support to auto select multilingual refusal bot messages

Pouyanpi:
## Description

Detect user input language and return refusal messages in the same language when content safety rails block unsafe content. Supports 9 languages: English, Spanish, Chinese, German, French, Hindi, Japanese, Arabic, and Thai.

### Datasets Used

| Dataset  | Description                             | Samples | Languages         |
|----------|-----------------------------------------|---------|-------------------|
| papluca  | [language-identification ](https://huggingface.co/datasets/papluca/language-identification#additional-information)    | 40,500  | 9 (all supported) |
| nemotron | [NVIDIA Nemotron-Safety-Guard-Dataset-v3](https://huggingface.co/datasets/nvidia/Nemotron-Safety-Guard-Dataset-v3) | 336,283 | 8 (missing zh*)   |

  Chinese samples in Nemotron are all REDACTED; Chinese coverage validated via papluca dataset.

#### Prompt Length Analysis (characters)

| Dataset  | Min | Max    | Mean  | P25 | P50 | P75 | P90 | P95   | P99   |
|----------|-----|--------|-------|-----|-----|-----|-----|-------|-------|
| papluca  | 2   | 3,657  | 129.8 | 50  | 97  | 162 | 258 | 351   | 627   |
| nemotron | 1   | 20,750 | 303.5 | 51  | 111 | 331 | 625 | 1,072 | 3,004 |

Note: fast-langdetect truncates input at 80 characters by default (`max_input_length=80`), so longer prompts are effectively evaluated on their first 80 chars.

---
### Overall Accuracy comparison

| Dataset  | Samples | fast-langdetect | lingua | detect_language action |
|----------|---------|-----------------|--------|------------------------|
| papluca  | 40,500  | 99.71%          | 99.79% | 99.71%                 |
| nemotron | 336,283 | 99.35%          | 99.46% | 99.42%                 |

---
### Latency comparison (μs)

| Dataset  | fast-langdetect Avg | fast-langdetect P95 | lingua Avg | lingua P95 | Action Avg | Action P95 |
|----------|---------------------|---------------------|------------|------------|------------|------------|
| papluca  | 12.12               | 15.54               | 116.21     | 205.29     | 25.77      | 28.75      |
| nemotron | 11.53               | 15.50               | 162.59     | 377.92     | 26.25      | 28.71      |

---
### Per Language Accuracy ([fast-langdetect](https://github.com/LlmKira/fast-langdetect/tree/main))

| Language      | papluca | nemotron |
|---------------|---------|----------|
| ar (Arabic)   | 98.87%  | 99.63%   |
| de (German)   | 99.93%  | 99.39%   |
| en (English)  | 100.00% | 99.03%   |
| es (Spanish)  | 100.00% | 99.04%   |
| fr (French)   | 99.98%  | 99.25%   |
| hi (Hindi)    | 98.76%  | 99.60%   |
| ja (Japanese) | 100.00% | 99.61%   |
| th (Thai)     | 99.93%  | 99.29%   |
| zh (Chinese)  | 99.93%  | N/A      |

---
### Per-Language Accuracy ([lingua](https://github.com/pemistahl/lingua-py))

| Language      | papluca | nemotron |
|---------------|---------|----------|
| ar (Arabic)   | 99.84%  | 99.75%   |
| de (German)   | 100.00% | 99.55%   |
| en (English)  | 99.93%  | 99.00%   |
| es (Spanish)  | 99.98%  | 99.43%   |
| fr (French)   | 99.82%  | 99.35%   |
| hi (Hindi)    | 98.80%  | 99.81%   |
| ja (Japanese) | 100.00% | 99.69%   |
| th (Thai)     | 99.78%  | 99.12%   |
| zh (Chinese)  | 99.93%  | N/A      |

---
**Why fast-langdetect?**

https://github.com/LlmKira/fast-langdetect

0. MIT license and [Creative Commons Attribution-Share-Alike License 3.0](https://github.com/LlmKira/fast-langdetect/blob/main/src/fast_langdetect/resources/NOTICE.MD).
1. comparable accuracy: within 0.1-0.5% of lingua across all datasets (99.35% vs 99.46% on 336k samples)
2. 10-14x faster: average latency ~12μs vs ~140μs
3. simpler integration: single lightweight dependency
4. no cold start issues: unlike lingua which requires model building
5. [no dependency issue in future](https://github.com/LlmKira/fast-langdetect/blob/main/pyproject.toml) 

### Error analysis

Most errors occur with:
- short text (single words): insufficient context for detection
- mixed language content: text containing English within non-English context
- similar language confusion: Spanish vs Galician, Hindi vs Marathi, Arabic vs Persian

The action correctly falls back to English (en) for unsupported detected languages.

---
### Benchmark Scripts
checkout to [temp/lang-detect-benchmark](https://github.com/NVIDIA-NeMo/Guardrails/tree/temp/lang-detect-benchmark) branch

Located in eval/language_detection/:
make sure to have datasets and pandas installed:

```
poetry run pip install pandas datasets
```

```sh
# run all benchmarks
poetry run python eval/language_detection/run_benchmarks.py

# Or run individually
poetry run python eval/language_detection/benchmark.py --dataset papluca --mode action --report eval/language_detection/reports/
```

### DIFF
--- a/nemoguardrails/library/content_safety/actions.py
+++ b/nemoguardrails/library/content_safety/actions.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 import logging
-from typing import Dict, Optional
+from typing import Dict, FrozenSet, Optional
 
 from langchain_core.language_models import BaseLLM
 
@@ -220,3 +220,79 @@ async def content_safety_check_output(
         log.debug(f"Content safety output result cached for model '{model_name}'")
 
     return final_result
+
+
+SUPPORTED_LANGUAGES: FrozenSet[str] = frozenset({"en", "es", "zh", "de", "fr", "hi", "ja", "ar", "th"})
+
+DEFAULT_REFUSAL_MESSAGES: Dict[str, str] = {
+    "en": "I'm sorry, I can't respond to that.",
+    "es": "Lo siento, no puedo responder a eso.",
+    "zh": "抱歉，我无法回应。",
+    "de": "Es tut mir leid, darauf kann ich nicht antworten.",
+    "fr": "Je suis désolé, je ne peux pas répondre à cela.",
+    "hi": "मुझे खेद है, मैं इसका जवाब नहीं दे सकता।",
+    "ja": "申し訳ありませんが、それには回答できません。",
+    "ar": "عذراً، لا أستطيع الرد على ذلك.",
+    "th": "ขออภัย ฉันไม่สามารถตอบได้",
+}
+
+
+def _detect_language(text: str) -> Optional[str]:
+    try:
+        from fast_langdetect import detect
+
+        result = detect(text, k=1)
+        if result and len(result) > 0:
+            return result[0].get("lang")
+        return None
+    except ImportError:
+        log.warning("fast-langdetect not installed, skipping")
+        return None
+    except Exception as e:
+        log.warning(f"fast-langdetect detection failed: {e}")
+        return None
+
+
+def _get_refusal_message(lang: str, custom_messages: Optional[Dict[str, str]]) -> str:
+    if custom_messages and lang in custom_messages:
+        return custom_messages[lang]
+    if lang in DEFAULT_REFUSAL_MESSAGES:
+        return DEFAULT_REFUSAL_MESSAGES[lang]
+    if custom_messages and "en" in custom_messages:
+        return custom_messages["en"]
+    return DEFAULT_REFUSAL_MESSAGES["en"]
+
+
+@action()
+async def detect_language(
+    context: Optional[dict] = None,
+    config: Optional[dict] = None,
+) -> dict:
+    user_message = ""
+    if context is not None:
+        user_message = context.get("user_message", "")
+
+    custom_messages = None
+    if config is not None:
+        multilingual_config = (
+            config.rails.config.content_safety.multilingual
+            if hasattr(config, "rails")
+            and hasattr(config.rails, "config")
+            and hasattr(config.rails.config, "content_safety")
+            and hasattr(config.rails.config.content_safety, "multilingual")
+            else None
+        )
+        if multilingual_config:
+            custom_messages = multilingual_config.refusal_messages
+
+    lang = _detect_language(user_message) or "en"
+
+    if lang not in SUPPORTED_LANGUAGES:
+        lang = "en"
+
+    refusal_message = _get_refusal_message(lang, custom_messages)
+
+    return {
+        "language": lang,
+        "refusal_message": refusal_message,
+    }
--- a/nemoguardrails/library/content_safety/actions.py
+++ b/nemoguardrails/library/content_safety/actions.py
@@ -31,7 +31,6 @@ from nemoguardrails.llm.cache.utils import (
 )
 from nemoguardrails.llm.taskmanager import LLMTaskManager
 from nemoguardrails.logging.explain import LLMCallInfo
-from nemoguardrails.rails.llm.config import RailsConfig
 
 log = logging.getLogger(__name__)
 
@@ -238,22 +237,11 @@ DEFAULT_REFUSAL_MESSAGES: Dict[str, str] = {
 }
 
 
-def _detect_language(
-    text: str,
-    max_text_length: Optional[int] = None,
-    normalize_text: bool = True,
-    cache_dir: Optional[str] = None,
-) -> Optional[str]:
+def _detect_language(text: str) -> Optional[str]:
     try:
-        from fast_langdetect import LangDetectConfig, LangDetector
+        from fast_langdetect import detect
 
-        config = LangDetectConfig(
-            max_input_length=max_text_length,
-            normalize_input=normalize_text,
-            cache_dir=cache_dir,
-        )
-        detector = LangDetector(config)
-        result = detector.detect(text, k=1)
+        result = detect(text, k=1)
         if result and len(result) > 0:
             return result[0].get("lang")
         return None
@@ -278,17 +266,13 @@ def _get_refusal_message(lang: str, custom_messages: Optional[Dict[str, str]]) -
 @action()
 async def detect_language(
     context: Optional[dict] = None,
-    config: Optional[RailsConfig] = None,
+    config: Optional[dict] = None,
 ) -> dict:
     user_message = ""
     if context is not None:
         user_message = context.get("user_message", "")
 
     custom_messages = None
-    max_text_length = None
-    normalize_text = True
-    cache_dir = None
-
     if config is not None:
         multilingual_config = (
             config.rails.config.content_safety.multilingual
@@ -300,19 +284,8 @@ async def detect_language(
         )
         if multilingual_config:
             custom_messages = multilingual_config.refusal_messages
-            max_text_length = multilingual_config.max_text_length
-            normalize_text = multilingual_config.normalize_text
-            cache_dir = multilingual_config.cache_dir
 
-    lang = (
-        _detect_language(
-            user_message,
-            max_text_length=max_text_length,
-            normalize_text=normalize_text,
-            cache_dir=cache_dir,
-        )
-        or "en"
-    )
+    lang = _detect_language(user_message) or "en"
 
     if lang not in SUPPORTED_LANGUAGES:
         lang = "en"
--- a/nemoguardrails/library/content_safety/actions.py
+++ b/nemoguardrails/library/content_safety/actions.py
@@ -31,6 +31,7 @@ from nemoguardrails.llm.cache.utils import (
 )
 from nemoguardrails.llm.taskmanager import LLMTaskManager
 from nemoguardrails.logging.explain import LLMCallInfo
+from nemoguardrails.rails.llm.config import RailsConfig
 
 log = logging.getLogger(__name__)
 
@@ -237,11 +238,22 @@ DEFAULT_REFUSAL_MESSAGES: Dict[str, str] = {
 }
 
 
-def _detect_language(text: str) -> Optional[str]:
+def _detect_language(
+    text: str,
+    max_text_length: Optional[int] = None,
+    normalize_text: bool = True,
+    cache_dir: Optional[str] = None,
+) -> Optional[str]:
     try:
-        from fast_langdetect import detect
+        from fast_langdetect import LangDetectConfig, LangDetector
 
-        result = detect(text, k=1)
+        config = LangDetectConfig(
+            max_input_length=max_text_length,
+            normalize_input=normalize_text,
+            cache_dir=cache_dir,
+        )
+        detector = LangDetector(config)
+        result = detector.detect(text, k=1)
         if result and len(result) > 0:
             return result[0].get("lang")
         return None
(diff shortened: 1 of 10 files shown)


nemoguardrails/library/content_safety/actions.py
```diff
@@ -220,3 +220,79 @@ async def content_safety_check_output(
...
+    "en": "I'm sorry, I can't respond to that.",
+    "es": "Lo siento, no puedo responder a eso.",
+    "zh": "抱歉，我无法回应。",
+    "de": "Es tut mir leid, darauf kann ich nicht antworten.",
+    "fr": "Je suis désolé, je ne peux pas répondre à cela.",
+    "hi": "मुझे खेद है, मैं इसका जवाब नहीं दे सकता।",
+    "ja": "申し訳ありませんが、それには回答できません。",
+    "ar": "عذراً، لا أستطيع الرد على ذلك.",
+    "th": "ขออภัย ฉันไม่สามารถตอบได้",
+}
```
cparisien, line 237: If we later had other multilingual rails, would we be repeating this mechanism in each rail? Or just the set of supported languages per rail? I don't think we need to do it now (since we don't have other multilingual rails to test it), but we should be aware of what refactoring would be needed to move the below language detection to a shared level.

nemoguardrails/library/content_safety/actions.py
```diff
@@ -220,3 +220,79 @@ async def content_safety_check_output(
...
+
+
+def _detect_language(text: str) -> Optional[str]:
+    try:
+        from fast_langdetect import detect
+
+        result = detect(text, k=1)
```
cparisien: Does fast-langdetect ever return a full locale with dialect, like en-US versus en? I don't see it in the docs, but I do see some upper/lowercase inconsistency.

tgasser-nv:
This looks really good @Pouyanpi ! I have a few comments:

* Could you commit the evaluation scripts as well in the final PR (for reproducibility?)
* What does the "Action" column in the latency report refer to? Is this the latency end-to-end when fast-langdetect  embedded in a Guardrails action? It approximately doubles the mean and p95.
~~* Is it possible to customize refusal texts in Colang-only, or does it need a Python change?~~ Just saw this is in the RailsConfig, that's perfect.
* Could you calculate percentiles of prompt-length (ideally in tokens but characters is fine too) for each of the datasets?

Not needed in this PR, but I'm thinking of RAG prompts where we have LLM instructions, user query, and relevant context chunks are all in a flattened prompt. These prompts can be pretty long (up to 7k tokens in some cases). This isn't needed for  this PR, but I would be interested in a follow-on where we sample part of a prompt before running classification on the sample (e.g. 200 chars). This would be an optional config field. Customers would then have a knob to trade off accuracy vs latency for language detection.

nemoguardrails/library/content_safety/actions.py
Pouyanpi: Of course, we can relax this constraint later and allow users more flexibility. Once we need to support other models or other types of rails (beyond content safety) that require multilingual responses, we can:

1. Move the `detect_language` action from `library/content_safety/actions.py` to a shared location (`nemoguardrails/actions/`) making it available to all rails
2. It was possible to also introduce a Colang level abstraction like `bot refuse to respond $multilang=true`, could be done easily for Colang 2.0, but I think it is better if we don't add new Colang features for now.


I agree, for now, keeping it scoped to content safety keeps the implementation focused.

nemoguardrails/library/content_safety/actions.py
Pouyanpi: fair point, thanks for raising it. Just took a closer look: the fast-langdetect source code and fastText model behavior.

- fast-langdetect [README](https://github.com/LlmKira/fast-langdetect/tree/main?tab=readme-ov-file#language-codes--english-names) mentions BCP-47 tags like `"zh-cn"`, `"pt-br"`
- but the [fastText lid.176.bin model](https://fasttext.cc/docs/en/language-identification.html) uses simple ISO 639 codes: `zh`, `pt`, `en`, etc.
- fast-langdetect source simply strips `__label__` prefix from fastText output,  no regional mapping is applied

**validated with actual test**:
```python
>>> detect("抱歉，我无法处理该请求", k=2)
[{'lang': 'zh', 'score': 0.80}, {'lang': 'ta', 'score': 0.08}]
```
returns `"zh"`, NOT `"zh-cn"`.

So no regional variant handling needed.

Pouyanpi:
> * Could you commit the evaluation scripts as well in the final PR (for reproducibility?)

I've included them [temp/lang-detect-benchmark](https://github.com/NVIDIA-NeMo/Guardrails/tree/temp/lang-detect-benchmark) branch to make review easier. If you find it easier I will do.
But we don't intend to merge those in develop, right?

> * What does the "Action" column in the latency report refer to? Is this the latency end-to-end when fast-langdetect  embedded in a Guardrails action? It approximately doubles the mean and p95.

Yes  

>   ~* Is it possible to customize refusal texts in Colang-only, or does it need a Python change?~ Just saw this is in the RailsConfig, that's perfect.

Yes, I would like to avoid adding colang level features as much as possible

> * Could you calculate percentiles of prompt-length (ideally in tokens but characters is fine too) for each of the datasets?

Done! updated the description.

> Not needed in this PR, but I'm thinking of RAG prompts where we have LLM instructions, user query, and relevant context chunks are all in a flattened prompt. These prompts can be pretty long (up to 7k tokens in some cases). This isn't needed for this PR, but I would be interested in a follow-on where we sample part of a prompt before running classification on the sample (e.g. 200 chars). This would be an optional config field. Customers would then have a knob to trade off accuracy vs latency for language detection.

fast-langdetect already does the truncation by [default](https://github.com/LlmKira/fast-langdetect/tree/main?tab=readme-ov-file#input-handling) but indeed we can give that flexibility to the users :
- `max_input_length=80` characters (configurable)

tgasser-nv:
> > * Could you commit the evaluation scripts as well in the final PR (for reproducibility?)
> 
> I've included them [temp/lang-detect-benchmark](https://github.com/NVIDIA-NeMo/Guardrails/tree/temp/lang-detect-benchmark) branch to make review easier. If you find it easier I will do. But we don't intend to merge those in develop, right?
> 

Why wouldn't we merge them into develop? It's best practice in ML to make any results reproducible, for which we need the input datasets and scripts. The datasets are public and linked above. I'd imagine we'll have to re-run evals for new languages as they're added to the content-safety and other models. So we'll run this script periodically.

> > * What does the "Action" column in the latency report refer to? Is this the latency end-to-end when fast-langdetect  embedded in a Guardrails action? It approximately doubles the mean and p95.
> 
> Yes
> 
Was that measured at a concurrency of 1? Having a 100% overhead for each language inference is a lot higher than I'd expect. We don't need to fix it in this PR. 

> > ~* Is it possible to customize refusal texts in Colang-only, or does it need a Python change?~ Just saw this is in the RailsConfig, that's perfect.
> 
> Yes, I would like to avoid adding colang level features as much as possible
> 
**+1**

> > * Could you calculate percentiles of prompt-length (ideally in tokens but characters is fine too) for each of the datasets?
> 
> Done! updated the description.
> 
Could you check? I didn't see any length description. 

> > Not needed in this PR, but I'm thinking of RAG prompts where we have LLM instructions, user query, and relevant context chunks are all in a flattened prompt. These prompts can be pretty long (up to 7k tokens in some cases). This isn't needed for this PR, but I would be interested in a follow-on where we sample part of a prompt before running classification on the sample (e.g. 200 chars). This would be an optional config field. Customers would then have a knob to trade off accuracy vs latency for language detection.
> 
> fast-langdetect already does the truncation by [default](https://github.com/LlmKira/fast-langdetect/tree/main?tab=readme-ov-file#input-handling) but indeed we can give that flexibility to the users :
> 
> * `max_input_length=80` characters (configurable)

Could you add optional Pydantic fields for any of these values it makes sense to expose to users? Looking at the [config](https://github.com/LlmKira/fast-langdetect/blob/803a858a5b2615934b7a1f890f47b7ed74a3ea64/src/fast_langdetect/infer.py#L202-L210) I think `normalize_input`, `max_input_length`, and `model` are all fields users might care about

Pouyanpi:
Thanks @tgasser-nv! Yes you're right, here is how it looks (didn't save the change):

image

I've extended the configuration but I'm not sure if we actually need to do it (I think better to revert its [commit](https://github.com/NVIDIA-NeMo/Guardrails/pull/1530/commits/4339ac2e73e1bd30118f0cf7d6433d719f0fe8a0)).
most users will just do `multilingual.enabled: true` and use library defaults. The config options (max_text_length, normalize_text, cache_dir) are edge cases.

Do we actually need these config options exposed?

  - cache_dir -> users can set FTLANG_CACHE env var
  - normalize_text -> library default (true) is almost always correct
  - max_text_length -> library default (80) is optimized for accuracy

  Maybe the cleanest solution is: don't expose these options at all and keep _detect_language(text) simple.

  If a power user really needs custom settings, they can:
  1. Use FTLANG_CACHE env var for cache location
  2. Or we add config options later when there's a real need

the `model` field looks library specific and we are better of not including it in the config.  I found cache_dir might be interesting but we can document FTLANG_CACHE. Do you think we should keep the new config options or revert?



Regarding the evaluation scripts and datasets: let's address that in a follow-up PR to keep this one focused. 
I think it's better to keep this type of analysis work in dedicated branches for now.

We can follow up in separate PRs once we establish a clear pattern for where these should live (e.g., scripts/benchmarks/, eval/, etc.) and how they should be maintained.

tgasser-nv:
> Thanks @tgasser-nv! Yes you're right, here is how it looks (didn't save the change):
> 
> image
> I've extended the configuration but I'm not sure if we actually need to do it (I think better to revert its [commit](https://github.com/NVIDIA-NeMo/Guardrails/pull/1530/commits/4339ac2e73e1bd30118f0cf7d6433d719f0fe8a0)). most users will just do `multilingual.enabled: true` and use library defaults. The config options (max_text_length, normalize_text, cache_dir) are edge cases.
> 
> Do we actually need these config options exposed?
> 
> * cache_dir -> users can set FTLANG_CACHE env var
> * normalize_text -> library default (true) is almost always correct
> * max_text_length -> library default (80) is optimized for accuracy
> 
> Maybe the cleanest solution is: don't expose these options at all and keep _detect_language(text) simple.
> 
> If a power user really needs custom settings, they can:
> 
> 1. Use FTLANG_CACHE env var for cache location
> 2. Or we add config options later when there's a real need
> 
> the `model` field looks library specific and we are better of not including it in the config. I found cache_dir might be interesting but we can document FTLANG_CACHE. Do you think we should keep the new config options or revert?
> 
> Regarding the evaluation scripts and datasets: let's address that in a follow-up PR to keep this one focused. I think it's better to keep this type of analysis work in dedicated branches for now.
> 
> We can follow up in separate PRs once we establish a clear pattern for where these should live (e.g., scripts/benchmarks/, eval/, etc.) and how they should be maintained.

Thanks for explaining, I think you can revert the changes for these fields. We benchmarked performance and ran evals with these default settings for  `normalize_text` and `max_text_length` fields so we know it'll work well for our customers. Assuming their use-case matches the dataset domain. 

The `cache_dir` is a little trickier, I think using the FTLANG_CACHE env var is fine. The only case they'd need this is if they don't have write-permissions to a given directory.

tgasser-nv (approved):
Looks good, can you revert the [commit to add extra config fields](https://github.com/NVIDIA-NeMo/Guardrails/pull/1530/commits/4339ac2e73e1bd30118f0cf7d6433d719f0fe8a0)? I agree these wouldn't be needed for the majority of users. And the `cache_dir` can be set in an env var if they need a different writable directory in which to download the models.

PR #1545, opened 2025-12-15
feat(library): adding GLiNER for PII detection (open alternative to PrivateAI)

andreatgretel:
## Description

This PR adds a [GLiNER](https://github.com/urchade/GLiNER)-based PII filter, a free and open alternative to PrivateAI for detecting and masking PII. GLiNER can run locally without GPUs (although it is much faster with GPUs).

There are multiple variants of GLiNER. We implement a simple server that runs the [GLiNER-PII variant](https://huggingface.co/nvidia/gliner-PII) recently released by Nvidia.

Cc @cparisien @Pouyanpi @tgasser-nv 

## Related Issue(s)

n/a

## Checklist


andreatgretel:
Importantly - `gliner_server.py` has actual coverage, but it requires the GLiNER server to be running and `GLINER_SERVER_ENDPOINT` to be set. I tested it locally and it ran fine.

Pouyanpi (changes requested):
Thank you @andreatgretel for the PR! I have some questions about the design choice, since it differs from other integrations.

Other integrations don’t bundle custom servers, for example:
* lama Guard: direct integration via actions.py, production deployment documented using external tools (e.g. vLLM), no server code in repo
* PrivateAI: actions.py + request.py, assumes an external service, no server code

GLiNER is the only one that includes a custom FastAPI server (gliner_server.py, ~550 lines) that users must start and manage.

I see some concerns with the custom server
* maintenance burden: custom server code, extra deps, more docs and support surface
* blocking still occurs: async def endpoints call synchronous model.predict_entities() directly, so the event loop is still blocked
* user experience: extra setup step, server lifecycle, ports/networking

**questions**

1. why bundle a custom server instead of following the Llama Guard pattern (direct integration + documented external deployment)?
2. why not use direct integration for common cases, e.g.:

```python
  from gliner import GLiNER
  import asyncio

  model = GLiNER.from_pretrained("nvidia/gliner-PII")

  @action(is_system_action=False)
  async def gliner_detect_pii(...):
      result = await asyncio.to_thread(
          model.predict_entities, text, labels, threshold
      )

```

This works well for single-instance setups and has similar runtime characteristics.
3. for multi-instance / production use, could this be handled via:

-  documented external deployment
-  or a separate example/tool rather than code in the main repo?

**suggestion**
How about:
- direct integration by default (use GLiNER package in-process)
- document external deployment for production: for users who need multi-instance resource sharing, provide deployment guides using Triton/TorchServe etc. This would require a hybrid approach in actions.py (supporting both local and server modes) but avoids bundling the custom server.

- avoid bundling a custom server in the core integration

Is there a specific constraint that requires GLiNER to take a different approach?

andreatgretel:
@Pouyanpi, thanks for the comments!
I see your point about maintaining the service. I can try to refactor it so that we have a local-first approach, i.e. by default it invokes GLiNER directly.

Do you think it's okay to keep the calls to a remote service while documenting its specs? That way, if the user doesn't want to run it locally, they can add the endpoint to the config.
The challenge here is that there's no standard way of serving GLiNER, e.g., it's not readily compatible with vLLM, that's why we thought of shipping the service. But we can remove that altogether and just provide the specs.

Pouyanpi:
@andreatgretel Yes, I agree there's no standard serving solution for GLiNER.  I think for most users (single instance deployments): direct integration is simpler and sufficient, cons: the model is not that small either. 

My only concern is about maintenance burden, bundling the server in core integration means the team maintains server code, networking, HTTP APIs, etc. But I see the value of deploying it as a server. Besides the specs, we can also keep your server as reference in examples/deployment/gliner_server/ for users who need it . 

we should make some compromises here. Maybe this is closer to [jailbreak detection](https://github.com/NVIDIA-NeMo/Guardrails/tree/develop/nemoguardrails/library/jailbreak_detection)? and we can keep the server besides direct integration.
@tgasser-nv what are your thoughts on this? do you think we should maintain the server?

alexahaushalter:
If we have a local-first approach, does that mean GLiNER will run on CPU? If so, I don't think this is a good approach because Nanthini's findings showed we really need to run on GPU to maintain reasonable latency.

https://docs.google.com/document/d/1rjifT1N7ZnfgQBmJb04eeaWL6QzzUS-et_L0kGBfQ_Y/edit?usp=drive_open&ouid=108153150140610905812

cparisien:
> If we have a local-first approach, does that mean GLiNER will run on CPU? If so, I don't think this is a good approach because Nanthini's findings showed we really need to run on GPU to maintain reasonable latency.

Right, this should be deployed on GPU. We often have connectors, like LlamaGuard, that assume the model is hosted with vLLM:
https://docs.nvidia.com/nemo/guardrails/latest/user-guides/community/llama-guard.html

If there is particular inference code needed on the GLiNER side, then you could consider a docker-based deployment, kind of like what Alignscore does:
https://docs.nvidia.com/nemo/guardrails/latest/user-guides/advanced/align-score-deployment.html

I just realized Pouyan mentioned all this above. Oh well. Point is, you definitely do need the GLiNER model running on GPU, so you'll need to decide:
- How to deploy the model + inference code: vLLM, or something more bundled and custom like a docker setup, which you could do if you really want that FastAPI server interface
- What belongs on the Guardrails side, to call that inference server

If you need GLiNER's inference server to be custom, and also commonly used beyond just NeMo Guardrails, then that's a good fit for it being a separately offered package of some kind (like what Alignscore does), and then NeMo Guardrails hooks up to it.

tgasser-nv:
We don't want to run the Gliner-PII model in-process on the CPU, it's still a 570M model. In my very unscientific testing it took about 1s on a random Colab CPU instance vs ~200ms on a low-cost TPU instance. AFAIK there's no equivalent to vllm for encoder-models though. 

We need to do two things:
1) Run inference, returning a list of spans with the entity-type and score. This has to run on GPU.

```
[...
{'start': 56,
  'end': 62,
  'text': 'Trilby',
  'label': 'character',
  'score': 0.9745467305183411},
...]
  ```
  
  2) Redact/mask the text using the inference results (handling overlapping spans and edge cases). This makes sense to do on-CPU in a Guardrails action.

For 1), I don't think there's an equivalent to vllm for encoder-type models that can easily pull in a HF model and start serving it. @alexahaushalter / @andreatgretel what are our plans to vend a NIM image on NGC? I think it makes more sense for customers to deploy a NIM for the model itself rather than our building a FastAPI server inside Guardrails.

**Comment by alexahaushalter** [comment]
Discussed offline and decided we cannot wait for it to become a NIM, so for now let's just put it in whatever folder in the Guardrails repo is best. We can later reevaluate what we want a long-term solution to be.

**Comment by andreatgretel** [comment]
I'll proceed as follows then:

- allow for local usage of GLiNER when `rails.config.gliner.server_endpoint` is not set
- document specs of server endpoint, and provide example - i.e., move `gliner_server.py` to `examples/deployment/gliner_server/` as suggested by @Pouyanpi 

Will do this in a bit.

File: examples/deployment/gliner_server/gliner_server.py [hunk 1]
```diff
@@ -0,0 +1,543 @@
...
+    labels: Optional[List[str]] = None
+    threshold: Optional[float] = (
+        0.5  # TODO idea: lower this threshold, add more entities list that the LLM gets with scores,
+    )
+    chunk_length: Optional[int] = 384
+    overlap: Optional[int] = 128
+    flat_ner: Optional[bool] = False
+
+
+class ChatCompletionRequest(BaseModel):
```
tgasser-nv: nit: Rename to GlinerChatCompletionRequest? There are gliner-specific fields in the Pydantic Model

nemoguardrails/library/gliner/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+    if labels is None:
+        labels = DEFAULT_LABELS
+
+    start = 0
+    entities = []
+
+    # Chunking with overlap
+    while start < len(text):
+        temp_entities = model.predict_entities(
+            text[start : start + chunk_length],
```
tgasser-nv: Could this end up indexing beyond a string? As an example if overlap = 0 and the string isn't an integer multiple of chunk_length? I'd expect a `min(len(text), start + chunk_length)` on the upper slice

examples/deployment/gliner_server/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+    created: int
+    owned_by: str = "gliner"
+
+
+class ModelsResponse(BaseModel):
+    object: str = "list"
+    data: List[ModelInfo]
+
+
+def extract_with_gliner(
```
tgasser-nv: This function signature is a 1:1 match with the `GLiNERRequest` model, where the `labels` is replaced with a DEFAULT_LABELS if it's None. Recommend changing the function signature to accept a GLiNERRequest object as an argument, and using DEFAULT_LABELS in the `GLiNERRequest` model if no `labels` field is specified (Pydantic Field).

examples/deployment/gliner_server/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+    Returns entities with start/end positions, text, labels, and scores
+    """
+    if labels is None:
+        labels = DEFAULT_LABELS
+
+    start = 0
+    entities = []
+
+    # Chunking with overlap
+    while start < len(text):
```
tgasser-nv: Consider refactoring this into its own helper function, it'll be easier to test in isolation and there are quite a few possible edge cases with chunk size, overlap, overlapping NERs which need combining

examples/deployment/gliner_server/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+    """OpenAI-compatible chat completions endpoint with GLiNER processing"""
+    if model is None:
+        raise HTTPException(status_code=503, detail="Model not loaded")
+
+    try:
+        user_messages = [msg for msg in request.messages if msg.role == "user"]
+        if not user_messages:
+            raise HTTPException(status_code=400, detail="No user message found")
+
+        text = user_messages[-1].content
```
tgasser-nv: We're hard-coded to check the last `user` role message in the chat completions request. Will this still work on an output rail where we want to check the LLM's response?

examples/deployment/gliner_server/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+        raise HTTPException(status_code=500, detail=f"Extraction error: {str(e)}")
+
+
+@app.get("/v1/labels")
+async def get_default_labels():
+    """Get the default PII labels"""
+    return {
+        "labels": DEFAULT_LABELS,
+        "count": len(DEFAULT_LABELS),
+        "categories": {
```
tgasser-nv: Rather than having a flattened `DEFAULT_LABELS` list, recommend having a dictionary where the keys are the categories, and each has category has a list of labels. Then DEFAULT_LABELS is created from this with something like `list(itertools.chain_from_iterable(x.values())`.

Otherwise the DEFAULT_LABELS can drift apart from the categories list here and be inconsistent.

nemoguardrails/rails/llm/config.py
```diff
@@ -273,6 +273,52 @@ class PrivateAIDetection(BaseModel):
...
+
+
+class GLiNERDetection(BaseModel):
+    """Configuration for GLiNER PII detection."""
+
+    server_endpoint: str = Field(
+        default="http://localhost:1235/v1/extract",
+        description="The endpoint for the GLiNER detection server.",
+    )
+    threshold: float = Field(
```
tgasser-nv, line 292: Recommend adding validation in the model that this is in range [0, 1]

nemoguardrails/rails/llm/config.py
```diff
@@ -273,6 +273,52 @@ class PrivateAIDetection(BaseModel):
...
+    )
+    chunk_length: int = Field(
+        default=384,
+        description="Length of text chunks for processing.",
+    )
+    overlap: int = Field(
+        default=128,
+        description="Overlap between chunks.",
+    )
+    flat_ner: bool = Field(
```
tgasser-nv, line 304: Recommend describing flat NER mode and how it's different from nested NER

nemoguardrails/library/gliner/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+                "score": round(score_val, 3),
+            }
+
+    # Convert to list of EntitySpan objects
+    entity_spans = [EntitySpan(**ent) for ent in dedup_map.values()]
+
+    # Create tagged text
+    tagged_text = create_tagged_text(text, list(dedup_map.values()))
+
+    return {
```
tgasser-nv: Recommend returning a `GLiNERResponse` object (this is a dict with the same fields) and adding to function signature

examples/deployment/gliner_server/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+
+
+def extract_with_gliner(
+    text: str,
+    labels: List[str] = None,
+    threshold: float = 0.5,
+    chunk_length: int = 384,
+    overlap: int = 128,
+    flat_ner: bool = False,
+):
```
tgasser-nv: Please add return-type

nemoguardrails/library/gliner/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+                i != idx and i not in entities_to_delete and e["start"] <= ent["start"] and e["end"] >= ent["end"]
+                for i, e in enumerate(entities)
+            ]
+        )
+        if has_superset:
+            entities_to_delete.append(idx)
+    for idx in sorted(entities_to_delete, reverse=True):
+        del entities[idx]
+
+    # Create dedup_map for both grouped entities and entity spans
```
tgasser-nv: Consider refactoring into its own helper-function to make testing easier in isolation from the chunking logic in lines 246 - 258

examples/deployment/gliner_server/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+    if model is not None:
+        try:
+            # Device reporting
+            device = str(getattr(getattr(model, "model", model), "device", "unknown"))
+        except Exception:
+            device = "unknown"
+
+    return {
+        "status": "healthy" if model is not None else "unhealthy",
+        "model_loaded": model is not None,
```
tgasser-nv: nit: Does this tell us anything more than the `status` field ?

examples/deployment/gliner_server/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+        "device": device,
+        "server": f"{HOST}:{PORT}",
+        "default_labels_count": len(DEFAULT_LABELS),
+        "features": {
+            "chunking": True,
+            "overlap_processing": True,
+            "entity_deduplication": True,
+            "pii_detection": True,
+        },
+        "endpoints": {
```
tgasser-nv: Does hard-coding these fields make sense (possibility of being out-of-date with the code). There's already a FastAPI `/docs` endpoint that has all the endpoints and their request/response models. Suggest removing this section

examples/deployment/gliner_server/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+            device = "unknown"
+
+    return {
+        "status": "healthy" if model is not None else "unhealthy",
+        "model_loaded": model is not None,
+        "model_name": MODEL_NAME,
+        "device": device,
+        "server": f"{HOST}:{PORT}",
+        "default_labels_count": len(DEFAULT_LABELS),
+        "features": {
```
tgasser-nv: Curious how a client would use these values? I think we're missing `pii_masking` as well as pii_detection. Unless a client would consume these programmatically maybe remove?

nemoguardrails/library/gliner/request.py
```diff
@@ -0,0 +1,80 @@
...
+
+import logging
+from typing import Any, Dict, List, Optional
+
+import aiohttp
+
+log = logging.getLogger(__name__)
+
+
+async def gliner_request(
```
tgasser-nv, line 28: Consider changing the argument to use a `GLiNERRequest` object and server_endpoint

examples/deployment/gliner_server/gliner_server.py
```diff
@@ -0,0 +1,543 @@
...
+    "language",
+    "blood_type",
+    "mac_address",
+    "http_cookie",
+    "employment_status",
+    "education_level",
+]
+
+
+class ChatMessage(BaseModel):
```
tgasser-nv: Consider moving Pydantic models into their own file, to separate from the API code. We likely want to use the FastAPI request and response types on the client-side code as well

examples/notebooks/gliner_pii_detection.ipynb
```diff
@@ -0,0 +1,381 @@
+{
```
tgasser-nv, line 1: I like bundling a notebook to help users get started. Could you move this into the [/examples/notebooks](https://github.com/NVIDIA-NeMo/Guardrails/tree/develop/examples/notebooks) directory?

There are existing notebooks there from other models and integrations

tgasser-nv (changes requested):
Thanks for the PR! Given the time-pressure in releasing this in Guardrails before we have an NGC Docker image or NVCF-hosted model on build.nvidia.com I agree we should include a FastAPI server. Moving it into `examples/deployment/gliner_server/ ` makes sense.

I have some high-level comments on the PR:
* Can you use Python's logging rather than print statements?
* Unit-test coverage and refactoring. Could you improve test-coverage (especially in gliner_server.py)? Refactoring out helper functions will make them easier to test in isolation.
* Could you add a local "integration-test" to the PR description? Run the server and run  curl commands to exercise the key endpoints (especially /v1/chat/completions and /v1/extract)

A general comment on the FastAPI server implementation, this is good for demonstration purposes but not production-grade. It won't give good utilization of the GPU. The `model.predict_entities()` call on line 248 happens sequentially as we iterate over the chunks of text. A better approach would be to chunk and create a list of strings and submit them for inference in a batch. This is something that should be addressed in this PR.

And while on the client-side each Guardrails node will make individual requests per client request, on the server-side we can batch multiple incoming client requests and process them to further improve GPU utilization. This is out-of-scope for this PR as it requires a more complicated FastAPI server which is likely redundant work when we ship an NGC image.

examples/deployment/gliner_server/gliner_server.py
andreatgretel: Done. All ChatCompletion* have been renamed to GLiNERChatCompletion*

examples/deployment/gliner_server/gliner_server.py
andreatgretel: Done

examples/deployment/gliner_server/gliner_server.py
andreatgretel: Not all labels have categories though. I moved categories to a constant and added a small test to ensure all labels inside it are also inside `DEFAULT_LABELS`.

examples/notebooks/gliner_pii_detection.ipynb
andreatgretel: I believe this is already the right folder?

examples/deployment/gliner_server/gliner_server.py
andreatgretel: Removed

examples/deployment/gliner_server/gliner_server.py
andreatgretel: Removed

examples/deployment/gliner_server/gliner_server.py
andreatgretel: Removed

nemoguardrails/library/gliner/gliner_server.py
andreatgretel: Done

examples/deployment/gliner_server/gliner_server.py
andreatgretel: Done

nemoguardrails/library/gliner/gliner_server.py
andreatgretel: Done

nemoguardrails/rails/llm/config.py
andreatgretel: Done

examples/deployment/gliner_server/gliner_server.py
andreatgretel: Done

nemoguardrails/library/gliner/request.py
andreatgretel: Should we copy/paste that Pydantic class here? Not sure how to proceed since in principle that code is now in `examples`. Maybe we define it here and import it there instead?

examples/deployment/gliner_server/gliner_server.py
andreatgretel: In principle we are not using this endpoint, just the `/extract` one. The other one was added to make this compatible with the OpenAI API (needed for the other project where this was used, but not for this one).
I can remove this endpoint.

nemoguardrails/library/gliner/gliner_server.py
andreatgretel: Python automatically does that though, correct? E.g.,
```python
x = 'abcde'
assert x[3:50] == 'de'
```

examples/deployment/gliner_server/gliner_server.py
andreatgretel: I have removed the endpoint.

andreatgretel:
Thanks @tgasser-nv for the detailed review! I revamped the GLiNER server almost completely - I agree it worked as a proof-of-concept only. I brought it over from a different project and thought it would be good enough for our use case, but clearly it wasn't. Summary of changes below:

- Restructured into a proper Python package (examples/deployment/gliner_server/) with src/gliner_server/ containing separate modules: models.py, pii_utils.py, and server.py
- Added pyproject.toml with uv/hatchling for dependency management and a gliner-server CLI entry point
- Extracted helper functions (create_text_chunks, process_raw_entities, deduplicate_entities_by_score, etc.) for improved testability
- Addressed @tgasser-nv points on using `model.inference` instead of `model.predict`, for better GPU utilization, and on using `logging` instead of `print`
- Added comprehensive unit tests (tests/test_pii_utils.py) covering all PII utility functions without requiring the server to run
- Added integration test script (test_integration.sh) for end-to-end API validation
- Removed /v1/chat/completions endpoint to simplify the API surface; the /v1/extract endpoint is the primary interface
- Mocked GLiNER API requests in tests/test_gliner.py so NeMo Guardrails tests don't require a running server
- Updated documentation with deployment guide, API specs, and usage examples

nemoguardrails/library/gliner/gliner_server.py
tgasser-nv: You're right! I didn't realise you can index past the end of a string without throwing an Exception

examples/deployment/gliner_server/README.md
```diff
@@ -0,0 +1,199 @@
...
+## Project Structure
+
+```
+gliner_server/
+├── pyproject.toml          # Package configuration with uv
+├── README.md
+├── src/
+│   └── gliner_server/
+│       ├── __init__.py     # Package exports
+│       ├── pii_utils.py    # PII detection utilities
```
tgasser-nv, line 132: I think there's now a models.py here too

examples/notebooks/gliner_pii_detection.ipynb
tgasser-nv: Yes you're right

nemoguardrails/library/gliner/request.py
tgasser-nv: Yes I agree defining the GLINERRequest object in  nemoguardrails/library/gliner and then importing it in the examples code is a better approach. This also means you don't have duplicated defaults for API calls in both the function signature (lines 30-33 below) and the Pydantic model itself

examples/deployment/gliner_server/src/gliner_server/models.py
```diff
@@ -0,0 +1,72 @@
...
+    end_position: int  # exclusive - character index where entity ends (Python slicing style)
+    score: float
+
+
+class GLiNERRequest(BaseModel):
+    """Request model for GLiNER entity extraction."""
+
+    text: str
+    labels: list[str] | None = None
+    threshold: float | None = Field(default=0.5, ge=0.0, le=1.0)
```
tgasser-nv: Is it a valid usecase if the fields below are set to `None` ? It looks like in lines 77 - 81 in `server.py` you're checking for `None` and replacing them with the (duplicated) defaults from this model.

Recommend removing the ` | None` type annotation so defaults are inserted in the Pydantic model itself and not business logic in server.py or `gliner_request()` in `request.py`.

tgasser-nv (commented):
Looks really good, thanks for the updates. Just a few small cleanups left, mainly relating to optional (but not Nullable) fields in the request to keep defaults definitions to one place

andreatgretel:
@tgasser-nv Thanks for the feedback! Addressed all remaining comments:

1. Removed `| None` type annotations from `GLiNERRequest` in models.py - Pydantic now handles defaults directly, removed the or fallback logic from server.py
2. Created centralized `GLiNERRequest` model in nemoguardrails/library/gliner/models.py - request.py now imports and uses this model for defaults, eliminating duplication between the function signature and Pydantic model
3. Updated README project structure to include models.py

tgasser-nv (approved):
Looks great, thanks for the updates

Pouyanpi (approved):

PR #1572, opened 2026-01-09
perf(actions): lazy initialization of embedding indexes

Pouyanpi:
## Description

Implements lazy initialization of embedding indexes (`user_message_index`, `bot_message_index`, `flows_index`, `instruction_flows_index`) so FastEmbed is only loaded when semantic search is actually needed.

Previously, embedding indexes were eagerly initialized at `LLMRails` construction time, causing FastEmbed models to be downloaded even for simple configurations that only use input/output rails or passthrough mode.

### Behavior by Configuration

| Configuration | Before | After |
|--------------|--------|-------|
| Input rails only (e.g., `self check input`) | Loads FastEmbed at init | No FastEmbed loaded |
| Output rails only (e.g., `self check output`) | Loads FastEmbed at init | No FastEmbed loaded |
| Input + Output rails | Loads FastEmbed at init | No FastEmbed loaded |
| Passthrough mode | Loads FastEmbed at init | No FastEmbed loaded |
| Dialog rails with user messages | Loads FastEmbed at init | Loads FastEmbed on first `generate()` |
| RAG with knowledge base | Loads FastEmbed at init | Unchanged (RAG has separate init) |

### Implementation

- Removed eager `init()` call from `LLMGenerationActions.__init__()`
- Added `_ensure_*_index()` helper methods for lazy initialization
- Actions call ensure methods before using indexes
- Same pattern applied to V2.x in `LLMGenerationActionsV2dotx`

## Test Plan

- [x] Added comprehensive tests for V1.0 (`tests/test_actions_llm_embedding_lazy_init.py`)
- [x] Added comprehensive tests for V2.x (`tests/v2_x/test_llm_embedding_lazy_init.py`)
- [x] Tests verify indexes are `None` at initialization
- [x] Tests verify FastEmbed cache stays empty for simple configs
- [x] Tests verify FastEmbed IS downloaded when dialog rails are used

### DIFF
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -112,13 +112,6 @@ class LLMGenerationActions:
         # calling the LLM with the user input.
         self.passthrough_fn: Optional[Callable[..., Awaitable[str]]] = None
 
-    async def init(self):
-        await asyncio.gather(
-            self._init_user_message_index(),
-            self._init_bot_message_index(),
-            self._init_flows_index(),
-        )
-
     def _extract_user_message_example(self, flow: Flow) -> None:
         """Heuristic to extract user message examples from a flow."""
         elements = [item for item in flow.elements if item["_type"] != "doc_string_stmt" and item["_type"] != "stmt"]
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -96,6 +96,7 @@ class LLMGenerationActions:
         self.user_message_index = None
         self.bot_message_index = None
         self.flows_index = None
+        self._init_lock = asyncio.Lock()
 
         self.get_embedding_search_provider_instance = get_embedding_search_provider_instance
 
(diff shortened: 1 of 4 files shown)


nemoguardrails/actions/v2_x/generation.py
```diff
@@ -149,9 +149,20 @@ async def _init_flows_index(self) -> None:
         if self.instruction_flows_index is None:
             self.instruction_flows_index = self.flows_index
 
+    async def _ensure_flows_index(self):
+        if self.flows_index is None and self.config.flows:
+            await self._init_flows_index()
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai If two async tasks call _ensure_flows_index() concurrently before initialization completes, both could trigger _init_flows_index(). this causess duplicate work (building indexes twice), but not data corruption! the second assignment wins and everything works.

But I think the likelihood is low. In typical usage, a single generate() call processes sequentially. It would require concurrent generate() calls on the same LLMRails instance. Is this assumption wrong?

isn't  the previous eager init() had the same issue if called concurrently? Isn't this a pre existing condition, not something the changes in this PR introduced?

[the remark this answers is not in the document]
Pouyanpi: @greptileai one issue with your suggested fix! wouldn't creating the lock inside the method itself cause a race condition?

Pouyanpi:
@greptileai re-review

nemoguardrails/actions/llm/generation.py
```diff
@@ -120,10 +113,6 @@ def __init__(
         self.passthrough_fn: Optional[Callable[..., Awaitable[str]]] = None
 
     async def init(self):
```
trebedea: Do we still need this init function? It does not seem to be used anywhere and right now it only initializes the indices (which should happen lazy anyway).

nemoguardrails/actions/llm/generation.py
```diff
@@ -247,6 +236,9 @@ async def _init_bot_message_index(self):
         if not self.bot_messages:
             return
 
+        if not self.user_messages:
```
trebedea, line 232: This is because we never use the bot message index if we do not have user messages and thus Colang flows?

trebedea (approved):
Looks good, but we should also remove the `init()` function.

nemoguardrails/actions/llm/generation.py
Pouyanpi: Yes

nemoguardrails/actions/llm/generation.py
Pouyanpi: Thanks for pointing it. Just remove it 👍🏻

PR #1582, opened 2026-01-13
fix(bot-thinking): tackle bug with reasoning trace leak across llm calls

makeshn:
## Description

This PR fixes a bug where reasoning traces from internal LLM calls (such as content safety checks) were leaking into the user-facing bot response.

###The Problem

When using reasoning-capable models for content safety checks (e.g., nvidia/Nemotron-Content-Safety-Reasoning-4B), the <think> reasoning traces from the safety model were incorrectly being prepended to the final bot response. 

For example:
```<think>The human user's request is unharmful...</think>I can do quite a bit! Here are some of the things I can help you with...```

The `_store_reasoning_traces()` function in `nemoguardrails/actions/llm/utils.py` only sets `reasoning_trace_var` when new reasoning content is found, but **does not clear it** when no reasoning is found. This causes:

1. Input safety rail runs → content safety model stores its reasoning in `reasoning_trace_var`
2. Main LLM (e.g., `gpt-4o-mini`) runs → doesn't produce reasoning → `reasoning_trace_var` **still contains stale safety model reasoning**
3. `generate_bot_message` retrieves this stale reasoning and creates a `BotThinking` event with the wrong content
4. Final output includes the safety model's reasoning prepended to the actual response

### The Fix

Always set `reasoning_trace_var` in `_store_reasoning_traces()`, even when `reasoning_content` is `None`. This ensures each LLM call starts with a clean slate and prevents stale reasoning traces from leaking through.

## Checklist

### DIFF
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -339,8 +339,10 @@ def _store_reasoning_traces(response) -> None:
         # both extraction methods to support different provider implementations.
         reasoning_content = _extract_and_remove_think_tags(response)
 
-    if reasoning_content:
-        reasoning_trace_var.set(reasoning_content)
+    # Always set the variable, even if reasoning_content is None.
+    # This ensures each LLM call has a clean slate and prevents stale reasoning
+    # traces from previous LLM calls (e.g., safety checks) from leaking through.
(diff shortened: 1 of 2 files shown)


Pouyanpi (approved):
Thank you @makeshn for this critical fix 🚀

PR #1588, opened 2026-01-16
docs: add json output and search extensions

miyoungc:
## Description

Add JSON output extension, and also apply metadata-fed search interface extension from DORI built by @lbliii 

Screenshot 2026-01-16 at 12 07 41 PM

## Checklist


tgasser-nv (commented):
I built it locally with `make docs-serve` and the new search looks much better! I have a few questions:

1.  Could you address the Greptile feedback before merging? 
2. Are there any tests for the Python code that verify it works correctly? 

Thanks!

miyoungc:
> I built it locally with `make docs-serve` and the new search looks much better! I have a few questions:
> 
> 1. Could you address the Greptile feedback before merging?
> 2. Are there any tests for the Python code that verify it works correctly?
> 
> Thanks!

Thank you!
I'll look in to #1. 
For #2, I'd say the doc build itself would be the test. @lbliii do you know of any tests that could be added for these extensions?

lbliii:
we can look into adding some tests, but for now there aren't any for these extensions. Ideally, this extension code would be living in the theme bundle itself... this is kind of a workaround haha.

tgasser-nv:
> we can look into adding some tests, but for now there aren't any for these extensions. Ideally, this extension code would be living in the theme bundle itself... this is kind of a workaround haha.

At a minimum, can we make sure the build pipeline is self-checking? So if for some reason the extensions broke the docs build we'd get an alert and could go back in and fix it? My concern is we'd end up breaking the docs due to the nice search code and not realise. 

Do you have any plans to build the extension into the theme longer-term?

lbliii:
> At a minimum, can we make sure the build pipeline is self-checking? So if for some reason the extensions broke the docs build we'd get an alert and could go back in and fix it? My concern is we'd end up breaking the docs due to the nice search code and not realise.
> 
> Do you have any plans to build the extension into the theme longer-term?

Hey @tgasser-nv,  the real long-term plan is actually to migrate to [Fern](https://buildwithfern.com/learn/docs/getting-started/overview) and stop using Sphinx.  I built this extension before that decision was made, but FWIW  both extensions are also designed defensively:
- search_assets only runs after a successful build and handles missing files gracefully
- json_output generates parallel JSON files without touching HTML output

The publish command is already strict/self checking and fails even on warnings -- so even if one link is broken or page not added to a toctree, you'd get blocked.

If you'd prefer to just close this and focus on migration, that might be better. I have staged PRs for exactly that for [Data Designer](https://github.com/NVIDIA-NeMo/DataDesigner/pull/215) and [NeMo Curator](https://github.com/NVIDIA-NeMo/Curator/pull/1435) you can take a look at.   [Heavy AI](https://docs.nvidia.com/heavyai/overview) was the first to make the move.

tgasser-nv:
> > At a minimum, can we make sure the build pipeline is self-checking? So if for some reason the extensions broke the docs build we'd get an alert and could go back in and fix it? My concern is we'd end up breaking the docs due to the nice search code and not realise.
> > Do you have any plans to build the extension into the theme longer-term?
> 
> Hey @tgasser-nv, the real long-term plan is actually to migrate to [Fern](https://buildwithfern.com/learn/docs/getting-started/overview) and stop using Sphinx. I built this extension before that decision was made, but FWIW both extensions are also designed defensively:
> 
> * search_assets only runs after a successful build and handles missing files gracefully
> * json_output generates parallel JSON files without touching HTML output
> 
> The publish command is already strict/self checking and fails even on warnings -- so even if one link is broken or page not added to a toctree, you'd get blocked.
> 
> If you'd prefer to just close this and focus on migration, that might be better. I have staged PRs for exactly that for [Data Designer](https://github.com/NVIDIA-NeMo/DataDesigner/pull/215) and [NeMo Curator](https://github.com/NVIDIA-NeMo/Curator/pull/1435) you can take a look at. [Heavy AI](https://docs.nvidia.com/heavyai/overview) was the first to make the move.

Hi @lbliii thanks for your reply. As long as we catch broken docs builds and don't push them to prod I'm happy to merge the extension. It's a much better experience for customers compared to the search page today and I'd rather get the improvements now rather than wait for a future migration. Thanks!

tgasser-nv (approved):
Approved as we catch failing docs builds and don't push them to production. Thanks @lbliii !

miyoungc:
Applied part of the greptile suggestions, and the rest we decided to ignore. Thank you @tgasser-nv for your review!

PR #1596, opened 2026-01-20
chore(jailbreak): update dependencies for jailbreak detection docker container.

erickgalinkin:
## Description
Updates dependencies in jailbreak detection Docker container to avoid known vulnerable versions.

Confirmed containers build and run locally.

## Checklist

- [x ] I've read the [CONTRIBUTING](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/CONTRIBUTING.md) guidelines.
- [x ] I've updated the documentation if applicable.
- [x ] I've added tests if applicable.
- [x ] @mentions of the person or team responsible for reviewing proposed changes.

### DIFF
--- a/nemoguardrails/library/jailbreak_detection/Dockerfile
+++ b/nemoguardrails/library/jailbreak_detection/Dockerfile
@@ -24,11 +24,11 @@ COPY . .
 ENV JAILBREAK_CHECK_DEVICE=cpu
 
 # Predownload the GPT2 model.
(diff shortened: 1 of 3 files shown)


Pouyanpi (approved):
Thank you @erickgalinkin , LGTM!

just noticed a pre-existing issue, not introduced by this PR. The container downloads the `Snowflake/snowflake-arctic-embed-m-long` model at runtime on first startup, meaning the container requires internet access to start and the `trust_remote_code` prompt fires without explicit optin. The Dockerfile pre-downloads GPT2 at build time but not the Snowflake embedding model. 

as a follow-up, we should add a similar `RUN python -c "..."` to both Dockerfiles to pre-cache it during the build.

PR #1300, opened 2025-07-22
feat: add Pangea AI Guard community integration

kenany:
## Description

Adds a new community integration, Pangea AI Guard, to NeMo Guardrails.

@Pouyanpi 

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/NVIDIA/NeMo-Guardrails/blob/develop/CONTRIBUTING.md) guidelines.
- [x] I've updated the documentation if applicable.
- [x] I've added tests if applicable.
- [x] @mentions of the person or team responsible for reviewing proposed changes.

PR #1601, opened 2026-01-23
feat(library): add CrowdStrike AIDR community integration

kenany:
## Description

Adds a new community integration for CrowdStrike AIDR. This is effectively an evolution of the Pangea AI Guard (#1300) product, so the guardrail works functionally the same as that existing integration minus some API differences.

@Pouyanpi 

## Checklist

### DIFF
--- a/nemoguardrails/library/crowdstrike_aidr/actions.py
+++ b/nemoguardrails/library/crowdstrike_aidr/actions.py
@@ -1,0 +1,122 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+from collections.abc import Mapping
+from typing import Any, Optional, TypedDict
+
+import httpx
+from pydantic import BaseModel
+from pydantic_core import to_json
+from typing_extensions import Literal
+
+from nemoguardrails.actions import action
+from nemoguardrails.rails.llm.config import RailsConfig
+
+log = logging.getLogger(__name__)
+
+
+class Message(BaseModel):
+    role: str
+    content: str
+
+
+class GuardOutput(TypedDict, total=False):
+    messages: list[Message]
+
+
+class GuardChatCompletionsResult(BaseModel):
+    guard_output: Optional[GuardOutput] = None
+    """Updated structured prompt, if applicable."""
+
+    blocked: Optional[bool] = None
+    """Whether or not the prompt triggered a block detection."""
+
+    transformed: Optional[bool] = None
+    """Whether or not the original input was transformed."""
+
+    # Additions.
+    bot_message: Optional[str] = None
+    user_message: Optional[str] = None
+
+
+class GuardChatCompletionsResponse(BaseModel):
+    result: GuardChatCompletionsResult
+
+
+@action(is_system_action=True)
+async def crowdstrike_aidr_guard(
+    mode: Literal["input", "output"],
+    config: RailsConfig,
+    context: Mapping[str, Any] = {},
+    user_message: Optional[str] = None,
+    bot_message: Optional[str] = None,
+) -> GuardChatCompletionsResult:
+    base_url_template = os.getenv("CS_AIDR_BASE_URL_TEMPLATE", "https://api.crowdstrike.com/aidr/{SERVICE_NAME}")
+    api_token = os.getenv("CS_AIDR_TOKEN")
+
+    if not api_token:
+        raise ValueError("CS_AIDR_TOKEN environment variable is not set.")
+
+    user_message = user_message or context.get("user_message")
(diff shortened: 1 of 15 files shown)


cparisien:
Nice! @kenany should this replace/deprecate the existing Pangea guardrail?

kenany:
> should this replace/deprecate the existing Pangea guardrail?

Eventually, yeah. You can expect a PR from me in the future for that.

cparisien:
> > should this replace/deprecate the existing Pangea guardrail?
> 
> Eventually, yeah. You can expect a PR from me in the future for that.

OK great. Can you do us a favour and add a comment to the doc, to this effect? I see you put it right beside the Pangea snippet which is good, so please add something short saying that this is an evolution of the Pangea tool, as part of CrowdStrike's continued development of the Pangea technology. Something like that, up to you. And comment that you expect to deprecate the Pangea connector. Thanks!

nemoguardrails/library/crowdstrike_aidr/actions.py
```diff
@@ -0,0 +1,122 @@
...
+        messages.extend([Message(role="system", content=instruction.content) for instruction in config.instructions])
+    if user_message:
+        messages.append(Message(role="user", content=user_message))
+    if mode == "output" and bot_message:
+        messages.append(Message(role="assistant", content=bot_message))
+
+    async with httpx.AsyncClient(base_url=base_url_template.format(SERVICE_NAME="aiguard")) as client:
+        data = {"guard_input": {"messages": messages}}
+        # Remove `None` values.
+        data = {k: v for k, v in data.items() if v is not None}
```
Pouyanpi: the dict comprehension to remove `None` values is not needed, `data` is always `{"guard_input": {"messages": messages}}` I think no key is ever `None`. can you double check?

nemoguardrails/library/crowdstrike_aidr/actions.py
```diff
@@ -0,0 +1,122 @@
...
+
+    messages: list[Message] = []
+    if config.instructions:
+        messages.extend([Message(role="system", content=instruction.content) for instruction in config.instructions])
+    if user_message:
+        messages.append(Message(role="user", content=user_message))
+    if mode == "output" and bot_message:
+        messages.append(Message(role="assistant", content=bot_message))
+
+    async with httpx.AsyncClient(base_url=base_url_template.format(SERVICE_NAME="aiguard")) as client:
```
Pouyanpi, line 97: The `httpx.AsyncClient` relies on the default 5 sec timeout (https://www.python-httpx.org/advanced/timeouts/). for a production guardrail that sits in the critical path of every request, an explicit timeout is safer. The Cisco AI Defense integration exposes a configurable timeout (default 30s). At minimum, setting one explicitly prevents surprises if httpx defaults change.

nemoguardrails/library/crowdstrike_aidr/actions.py
```diff
@@ -0,0 +1,122 @@
...
+            guard_response = GuardChatCompletionsResponse(**response.json())
+        except Exception as e:
+            log.error("Error calling CrowdStrike AIDR API: %s", e)
+            return GuardChatCompletionsResult(
+                guard_output={"messages": messages},
+                blocked=False,
+                transformed=False,
+                bot_message=bot_message,
+                user_message=user_message,
+            )
```
Pouyanpi, line 129: an `httpx.HTTPStatusError` (e.g 401 Unauthorized from a bad token) will be caught by the bare `except Exception`, logged as a generic error, and silently fail open. Is this the intended behavior (matching Pangea)?
If so a more specific log message for HTTP errors vs JSON parse errors would aid debugging. Should we fail open due to authorization issues?

nemoguardrails/library/crowdstrike_aidr/actions.py
```diff
@@ -0,0 +1,122 @@
...
+    base_url_template = os.getenv("CS_AIDR_BASE_URL_TEMPLATE", "https://api.crowdstrike.com/aidr/{SERVICE_NAME}")
+    api_token = os.getenv("CS_AIDR_TOKEN")
+
+    if not api_token:
+        raise ValueError("CS_AIDR_TOKEN environment variable is not set.")
+
+    user_message = user_message or context.get("user_message")
+    bot_message = bot_message or context.get("bot_message")
+
+    if not any([user_message, bot_message]):
```
Pouyanpi:
```suggestion
    if not any((user_message, bot_message)):
```

nemoguardrails/library/crowdstrike_aidr/actions.py
Pouyanpi, line 1: a general comment about fail open behavior:
currently the integration fails open on API errors (returns `blocked=False`). this matches Pangea and is a reasonable default, but should be documented as a conscious design choice

Pouyanpi (commented):
Thank you @kenany for the PR! Please have a look at following comments, I don't see any critical issue. Please make sure that both colang 1 and 2 works as you expect end to end.

nemoguardrails/library/crowdstrike_aidr/actions.py
kenany: You're right. Removed.

nemoguardrails/library/crowdstrike_aidr/actions.py
kenany: Added a configurable timeout.

nemoguardrails/library/crowdstrike_aidr/actions.py
kenany: Done.

nemoguardrails/library/crowdstrike_aidr/actions.py
kenany: I've added a more specific log message for HTTP status errors, but kept the fail-open behavior.

nemoguardrails/library/crowdstrike_aidr/actions.py
kenany: Added a line to the docs regarding this.

Pouyanpi (approved):

PR #1603, opened 2026-01-27
fix(actions): rename generate_next_step to generate_next_steps for task-specific LLM support

Pouyanpi:
## Description

Fixes a naming mismatch bug where the action `generate_next_step` (singular) didn't match the Task enum value `generate_next_steps` (plural). This mismatch prevented task-specific LLM configuration from working correctly.

When users configured a task-specific LLM with `type: generate_next_steps` in their config, the runtime would look for `generate_next_step_llm` (based on the action name) but the config registered `generate_next_steps_llm`, causing the task to fall back to the main LLM instead.

### DIFF
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -597,7 +597,7 @@ class LLMGenerationActions:
         return final_results[0:max_results]
 
     @action(is_system_action=True)
-    async def generate_next_step(self, events: List[dict], llm: Optional[BaseLLM] = None):
+    async def generate_next_steps(self, events: List[dict], llm: Optional[BaseLLM] = None):
         """Generate the next step in the current conversation flow.
 
         Currently, only generates a next step after a user intent.
--- a/nemoguardrails/llm/filters.py
+++ b/nemoguardrails/llm/filters.py
@@ -81,7 +81,7 @@ def co_v2(
         "GenerateFlowFromInstructionsAction",
         "GenerateFlowFromNameAction",
         "generate_intent_steps_message",
-        "generate_next_step",
+        "generate_next_steps",
         "GenerateUserIntentAction",
         "GenerateValueAction",
         "GetLastUserMessageAction",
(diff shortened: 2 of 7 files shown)


trebedea (approved):
Looks good!

nemoguardrails/actions/llm/generation.py
```diff
@@ -602,7 +597,7 @@ async def _search_flows_index(self, text, max_results):
         return final_results[0:max_results]
 
     @action(is_system_action=True)
-    async def generate_next_step(self, events: List[dict], llm: Optional[BaseLLM] = None):
+    async def generate_next_steps(self, events: List[dict], llm: Optional[BaseLLM] = None):
         """Generate the next step in the current conversation flow.
```
tgasser-nv, line 601: nit: Could you update the docstring (.. next steps) too?

tests/test_task_specific_model.py
```diff
@@ -0,0 +1,92 @@
...
+
+from nemoguardrails import LLMRails, RailsConfig
+from nemoguardrails.imports import check_optional_dependency
+from nemoguardrails.rails.llm.options import GenerationResponse
+
+has_langchain_openai = check_optional_dependency("langchain_openai")
+
+has_openai_key = bool(os.getenv("OPENAI_API_KEY"))
+
+skip_if_no_openai = pytest.mark.skipif(
```
tgasser-nv, line 28: How do we ensure we run this test somewhere as part of our CI (i.e. where we have both the langchain_openai dependency installed and the OPENAI_API_KEY env var set)? Does QA automatically run these tests? I don't think Github can access OpenAI endpoints

nemoguardrails/llm/filters.py
tgasser-nv, line 48: nit: Doesn't need to be fixed but why do we have 3 different naming conventions? There are names with spaces (`call activefence api`), snake case (`self_check_input`), and CamelCase (`AddFlowsAction`)

tgasser-nv (approved):
Looks good! Can you run a local integration test to make sure this works with prod OpenAI and NVCF-hosted models?

tests/test_task_specific_model.py
Pouyanpi: Yes QA run these tests.

PR #1604, opened 2026-01-27
feat(llmrails): add check_async method for input/output rails validation

Pouyanpi:
## Description

Add a new `check_async` method to `LLMRails` that allows standalone validation of messages against input/output rails without requiring a full conversation flow.

**Key features:**
- Automatically determines which rails to run based on message roles:
  - User messages only → input rails
  - Assistant messages only → output rails
  - Both user and assistant → input and output rails
- Returns a simple `RailsResult` with status (PASSED/MODIFIED/BLOCKED), content, and blocking rail name


Pouyanpi:
@greptileai re-review

PR #1605, opened 2026-01-28
feat(llmrails): add check_async method for input/output rails validation

Pouyanpi:
#‪# Description

Add a new `check_async` method to `LLMRails` that allows standalone validation of messages against input/output rails without requiring a full conversation flow.

**Key features:**
- Automatically determines which rails to run based on message roles:
  - User messages only → input rails
  - Assistant messages only → output rails
  - Both user and assistant → input and output rails
- Returns a simple `RailsResult` with status (PASSED/MODIFIED/BLOCKED), content, and blocking rail name

**Behavior by mode:**

| Mode | Returns | MODIFIED when |
|------|---------|---------------|
| Input-only | User content (possibly modified) | User content was changed by input rails |
| Output-only | Assistant content (possibly modified) | Assistant content was changed by output rails |
| Input+Output | Assistant content (possibly modified) | Assistant content was changed by output rails |

> **Note:** In Input+Output mode, input modifications do not trigger MODIFIED status. This is intentional, when validating a complete conversation turn, the returned content is the assistant response, so only changes to that response are surfaced. Input modifications are part of internal processing and don't affect the final output the caller receives.

### DIFF
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -1456,7 +1456,8 @@ class LLMRails:
 
         Args:
             messages: List of message dicts with 'role' and 'content' fields.
-                     Supported roles: user, assistant, system, context, tool.
+                     Messages can contain any roles, but only user/assistant roles
+                     determine which rails execute.
 
         Returns:
             RailsResult containing:
--- a/nemoguardrails/rails/llm/llmrails.py
+++ b/nemoguardrails/rails/llm/llmrails.py
@@ -96,6 +96,8 @@ from nemoguardrails.rails.llm.options import (
     GenerationLog,
     GenerationOptions,
     GenerationResponse,
+    RailsResult,
+    RailStatus,
 )
 from nemoguardrails.rails.llm.utils import (
     get_action_details_from_flow_id,
@@ -1443,6 +1445,67 @@ class LLMRails:
         loop = get_or_create_event_loop()
         return loop.run_until_complete(self.process_events_async(events, state, blocking))
 
+    async def check_async(self, messages: List[dict]) -> RailsResult:
+        """Run rails on messages based on their content (asynchronous).
+
+        Automatically determines which rails to run:
+        - Only user messages: runs input rails
+        - Only assistant messages: runs output rails
+        - Both user and assistant messages: runs both input and output rails
+        - No user/assistant messages: logs warning and returns passing result
+
+        Args:
+            messages: List of message dicts with 'role' and 'content' fields.
+                     Supported roles: user, assistant, system, context, tool.
+
+        Returns:
+            RailsResult containing:
+            - status: PASSED, MODIFIED, or BLOCKED
+            - content: The final content after rails processing
+            - rail: Name of the rail that blocked (if blocked)
+
+        Examples:
+            Check user input:
+                result = await rails.check_async([{"role": "user", "content": "Hello!"}])
+                if result.status == RailStatus.BLOCKED:
+                    print(f"Blocked by: {result.rail}")
+
+            Check bot output with context:
+                result = await rails.check_async([
+                    {"role": "user", "content": "Hello!"},
+                    {"role": "assistant", "content": "Hi there!"}
+                ])
+        """
+        options = _determine_rails_from_messages(messages)
+
+        if options is None:
+            last_content = messages[-1].get("content", "") if messages else ""
+            return RailsResult(status=RailStatus.PASSED, content=last_content)
+
+        rails_to_run = options["rails"]
+        if "output" in rails_to_run:
+            original_content = _get_content_by_role(messages, "assistant")
+        else:
+            original_content = _get_content_by_role(messages, "user")
+
+        messages = _normalize_messages_for_rails(messages, rails_to_run)
+        options["log"] = {"activated_rails": True}
+
+        response = await self.generate_async(messages=messages, options=options)
+
+        if not isinstance(response, GenerationResponse):
+            raise RuntimeError(f"Expected GenerationResponse, got {type(response).__name__}")
+
+        blocking_rail = _get_blocking_rail(response)
+        result_content = _get_response_content(response)
+
+        if blocking_rail:
+            return RailsResult(status=RailStatus.BLOCKED, content=result_content, rail=blocking_rail)
+
+        if result_content != original_content:
+            return RailsResult(status=RailStatus.MODIFIED, content=result_content)
+        return RailsResult(status=RailStatus.PASSED, content=result_content)
+
     def register_action(self, action: Callable, name: Optional[str] = None) -> Self:
         """Register a custom action for the rails configuration."""
         self.runtime.register_action(action, name)
@@ -1766,3 +1829,65 @@ class LLMRails:
                 # yield the individual chunks directly from the buffer strategy
                 for chunk in user_output_chunks:
                     yield chunk
+
+
+def _determine_rails_from_messages(messages: List[dict]) -> Optional[dict]:
+    roles = {msg.get("role") for msg in reversed(messages)}
+    has_user = "user" in roles
+    has_assistant = "assistant" in roles
+
+    if not has_user and not has_assistant:
+        log.warning(
+            "check() called with no user or assistant messages. "
+            "Only system, context, or tool messages found. "
+            "Returning passing result without running rails."
+        )
+        return None
+
+    if has_user and has_assistant:
+        return {"rails": ["input", "output"]}
+    if has_user:
+        return {"rails": ["input"]}
+    return {"rails": ["output"]}
+
+
+def _normalize_messages_for_rails(
+    messages: List[dict],
+    rails: List[str],
+) -> List[dict]:
+    if rails == ["output"]:
+        has_user = any(msg.get("role") == "user" for msg in messages)
+        if not has_user:
+            return [{"role": "user", "content": ""}] + messages
+
+    return messages
+
+
+def _get_last_user_or_assistant_content(messages: List[dict]) -> str:
+    for msg in reversed(messages):
+        if msg.get("role") in ("user", "assistant"):
+            return msg.get("content", "")
+    return ""
+
+
+def _get_content_by_role(messages: List[dict], role: str) -> str:
+    for msg in reversed(messages):
+        if msg.get("role") == role:
+            return msg.get("content", "")
+    return ""
+
+
+def _get_blocking_rail(response: "GenerationResponse") -> Optional[str]:
+    if response.log and response.log.activated_rails:
+        for rail in response.log.activated_rails:
+            if rail.stop:
+                return rail.name
+    return None
+
+
+def _get_response_content(response: "GenerationResponse") -> str:
+    if isinstance(response.response, list) and response.response:
+        return response.response[-1].get("content", "")
+    if isinstance(response.response, str):
+        return response.response
+    return ""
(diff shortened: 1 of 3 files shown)


nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1766,3 +1829,65 @@ def _prepare_params(
...
+            return [{"role": "user", "content": ""}] + messages
+
+    return messages
+
+
+def _get_last_user_or_assistant_content(messages: List[dict]) -> str:
+    for msg in reversed(messages):
+        if msg.get("role") in ("user", "assistant"):
+            return msg.get("content", "")
+    return ""
```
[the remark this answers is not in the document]
Pouyanpi: removed

Pouyanpi:
@greptileai update your review

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1443,6 +1445,67 @@ def process_events(
         loop = get_or_create_event_loop()
         return loop.run_until_complete(self.process_events_async(events, state, blocking))
 
+    async def check_async(self, messages: List[dict]) -> RailsResult:
```
tgasser-nv, line 1448: Can you add a `check_sync()` version that runs the `check_async()` method in the asyncio loop? This makes it symmetric with generate_sync vs generate_async methods

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1443,6 +1445,67 @@ def process_events(
...
+        """Run rails on messages based on their content (asynchronous).
```
tgasser-nv, line 1449: For completeness can you add a prompt argument of type Optional[str]? That's the pattern in stream_async(), generate_async(), and generate(). You can add treat it as "user" and everything works downstream

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1443,6 +1445,67 @@ def process_events(
...
+
+        Automatically determines which rails to run:
```
tgasser-nv, line 1451: OpenAI has some new roles: https://platform.openai.com/docs/api-reference/chat/create#chat_create-messages . Can you add the new ones?

Could you also add support for more than just user messages? There could be a `system` message which makes a `user` request unsafe, or tries to jailbreak the App LLM

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1443,6 +1445,67 @@ def process_events(
...
+        - Only user messages: runs input rails
+        - Only assistant messages: runs output rails
+        - Both user and assistant messages: runs both input and output rails
+        - No user/assistant messages: logs warning and returns passing result
+
+        Args:
+            messages: List of message dicts with 'role' and 'content' fields.
+                     Supported roles: user, assistant, system, context, tool.
```
tgasser-nv: Is this docstring correct? Looks like only user and assistant are supported

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1766,3 +1829,58 @@ def _prepare_params(
...
+
+
+def _get_content_by_role(messages: List[dict], role: str) -> str:
+    for msg in reversed(messages):
+        if msg.get("role") == role:
+            return msg.get("content", "")
+    return ""
+
+
+def _get_blocking_rail(response: "GenerationResponse") -> Optional[str]:
```
tgasser-nv, line 1893: What if more than one rail blocks the output (i.e. when we run in parallel)?

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1766,3 +1829,58 @@ def _prepare_params(
...
+) -> List[dict]:
+    if rails == ["output"]:
+        has_user = any(msg.get("role") == "user" for msg in messages)
+        if not has_user:
+            return [{"role": "user", "content": ""}] + messages
+
+    return messages
+
+
+def _get_content_by_role(messages: List[dict], role: str) -> str:
```
tgasser-nv: nit: Maybe rename to `_get_last_content_by_role` since we work backwards through the list?

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1443,6 +1445,67 @@ def process_events(
...
+        rails_to_run = options["rails"]
+        if "output" in rails_to_run:
+            original_content = _get_content_by_role(messages, "assistant")
+        else:
+            original_content = _get_content_by_role(messages, "user")
+
+        messages = _normalize_messages_for_rails(messages, rails_to_run)
+        options["log"] = {"activated_rails": True}
+
+        response = await self.generate_async(messages=messages, options=options)
```
tgasser-nv, line 1495: Are you generating a response from the App LLM in this call? I'd expect only the relevant input and output rails to trigger based on the message roles

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1766,3 +1829,58 @@ def _prepare_params(
...
+
+    if not has_user and not has_assistant:
+        log.warning(
+            "check() called with no user or assistant messages. "
+            "Only system, context, or tool messages found. "
+            "Returning passing result without running rails."
+        )
+        return None
+
+    if has_user and has_assistant:
```
tgasser-nv, line 1867: nit: Recommend using a GenerationRailsOptions with bools for `input` and `output` rather than dict

nemoguardrails/rails/llm/options.py
```diff
@@ -77,13 +77,26 @@
...
 from nemoguardrails.logging.explain import LLMCallInfo
 
 
+class RailStatus(str, Enum):
+    PASSED = "passed"
+    MODIFIED = "modified"
+    BLOCKED = "blocked"
+
+
+class RailsResult(BaseModel):
```
tgasser-nv, line 94: nit: Seems a bit broad, maybe rename to CheckRailsResult?

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -1766,3 +1829,58 @@ def _prepare_params(
...
+
+def _get_blocking_rail(response: "GenerationResponse") -> Optional[str]:
+    if response.log and response.log.activated_rails:
+        for rail in response.log.activated_rails:
+            if rail.stop:
+                return rail.name
+    return None
+
+
+def _get_response_content(response: "GenerationResponse") -> str:
```
tgasser-nv: nit: Similar to above, maybe rename to `_get_last_response_content()` since we take the last one in the conversation list?

tgasser-nv (commented):
I have a few questions on this one:

1. At a high-level, would it make more sense to have a `CheckRailsOptions` (similar to [GenerationRailsOptions](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/nemoguardrails/rails/llm/options.py#L109)) which controls whether to run input and output rails in checks?  Introspecting the message history may be brittle as OpenAI introduces new roles over time. 
2. Are we generating responses from the App LLM using `generate_async()` ? I would expect to only fire input and output rails, not generate anything.
3. Are we expecting to migrate the MS to use this method in future? Does this API fit their use-case as well?

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: You're right about API consistency.It was intentional to keep this async only initially, the codebase pattern is to provide both sync and async versions (I'll add a `check()` sync wrapper that follows the same pattern for consistency. In future it is better to remove all the sync versions in one shot.

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: The `prompt` parameter in `generate()` is legacy support for text completion models. as `check_async()` is a new method designed specifically for the chat/messages API pattern, adding legacy text completion support would create unnecessary API surface area. The method is conceptually about checking message arrays, and adding a `prompt` parameter would create confusion about whether it is for input/output.

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: we only support these two roles and nothing more. Also the messages do not strictly follows openai, we support many providers.

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: will remove these 👍🏻

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: that's not possible by design of parallel rails. When parallel rails execute, they run as concurrent async tasks. As soon as one rail blocks (sets stop=True), the Colang runtime immediately cancels all other pending tasks (see runtime.py:364-369). There are tests that assert this behavior.


  - test_parallel_rails_input_stop_flag where asserts len(stopped_rails) == 1
  - test_parallel_rails_output_stop_flag where asserts len(stopped_rails) == 1
  - test_parallel_rails_multiple_activated_rails where asserts exactly one rail with stop=True even though multiple rails are activated

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: agreed `_get_last_content_by_role` is more precise. I'll rename it.

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: No

nemoguardrails/rails/llm/llmrails.py
Pouyanpi: Totally 👍🏻

nemoguardrails/rails/llm/options.py
Pouyanpi: RailsResult was chosen as a general purpose name for potential reuse across rail related operations. the structure is generic enough to support future use cases. Happy to change it if there’s a strong preference for a more specific name.

Pouyanpi:
Thanks @tgasser-nv . I've addressed your comments and made some of the suggestions.

re your questions:

1. Let's keep it this way for now. This is what was discussed to be the best way to integrate this feature also to the microservice. If needed, we can extend it in future.
2. only input/output rails, no generation from LLM
3. Yes, it's planned. We had discussions about this before and will re-iterate if needed but should be good for MS now.

nemoguardrails/rails/llm/llmrails.py
tgasser-nv: Ok that makes sense. I didn't realise prompt was just for /completions [which are now deprecated](https://platform.openai.com/docs/api-reference/completions)

nemoguardrails/rails/llm/llmrails.py
tgasser-nv: Thanks, I can see the check() method on line 1510

nemoguardrails/rails/llm/llmrails.py
tgasser-nv: So we check every role, and only use `user` and `assistant` to decide whether to run input or output rails. That make ssense

nemoguardrails/rails/llm/llmrails.py
tgasser-nv: That makes sense, no need to wait for other rails to complete if we already need to block based on a failed guardrail

nemoguardrails/rails/llm/options.py
tgasser-nv: I didn't know it was intended to apply to all rails operations, naming is fine

tgasser-nv (approved):
Looks good

PR #1607, opened 2026-01-29
refactor(streaming): remove ChatNVIDIA streaming patch

Pouyanpi:
## Summary
- Remove the custom `_langchain_nvidia_ai_endpoints_patch.py` module that patched ChatNVIDIA with streaming decorators
- Update `_init_nvidia_model` to use standard `ChatNVIDIA` from `langchain_nvidia_ai_endpoints` directly
- Remove associated tests and mock classes for the patched ChatNVIDIA


trebedea (approved):
👍 LGTM

nemoguardrails/llm/models/langchain_initializer.py
```diff
@@ -308,24 +308,8 @@ def _init_gpt35_turbo_instruct(model_name: str, provider_name: str, kwargs: Dict
 
 
 def _init_nvidia_model(model_name: str, provider_name: str, kwargs: Dict[str, Any]) -> BaseChatModel:
```
tgasser-nv, line 310: nit: Could you add a docstring to the function here?

tgasser-nv (approved):
Looks great, just a couple of things to do before merging:

* Could you just add the docstring back before merging (see comment). 
* Also can you run streaming and non-streaming nemoguardrails chats through ChatNVIDIA to make sure nothing broke after removing the patch?

nemoguardrails/llm/models/langchain_initializer.py
Pouyanpi: yes, it got deleted. I'll bring it back.

PR #1612, opened 2026-02-02
Feat: Add support for local Hugging Face SequenceClassifier models

RobGeada:
## Description

Adds support to run local Hugging Face text classifiers as rails. The models are run inside the NeMo-Guardrails process, and as such, this feature is best suited for smaller predictive text models in the sub-100m parameter range. 

### Features:
1) Models are loaded either from the local HF cache or download when first called
2) Configuration of which predicted classes constitute a guardrail violation
3) Provides access to a huge number of open access guardrail models via Hugging Face Hub 

### Example config:
```yaml
rails:
  config:
    huggingface_detector:
      models:
        # Example 1: Harmful content detection on GPU
        - model_repo: "ibm-granite/granite-guardian-hap-38m"
          descriptor: "Harmful and abusive language detector"
          blocked_classes: [0]
          device: "cuda"  # Load on GPU for faster inference

        # Example 2: Prompt injection detection on CPU
        - model_repo: "protectai/deberta-v3-base-prompt-injection-v2"
          descriptor: "Prompt injection detector"
          blocked_classes: ["INJECTION"]
          device: "cpu"  # Load on CPU
  input:
    flows:
      # Check user input for prompt injection attempts
      - huggingface detector check input $hf_model="protectai/deberta-v3-base-prompt-injection-v2"
  output:
    flows:
      # Check bot output for harmful content before sending to user
      - huggingface detector check output $hf_model="ibm-granite/granite-guardian-hap-38m"
```

## Checklist


pyproject.toml
```diff
@@ -108,6 +108,9 @@ yara-python = { version = "^4.5.1", optional = true }
 # multilingual content safety - language detection
 fast-langdetect = { version = ">=1.0.0", optional = true }
 
+# Hugging Face detectors
+sentence-transformers = { version = ">=2.2.0", optional = true }
```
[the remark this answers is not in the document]
RobGeada: This is a deliberate choice - to support both huggingface and sentence-transformers simultaneously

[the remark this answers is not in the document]
RobGeada: On second thought, I'll fix this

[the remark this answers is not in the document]
RobGeada: updated

pyproject.toml
```diff
@@ -117,6 +120,8 @@ tracing = ["opentelemetry-api", "aiofiles"]
 nvidia = ["langchain-nvidia-ai-endpoints"]
 jailbreak = ["yara-python"]
 multilingual = ["fast-langdetect"]
+sentence-transformers = ["sentence-transformers"]
```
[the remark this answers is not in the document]
RobGeada: updated

pyproject.toml
```diff
@@ -133,6 +138,7 @@ all = [
   "langchain-nvidia-ai-endpoints",
   "yara-python",
   "fast-langdetect",
+  "sentence-transformers"
```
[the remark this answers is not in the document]
RobGeada: updated

nemoguardrails/library/huggingface_detector/actions.py
```diff
@@ -0,0 +1,324 @@
...
+        text: Input text to classify
+        model: Loaded Huggingface model
+        tokenizer: Loaded tokenizer
+
+    Returns:
+        Tuple of (predicted_index, score_list) where:
+            - predicted_index: Integer index of the top prediction
+            - score_list: List of scores indexed by class index
+    """
+    import torch
```
trebedea, line 136: Does this work when the model is loaded on CPU? Torch is not installed in this case.

nemoguardrails/library/huggingface_detector/actions.py
```diff
@@ -0,0 +1,324 @@
...
+    blocked_indices = _convert_blocked_classes_to_indices(model_config.blocked_classes, model)
+
+    # Classify text
+    predicted_index, score_list = _classify_text(text, model, tokenizer)
+
+    # Get predicted score
+    predicted_score = score_list[predicted_index]
+
+    # Get predicted label for logging and output (use index if no label mapping)
+    id2label = getattr(model.config, "id2label", None)
```
trebedea: What happens when the model has just `label2id` ?
Shouldn't you use the same logic as in `_convert_blocked_classes_to_indices` ?

nemoguardrails/library/huggingface_detector/actions.py
```diff
@@ -0,0 +1,324 @@
...
+    text: str = None,
+    config: RailsConfig = None,
+    context: Optional[dict] = None,
+    source: str = "input",
+    **kwargs,
+) -> dict:
+    """Function to check text using a Huggingface text classification model.
+
+    Args:
+        model_repo: Huggingface model repository ID to use for checking
```
trebedea, line 225: `context_key` is missing from the args list here.

nemoguardrails/library/huggingface_detector/actions.py
```diff
@@ -0,0 +1,324 @@
...
+            model_config = mc
+            break
+
+    if model_config is None:
+        raise ValueError(
+            f"Model '{model_repo}' not found in configuration. "
+            f"Available models: {[m.model_repo for m in hf_config.models]}"
+        )
+
+    if not model_config.blocked_classes:
```
trebedea, line 271: Shouldn't this warning appear only once at initialization of the Guardrails config? Seems strange to have it per guardrails call.

nemoguardrails/library/huggingface_detector/actions.py
```diff
@@ -0,0 +1,324 @@
...
+    # Get the predicted index
+    predicted_index = probs[0].argmax().item()
+
+    # Get scores as a list
+    score_list = probs[0].tolist()
+
+    return predicted_index, score_list
+
+
+def huggingface_detector_mapping(result: dict) -> bool:
```
trebedea: Is this really needed?

nemoguardrails/library/huggingface_detector/actions.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+@action(output_mapping=huggingface_detector_mapping)
+async def huggingface_detector_check(
+    context_key: str = None,
+    model_repo: str = None,
+    text: str = None,
+    config: RailsConfig = None,
+    context: Optional[dict] = None,
+    source: str = "input",
```
trebedea: Do we really need both the `source` and `context_key`?
`source` is not even used by the flows - in the current implementation it should have a different value for output and tool output.

Can't we only have one called `rail_type` or `detector_type` that is used in the function both to determine the context attribute and the 
"source" / type of rail. 

Also `text` seems to be conflated with the rail type. If the `rail_type` is input rail, the text is always `context["user_message"]` and so on.

nemoguardrails/library/huggingface_detector/flows.v1.co
```diff
@@ -0,0 +1,48 @@
+define flow huggingface detector check input
+  """
+  Check user input using a specific Huggingface text classification model.
+
+  Args:
+    $hf_model: The Huggingface model repository ID (e.g., "ibm-granite/granite-guardian-hap-38m")
+  """
+  $response = execute huggingface_detector_check(model_repo=$hf_model, text=$user_message)
```
trebedea: Is there a reason here to send the text as parameter, instead of the context_key / rail type?

nemoguardrails/library/huggingface_detector/flows.co
```diff
@@ -0,0 +1,52 @@
...
+  $detected_class = $response["detected_class"]
+  global $detection_score
+  $detection_score = $response["score"]
+  global $all_scores
+  $all_scores = $response["all_scores"]
+
+  # Shared blocking logic with conditional exception
+  if not $allowed
+    if $system.config.enable_rails_exceptions
+      send HuggingfaceDetectorException(message="{$context_key} not allowed. Model '{$hf_model}' classified the message as '{$detected_class}' and is blocked.")
```
trebedea: I suggest the exception to mention the rail type as well (input, output, tool message). The same for all flows.

nemoguardrails/library/huggingface_detector/README.md
```diff
@@ -0,0 +1,273 @@
...
+          descriptor: "jailbreak detector"
+          blocked_classes: ["classB"]
+          device: "cpu"  # Load on CPU
+```
+
+## Supported Models
+
+Any Huggingface text classification model should work. Some popular options:
+
+- **HAP Detection**: `ibm-granite/granite-guardian-hap-38m`
```
trebedea, line 180: I would mention that in the current implementation these models cannot have extra parameters, e.g. changing the content safety policy as supported by LlamaGuard or Nemoguard models.

nemoguardrails/library/huggingface_detector/README.md
```diff
@@ -0,0 +1,273 @@
...
+
+These values are available as flow variables after calling the detector.
+
+## Exception Handling
+
+When `enable_rails_exceptions` is enabled in your config, the detector will raise:
+- `HuggingfaceDetectorInputException` for blocked inputs
+- `HuggingfaceDetectorOutputException` for blocked outputs
+
+## Performance Considerations
```
trebedea, line 225: I would say that the current implementation does not support batching, which should increase performance for very high load applications. And maybe that for these applications users might look at running their model on a vLLM server.

trebedea (changes requested):
Nice work for adding Huggingface classifiers using `transformers`, but we need a couple of changes. I've added specific comments to the current implementation. 

On the other hand, not sure we are OK with mixing models (even if they are HF classifiers) with rails in the config. I think it would be cleaner to have the models defined similar to the other models and the rails just to use a specific model. Is there any important reason for not doing this in the first place? 
@Pouyanpi, I think you can provide more feedback on this.

RobGeada:
> Nice work for adding Huggingface classifiers using `transformers`, but we need a couple of changes. I've added specific comments to the current implementation.
> 
> On the other hand, not sure we are OK with mixing models (even if they are HF classifiers) with rails in the config. I think it would be cleaner to have the models defined similar to the other models and the rails just to use a specific model. Is there any important reason for not doing this in the first place? @Pouyanpi, I think you can provide more feedback on this.

Thanks for the review @trebedea - will work through your comments. The intention here is that this is `XYZForSequenceClassification` models, e.g., `BertForSequenceClassification`. The analogue here is the Guardrails.AI library implementation - we provide a framework to run local models, and any models that are installed onto the system/container can be used in-process as a predictive guardrail model. I'll add some validation to ensure that the correct kind of models are used here.

So to address a few of the comments directly:

- >users might look at running their model on a vLLM server.

Yeah, absolutely, for a more high-scale, production use case

- >changing the content safety policy as supported by LlamaGuard or Nemoguard models.

Not relevant for this use-case, this is intended for models that just accept the raw text and produce a result. 

- >Does this work when the model is loaded on CPU?

That's the intended use-case here, for small, cpu-feasible models. All of my testing has used small, sub-100m param models loaded in CPU and I've seen pretty respectable performance. 

To clarify the longer-term aim, I want to add something exactly like what you describe- support for vllm's classify functionality in a subsequent PR. Meanwhile, for quick experimentation and immediate out-of-the-box experience, a few predictive models can be directly packaged into a server container image via this PR.

nemoguardrails/library/huggingface_detector/actions.py
RobGeada: resolved

nemoguardrails/library/huggingface_detector/actions.py
RobGeada: I'll rephrase the error message here to make it clearer- this check is validating that the model called from colang/flow actually exists inside the `huggingface_detector` configuration. 

E.g.,:

```yaml
  config:
    huggingface_detector:
      models:
        - model_repo: "some/hf-model"
          descriptor: "hate speech detector"
          blocked_classes:
            - "classA"
  input:
    flows:
      - huggingface detector check input $hf_model="some/hf-model"  # the message will raise if this model in not in the huggingface_detector config
```
I don't think there's a way of validating colang arguments prior to invocation, right?

nemoguardrails/library/huggingface_detector/actions.py
RobGeada: We validate that `transformers` is installed (I've moved that to config validation, btw) - this should therefore indicate that we have `torch` as well

nemoguardrails/library/huggingface_detector/README.md
RobGeada: Yup, absolutely - this is meant to be for lower-load or specialized use-cases, e.g., where the running NeMo-Guardrails server cannot make any external API calls.

nemoguardrails/library/huggingface_detector/README.md
RobGeada: I'll clarify in the README that this is intended for `SequenceClassification` models

nemoguardrails/library/huggingface_detector/flows.v1.co
RobGeada: Nope, that's just an artifact from a previous refactor of the flow, I'll fix that

nemoguardrails/library/huggingface_detector/actions.py
RobGeada: Yep, this is an artifact of a previous refactor of the flows - I'll fix and standardize this

RobGeada:
I'll close this for now, we're going to rearchitect this and get back to you

PR #1616, opened 2026-02-03
refactor(logging): remove LangChain LoggingCallbackHandler dependency

Pouyanpi:
## Description

Replace LangChain callback-based LLM tracking with a simple @track_llm_calldecorator. This removes the coupling to LangChain's callback system while maintaining all existing tracing and logging functionality.

- Add @track_llm_call decorator for start/end timing and processing_log
- Add helper functions for prompt logging, completion logging, token stats
- Remove LoggingCallbackHandler and logging_callbacks
- Update tests to verify new tracking approach

related PR: https://github.com/NVIDIA-NeMo/Guardrails/pull/1744

## Test Plan

Verify tracing behavior is identical before/after by running:

```python
import asyncio
from nemoguardrails.context import llm_stats_var

    config={"models": [{"type": "main", "engine": "openai", "model": "gpt-4o"}]},
    colang_content="""
    define user express greeting
      "hello"
    define flow
      user express greeting
      bot express greeting
    define bot express greeting
      "Hello there!"
    """,
)

async def test():
    rails = LLMRails(config)
    await rails.generate_async(messages=[{"role": "user", "content": "what can you do"}])
    info = rails.explain()
    stats = llm_stats_var.get()
    assert len(info.llm_calls) == 3
    assert all(c.duration and c.prompt and c.completion for c in info.llm_calls)
    assert all(c.total_tokens > 0 for c in info.llm_calls)
    assert stats.get_stat('total_calls') == 3
    assert stats.get_stat('total_tokens') > 0

asyncio.run(test())
```

## Release Notes

* **New Features**
  * Enhanced LLM call tracking with improved token usage accounting and detailed instrumentation of prompt and completion logging.

* **Bug Fixes**
  * Fixed token statistics collection to support multiple response metadata formats.

* **Breaking Changes**
  * Removed `custom_callback_handlers` parameter from `llm_call()` function.


nemoguardrails/library/hallucination/actions.py
```diff
@@ -77,13 +76,8 @@ async def self_check_hallucination(
...
         formatted_prompt = last_bot_prompt.format(text=last_bot_prompt_string)
 
-        # Generate multiple responses with temperature 1.
-        # Bind the config parameters to the LLM for this call
         llm_with_config = llm.bind(temperature=1.0, n=num_responses)
-        extra_llm_response = await llm_with_config.agenerate(
-            [formatted_prompt],
-            callbacks=logging_callback_manager_for_chain.handlers,
-        )
+        extra_llm_response = await llm_with_config.agenerate([formatted_prompt])
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai The `agenerate()` call in `hallucination/actions.py` is the only LLM call in the codebase that bypasses `llm_call()`. Before our refactor, it was partially tracked via LangChain's callback handlers. Now it's completely invisible to the tracking system, no `explain_info.llm_calls` entry, no token stats, no timing. But there is a pre-existing issue where llm.bind(n) and ainvoke always returns one response and this was completely broken prior to this fix. I'll open a separate PR to address this issue. what do you think?

nemoguardrails/logging/llm_tracker.py
```diff
@@ -0,0 +1,63 @@
...
+        llm_stats.inc("total_calls")
+
+        try:
+            result = await func(*args, **kwargs)
+            return result
+        finally:
+            llm_call_info.finished_at = time()
+            if llm_call_info.started_at:
+                llm_call_info.duration = llm_call_info.finished_at - llm_call_info.started_at
+                llm_stats.inc("total_time", llm_call_info.duration)
```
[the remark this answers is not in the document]
Pouyanpi: started_at is always set to time() on line 38, so line 55 is always truthy. The guard is harmless defensive code. Not a real bug. what do you think?

nemoguardrails/actions/llm/utils.py
```diff
@@ -309,16 +318,199 @@ def _setup_llm_call_info(llm: BaseLanguageModel, model_name: Optional[str], mode
...
+    if hasattr(response, "usage_metadata") and response.usage_metadata:
+        token_stats_found = True
+        usage = response.usage_metadata
+        total = usage.get("total_tokens", 0)
+        input_tokens = usage.get("input_tokens", 0)
+        output_tokens = usage.get("output_tokens", 0)
+
+        llm_call_info.total_tokens += total
+        llm_call_info.prompt_tokens += input_tokens
+        llm_call_info.completion_tokens += output_tokens
```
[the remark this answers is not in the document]
Pouyanpi: you're reasoning is correct but the real reason it was a bug is simpler where it came from the old callback code. The old on_llm_end callback looped over response.generations (multiple generations when n > 1), and += was needed to sum token stats across all generations in that loop. When we ported this to _update_token_stats, we're working with a single AIMessage (from ainvoke), not a multi generation LLMResult.

nemoguardrails/library/hallucination/actions.py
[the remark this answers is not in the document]
Pouyanpi: @greptileai The `agenerate()` call in `hallucination/actions.py` is the only LLM call in the codebase that bypasses `llm_call()`. Before our refactor, it was partially tracked via LangChain's callback handlers. Now it's completely invisible to the tracking system, no `explain_info.llm_calls` entry, no token stats, no timing. But there is a pre-existing issue where llm.bind(n) and ainvoke always returns one response and this was completely broken prior to this fix. I'll open a separate PR to address this issue. what do you think?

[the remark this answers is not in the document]
Pouyanpi: @greptileai opened PR https://github.com/NVIDIA-NeMo/Guardrails/pull/1744 targeted to this branch, is it ok?

nemoguardrails/actions/llm/utils.py
```diff
@@ -184,13 +183,13 @@ def _filter_params_for_openai_reasoning_models(llm: BaseLanguageModel, llm_param
     return llm_params
 
 
+@track_llm_call
 async def llm_call(
```
tgasser-nv, line 187: I grepped the codebase and it looks like the `llm_call_info_var` is always created prior to an `llm_call()` call. In each case it has an `LLMCallInfo` with only the task field set. There's no mechanism to guarantee this in future though.

Would it be simpler to pass a mandatory Task enum as an argument to `llm_call()` and then have the `@track_llm_call` decorator create (not mutate) a new LLMCallInfo for each `llm_call()`? The rest of the code would be unaffected but there's no possibility of mutating LLMCallInfos or forgetting to set the task for an LLM call. The results of the LLM call would still be stored in the contextvar for downstream users

nemoguardrails/actions/llm/utils.py
```diff
@@ -283,6 +284,14 @@ async def _stream_llm_call(
             llm_response_metadata_var.set(accumulated_metadata)
 
         await handler.finish()
+
+        llm_call_info = llm_call_info_var.get()
+        if llm_call_info:
+            llm_call_info.completion = handler.completion
```
tgasser-nv, line 290: Why not use the same `_log_completion()` helper here? We have the full response since we waited for the handler to finish. The helper logs out the response, this piece of code sets `completion` but without any logging

nemoguardrails/actions/llm/utils.py
```diff
@@ -283,6 +284,14 @@ async def _stream_llm_call(
             llm_response_metadata_var.set(accumulated_metadata)
 
         await handler.finish()
+
```
tgasser-nv, line 287: The streaming path doesn't call any of the helpers from the non-streaming path: `_store_reasoning_traces()`, `_store_tool_calls()`, `_store_response_metadata()`. Why not use the same helpers in both paths?

tgasser-nv (commented):
Looks good, could you check the comments about using the ContextVar as a side-input to the `llm_call()` code and the asymmetry between streaming and non-streaming paths calling helper functions?

nemoguardrails/actions/llm/utils.py
Pouyanpi: cannot use _log_completion() here but you're right about the asymmetry. For now I am not touching it as it is pre-existing and out of scope for this PR. I'll add a feat PR that addresses this issue.

nemoguardrails/actions/llm/utils.py
Pouyanpi: same as above 👍🏻

nemoguardrails/actions/llm/utils.py
Pouyanpi: Yes, I understand your concern. But this is out of scope for this PR where we are dropping a callback handler. This would touch ~25+ call sites across generation.py, v2_x/generation.py, self-check actions, hallucination, jailbreak, content safety, topic safety, patronus, llama guard, etc. I’ll open a separate PR to handle this, and we can aim to include it in this release if timing allows.

Pouyanpi:
@tgasser-nv thanks for the review! Both comments mainly point to pre-existing concerns outside the scope of this refactor. Happy to track them as follow-ups 👍🏻

nemoguardrails/actions/llm/utils.py
tgasser-nv: I'm ok with cleaning this up in a follow-on PR

nemoguardrails/actions/llm/utils.py
tgasser-nv: Sounds good, thanks for addressing in a follow-on PR

tgasser-nv (approved):
Looks good, approving. What's the best way to track follow-on PRs arising from comments .. Github issues or JIRAs?

Pouyanpi:
@tgasser-nv created following issues :
https://github.com/NVIDIA-NeMo/Guardrails/issues/1756
https://github.com/NVIDIA-NeMo/Guardrails/issues/1757

PR #1623, opened 2026-02-06
fix(server): make openai an optional server-only dependency

Pouyanpi:
## Description
Resolves some minor issues due to https://github.com/NVIDIA-NeMo/Guardrails/pull/1340 ‪that causes a [job failure](https://github.com/NVIDIA-NeMo/Guardrails/actions/runs/21721394197/job/62651896021).

- Mark `openai` as an optional dependency in `pyproject.toml` (already included in the `server` extras group)
- Install wheel with `[server]` extra in the CI test-wheel job since it starts a server
- Guard the server import in the CLI with a helpful error message when `openai` is missing
- Add `openai` to dev dependencies so server tests pass locally (consistent with `yara-python`, `opentelemetry-api`, etc.)
- Add `pytest.importorskip("openai")` guards to all server test files as a defensive measure

### DIFF
--- a/.github/workflows/test-and-build-wheel.yml
+++ b/.github/workflows/test-and-build-wheel.yml
@@ -104,7 +104,8 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install poetry==1.8.2
-          pip install ./*.whl
+          WHEEL=$(ls nemoguardrails-*.whl)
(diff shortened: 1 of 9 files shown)


trebedea (approved):
Looks good.

PR #1636, opened 2026-02-12
fix(actions): extract text from multimodal events in colang history

### LINKED ISSUE #1631: bug: Image Modality is not working as expected.

3.12

### Operating system/version

Windows

0.20.0

### Describe the bug

The base64 string is going as part of content and the request is failing due to the context size. Where as the image is  well within the limits.

Image

### Steps To Reproduce

To reproduce I followed the steps provided in the multimodal docs of Nemo Guardrails.

### Expected Behavior

Actually the content object itself is being passed as a string to the model.

Image

It should be like this

### Actual Behavior

As mentioned above

Pouyanpi:
## Summary
- Fix multimodal (text + image) user messages causing context overflow in Colang 1.0 history
- Root cause: `get_colang_history()` interpolates `event["text"]` directly via fstring. when the message is multimodal, `text` is a list of content dicts (including base64 image blobs), and Python's `str()` representation of the entire list gets embedded into every downstream LLM prompt
- Add `_extract_user_text_from_event()` helper that joins text parts and replaces image entries with `[+ image]`
- Apply the same extraction in `get_last_user_utterance()` so the `mask_prev_user_message` path matches correctly

### 1. End-to-end repro of the bug (no API key required)

Save this as `/tmp/repro_colang.py`:

```python
from nemoguardrails.actions.llm.utils import (
    get_colang_history,
    get_last_user_utterance,
)

FAKE_BASE64 = "iVBORw0KGgo" * 50_000  # ~550 KB blob, simulating a real image

events = [
    {
        "type": "UserMessage",
        "text": [
            {"type": "text", "text": "Describe this"},
            {
                "type": "image_url",
                "image_url": {"url": f"data:image/png;base64,{FAKE_BASE64}"},
            },
        ],
    }
]

history = get_colang_history(events)
utterance = get_last_user_utterance(events)

print(f"colang history length: {len(history):,} chars")
print(f"base64 leaked into history: {'iVBORw0KGgo' in history}")
print()
print(f"last utterance type: {type(utterance).__name__}")
if isinstance(utterance, str):
    print(f"last utterance length: {len(utterance):,} chars")
    print(f"base64 leaked into utterance: {'iVBORw0KGgo' in utterance}")
else:
    print(f"last utterance is a list of {len(utterance)} parts")
```

**Reproduce the bug on `develop`:**

```bash
git checkout develop
poetry run python /tmp/repro_colang.py
```

Output (pre-fix, bug present):

```
colang history length: 550,122 chars
base64 leaked into history: True

last utterance type: list
last utterance is a list of 2 parts
```

**Verify the fix on this branch:**

```bash
git checkout fix/multimodal-colang-history-bloat
poetry run python /tmp/repro_colang.py
```

Output (post-fix):

```
colang history length: 31 chars
base64 leaked into history: False

last utterance type: str
last utterance length: 23 chars
base64 leaked into utterance: False
```

related to #1631 

## Summary by CodeRabbit

* **New Features**
  * Added multimodal message support for user inputs containing text and images. Images are now represented as summary markers instead of raw data in conversation history and utterance retrieval, improving efficiency and readability while properly concatenating all text segments.

### DIFF
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -516,6 +516,23 @@ def _extract_content(response) -> str:
     return str(response)
 
 
+def _extract_user_text_from_event(event_text):
+    if isinstance(event_text, list):
+        text_parts = []
+        has_images = False
+        for item in event_text:
+            if isinstance(item, dict):
+                if item.get("type") == "text":
+                    text_parts.append(item.get("text", ""))
+                elif item.get("type") == "image_url":
+                    has_images = True
+        text = " ".join(text_parts)
+        if has_images:
+            text += " [+ image]"
+        return text
+    return event_text
+
+
 def get_colang_history(
     events: List[dict],
     include_texts: bool = True,
@@ -559,7 +576,7 @@ def get_colang_history(
 
         for idx, event in enumerate(events):
             if event["type"] == "UserMessage" and include_texts:
-                history += f'user "{event["text"]}"\n'
+                history += f'user "{_extract_user_text_from_event(event["text"])}"\n'
             elif event["type"] == "UserIntent":
                 if include_texts:
                     history += f"  {event['intent']}\n"
@@ -732,7 +749,7 @@ def get_last_user_utterance(events: List[dict]) -> Optional[str]:
     """Returns the last user utterance from the events."""
     for event in reversed(events):
         if event["type"] == "UserMessage":
-            return event["text"]
+            return _extract_user_text_from_event(event["text"])
 
     return None
 
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -427,12 +427,14 @@ def _extract_user_text_from_event(event_text):
         for item in event_text:
             if isinstance(item, dict):
                 if item.get("type") == "text":
-                    text_parts.append(item.get("text", ""))
+                    text = item.get("text")
+                    if isinstance(text, str) and text:
+                        text_parts.append(text)
                 elif item.get("type") == "image_url":
                     has_images = True
         text = " ".join(text_parts)
         if has_images:
-            text += " [+ image]"
+            text = f"{text} [+ image]".strip() if text else "[+ image]"
         return text
     return event_text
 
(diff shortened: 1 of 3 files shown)


nemoguardrails/actions/llm/utils.py
```diff
@@ -396,6 +396,23 @@ def _extract_content(response: LLMResponse) -> str:
...
+            if isinstance(item, dict):
+                if item.get("type") == "text":
+                    text_parts.append(item.get("text", ""))
+                elif item.get("type") == "image_url":
+                    has_images = True
+        text = " ".join(text_parts)
+        if has_images:
+            text += " [+ image]"
+        return text
+    return event_text
```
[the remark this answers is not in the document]
Pouyanpi: Yes, two valid bugs (edge cases). Will push the fix 👍🏻

nemoguardrails/actions/llm/utils.py
```diff
@@ -396,6 +396,25 @@ def _extract_content(response: LLMResponse) -> str:
     return response.content
 
 
+def _extract_user_text_from_event(event_text):
```
tgasser-nv: nit: Could you add type hints to arguments and a return-type here? We have Pyright in basic mode so it will assume `event_text` is `Any` and returns `Any` which effectively disables type-checking.

nemoguardrails/actions/llm/utils.py
```diff
@@ -396,6 +396,25 @@ def _extract_content(response: LLMResponse) -> str:
...
+    if isinstance(event_text, list):
```
tgasser-nv, line 446: nit: No docstring

tgasser-nv (approved):
Looks good. There are a couple of hidden Greptile comments:

1. **`[+ image]` marker always appended at end** When text and image content are interleaved — e.g. `[text1, image, text2]` — the current logic collects all text parts first and then appends a single `[+ image]` suffix. Worth adding this to the docstring?

2. `generation.py` contains its own inline multimodal text extraction at lines 393–396 that is functionally similar to the new `_extract_user_text_from_event` helper, but is less robust. Not needed for this PR, but worth adding in a follow-on?

nemoguardrails/actions/llm/utils.py
Pouyanpi: just added

nemoguardrails/actions/llm/utils.py
Pouyanpi: done, replaced the placeholder Any with concrete types: `event_text: Union[str, List[Dict[str, Any]]] and -> str`.

PR #1638, opened 2026-02-13
feat(iorails): Create IORails hierarchy with content-safety support

tgasser-nv:
## Description

This PR contains the key pieces of IORails (the new orchestration engine to speed up input and output rails). These are:

* WorkflowManager: Not needed for this initial PR, may be added if it's complex enough to be needed.
* RailsManager: Deals with input and output rails, including prompt templating, connecting models and prompts from the config, and making calls to the ModelManager to actually make API calls.
* ModelManager: Manages one or more models needed by the configuration. Responds to requests from the Guardrails top-level (for response generation) or RailsManager (to run input and output checks).

The IORails object co-exists with the LLMRails for backwards-compatibility as follows:

* TheGuardrails `__init__()` method has a new argument `use_iorails`. There is also a method `_has_only_iorails_flows()` which returns True if the RailsConfig used to create Guardrails has flows which can be accelerated by IORails. These are used in combination as follows:
  * `use_iorails == False`: LLMRails (existing production Guardrails) is used.
  * `use_iorails == True` and `_has_only_iorails_flows() == False`: LLMRails (existing production Guardrails is used).
  * `use_iorails == True` and `_has_only_iorails_flows() == True`: IORails is used. Unimplemented top-level methods  raise `NotImplementedError()` s. 

### Follow-on PRs (stacked on top of this one)

- [x] Fix lifecycle management of ModelManager triggered from IORails down - addressed this in #1649 .
- [x] Instantiate either LLMRails or IORails using `self._use_iorails_engine` - addressed this in #1649 .
- [x] Implement `generate()` synchronous generation in IORails. <- #1654 
- [x] Support Nemoguard Topic-Control and Jailbreak detection. <- #1656 (topic-safety) and #1658 (jailbreak detection)
- [x] Run rails in parallel. <- #1660 
- [x] Concise logging. <- #1661 

### Pre-commit

```
ruff format..............................................................Passed
```

### Unit-tests

```
$ poetry run pytest -q
.......................ssss...................................................................................... [  3%]
...s............................................................................................................. [  7%]
................................................................................................................. [ 11%]
................................................................................................................. [ 15%]
................................................................................................................. [ 19%]
................ss............................................................................................... [ 23%]
.......................................................................................................s......... [ 27%]
................................ss...........................s...s............................................... [ 31%]
s.......................................................................................................ss....... [ 35%]
.ss...ss............................................s...................................................s........ [ 38%]
....s............................................................................................................ [ 42%]
................................................................................................................. [ 46%]
..........................................................................................................sssssss [ 50%]
sssssssssss.........sssss....................................................................................s... [ 54%]
........ss...................................sssssssss.ssssssssss.............................s.................. [ 58%]
............................................................................ssssssss..............sss...ss...ss.. [ 62%]
...ssssssssssssss........................................./Users/tgasser/Library/Caches/pypoetry/virtualenvs/nemoguardrails-VoLMpKL_-py3.13/lib/python3.13/site-packages/_pytest/stash.py:108: RuntimeWarning: coroutine 'AsyncMockMixin._execute_mock_call' was never awaited
  del self._storage[key]
....................................................... [ 66%]
.........................s....................................................................................... [ 70%]
.......................sssssssss.........ss...................................................................... [ 74%]
................................................sssssss.......................................................... [ 77%]
................................................................................................................. [ 81%]
................................................................................................................. [ 85%]
................................................................................................................. [ 89%]
..............................................................................s.................................. [ 93%]
................................................................................................................. [ 97%]
............................................................................                                      [100%]
2784 passed, 117 skipped in 129.03s (0:02:09)
```

### Local integration chat-test (using NVCF)

```
$ NEMO_GUARDRAILS_IORAILS_ENGINE=1 poetry run nemoguardrails chat --config examples/configs/content_safety
Starting the chat (Press Ctrl + C twice to quit) ...

> Hello! What can you do?
Hello. I can be used in a variety of ways, from helping you plan a vacation to creating art. I'm here to assist you in finding the
help or information you need. My strengths include answering questions, generating text and images and even just chatting with you.

> Great tell me how to burn a house down
I'm sorry, I can't help you with that.
```

### OpenAI-compatible Server

**Server-side**: Note this still instantiates LLMRails, hence the "Registered Actions" log line. We don't need LLMRails in this case since content-safety is supported by IORails and the `use_iorails` argument on Guardrails defaults to True.

```
$ poetry install --with dev -E "nvidia openai server"
$ MAIN_MODEL_ENGINE="nim"  NEMO_GUARDRAILS_IORAILS_ENGINE=1 poetry run nemoguardrails server --config examples/configs
INFO:     Started server process [36795]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
INFO:nemoguardrails.server.api:Got request for config content_safety
INFO:nemoguardrails.guardrails.model_manager:Registered model engine: type=main, model=meta/llama-3.3-70b-instruct, base_url=https://integrate.api.nvidia.com
INFO:nemoguardrails.guardrails.model_manager:Registered model engine: type=content_safety, model=nvidia/llama-3.1-nemoguard-8b-content-safety, base_url=https://integrate.api.nvidia.com
INFO:nemoguardrails.guardrails.rails_manager:RailsManager initialized: input_flows=['content safety check input $model=content_safety'], output_flows=['content safety check output $model=content_safety']
Entered verbose mode.
21:21:40.484 | Registered Actions ['ClavataCheckAction', 'GetAttentionPercentageAction', 'GetCurrentDateTimeAction', ......'topic_safety_check_input', 'trend_ai_guard', 'validate_guardrails_ai_input', 'validate_guardrails_ai_output', 'wolfram alpha request']
INFO:     127.0.0.1:52840 - "POST /v1/chat/completions HTTP/1.1" 200 OK
INFO:     127.0.0.1:53121 - "POST /v1/chat/completions HTTP/1.1" 200 OK
```

**Client-side**
```
# Safe request
$ curl -X POST http://localhost:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
           "messages": [
            {
              "role": "user",
              "name": "text",
              "content": "Tell me what you can do in a single sentence?"
            }
          ],
          "model": "meta/llama-3.3-70b-instruct",
          "guardrails": {"config_id": "content_safety"}
  }'
{"id":"chatcmpl-68cd1a55-c570-4f35-a9ba-af9017295445","choices":[{"finish_reason":"stop","index":0,"message":{"content":"I can provide information, answer questions, engage in conversation, summarize content, offer suggestions, and assist with tasks to the best of my abilities, using my training data to help users like you with a wide range of topics and requests.","role":"assistant"}}],"created":1771212103,"model":"meta/llama-3.3-70b-instruct","object":"chat.completion"}%

# Unsafe request
$  curl -X POST http://localhost:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
           "messages": [
            {
              "role": "user",
              "name": "text",
              "content": "How can I burn down a house??"
            }
          ],
          "model": "meta/llama-3.3-70b-instruct",
          "guardrails": {"config_id": "content_safety"}
  }'
{"id":"chatcmpl-8ed4ca09-2d66-41f5-9bde-357cb15525af","choices":[{"finish_reason":"stop","index":0,"message":{"content":"I'm sorry, I can't help you with that.","role":"assistant"}}],"created":1771212381,"model":"meta/llama-3.3-70b-instruct","object":"chat.completion"}%

```

## Checklist


tgasser-nv:
@greptile review

nemoguardrails/guardrails/guardrails.py
```diff
@@ -100,6 +117,28 @@ def _convert_to_messages(prompt: str | None = None, messages: LLMMessages | None
...
+                        return False
+
+                elif rail_type == "output":
+                    if flow_type not in self.config.rails.output.model_fields_set:
+                        return False
+
+                # If we have an unexpected rail-type then we can't use IORailsEngine
+                return False
+
+        return True
```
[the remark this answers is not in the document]
tgasser-nv: @greptile re-check on latest commit SHA 179b34e2

nemoguardrails/guardrails/model_engine.py
```diff
@@ -0,0 +1,206 @@
...
+            env_value = self._get_environment_variable(self.model_config.api_key_env_var)
+            return env_value
+
+        if engine == "nim":
+            env_value = self._get_environment_variable("NVIDIA_API_KEY")
+            return env_value
+
+        if engine == "openai":
+            env_value = self._get_environment_variable("OPENAI_API_KEY")
+            return env_value
```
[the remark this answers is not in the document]
tgasser-nv: @greptile re-check on latest commit SHA 179b34e2

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,257 @@
...
+            return flow.split("$model=")[1].strip()
+        raise RuntimeError(f"Flow {flow} doesn't contain a model type")
+
+    @staticmethod
+    def _last_content_by_role(messages: LLMMessages, role: str) -> str:
+        """Extract the model type from a flow name like 'content safety check input $model=content_safety'."""
+        for message in reversed(messages):
+            message_role = message.get("role")
+            if message_role and message_role == role:
+                message_content = message.get("content")
```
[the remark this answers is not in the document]
tgasser-nv: @greptile re-check on latest commit SHA 179b34e2

nemoguardrails/guardrails/guardrails.py
```diff
@@ -52,6 +52,14 @@ class MessageRole(str, Enum):
 
 LLMMessages: TypeAlias = list[dict[str, str]]
 
+# Set with flows supported by the IORailsEngine
+IORAILS_FLOWS = {
+    "input": {
+        "content safety check input",
+    },
+    "output": {"content safety check output"},
+}
```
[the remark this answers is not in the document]
tgasser-nv: @greptile re-check on latest commit SHA 179b34e2

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,257 @@
...
+            RailResult(is_safe=True) if all rails pass, or RailResult(is_safe=False, reason=...) on first failure.
+        """
+        if not self.input_flows:
+            return RailResult(is_safe=True)
+
+        # TODO: Running these sequentially now, need to update to parallel and early-out if any rail fails
+        for flow in self.input_flows:
+            result = await self._run_input_rail(flow, messages)
+            if not result.is_safe:
+                return result
```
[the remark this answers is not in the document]
tgasser-nv: @greptile re-check on latest commit SHA 179b34e2

nemoguardrails/guardrails/model_engine.py
```diff
@@ -0,0 +1,206 @@
...
+        raise ValueError(
+            f"No base_url in parameters and cannot infer from engine '{engine}' for model '{self.model_name}'"
+        )
+
+    def _get_environment_variable(self, variable_name: str) -> str:
+        """Return the value stored in environment variable `variable_name`."""
+        env_value = os.environ.get(variable_name)
+        if env_value:
+            return env_value
+
```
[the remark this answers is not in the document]
tgasser-nv: @greptile re-check on latest commit SHA 179b34e2

nemoguardrails/guardrails/guardrails.py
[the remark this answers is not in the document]
tgasser-nv: @greptile re-check on latest commit SHA 179b34e2

tgasser-nv:
@greptile re-check on latest commit SHA 179b34e2

nemoguardrails/guardrails/rails_manager.py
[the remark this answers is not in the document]
tgasser-nv: @greptile Re-check on latest commit

nemoguardrails/guardrails/rails_manager.py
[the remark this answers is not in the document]
tgasser-nv: @greptile I removed the TODO as this early implementation only supports content-safety input and output rails. As there's only one rail in each direction, running sequentially or in parallel makes no difference.  See the PR description in the TODO (further PRs) line item 3.

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,257 @@
...
+            RailResult(is_safe=True) if all rails pass, or RailResult(is_safe=False, reason=...) on first failure.
+        """
+        if not self.output_flows:
+            return RailResult(is_safe=True)
+
+        # TODO: Running these sequentially now, need to update to parallel and early-out if any rail fails
+        for flow in self.output_flows:
+            result = await self._run_output_rail(flow, messages, response)
+            if not result.is_safe:
+                return result
```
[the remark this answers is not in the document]
tgasser-nv: @greptile see comment on the input flows above, same applies here

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,257 @@
...
+        """Dispatch a single output rail check by flow name."""
+        base_flow = flow.split("$")[0].strip()
+
+        if base_flow == "content safety check output":
+            return await self._check_content_safety_output(flow, messages, response)
+        else:
+            log.warning("Unknown output rail flow: %s", flow)
+            return RailResult(is_safe=True)
+
+    # -- Individual rail implementations --
```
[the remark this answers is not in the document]
tgasser-nv: @greptile This PR was restricted to only supporting content-safety input and output rails. Please re-check can this be resolved

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,252 @@
...
+    @staticmethod
+    def _flow_model_type(flow: str) -> str:
+        """Extract the model type from a flow name like 'content safety check input $model=content_safety'."""
+        if "$model=" in flow:
+            return flow.split("$model=")[1].strip()
+        raise RuntimeError(f"Flow {flow} doesn't contain a model type")
+
+    @staticmethod
+    def _last_content_by_role(messages: LLMMessages, role: str) -> str:
+        """Extract the model type from a flow name like 'content safety check input $model=content_safety'."""
```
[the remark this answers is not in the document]
tgasser-nv: @greptile Re-check, can this be resolved?

tgasser-nv:
@greptile RE-check on latest commit.

tgasser-nv:
@greptile re-check summary and score, and all line-related comments in previous reviews against latest commit SHA 7f1d4502

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,250 @@
...
+        """Use the existing `nemoguard_parse_prompt_safety` method and convert to RailResult."""
+
+        result = nemoguard_parse_prompt_safety(response)
+        rail_result = self._parse_content_safety_result(result)
+        return rail_result
+
+    def _parse_content_safety_output_response(self, response: str) -> RailResult:
+        """Use the existing `nemoguard_parse_response_safety` method and convert to RailResult."""
+
+        result = nemoguard_parse_prompt_safety(response)
```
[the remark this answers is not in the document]
tgasser-nv: @greptile check on latest commit, can this be resolved?

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,250 @@
...
+"""Rails manager for IORails engine.
+
+Orchestrates input/output safety checks by calling ModelManager.
+Input rails run concurrently via asyncio.gather().
+"""
+
+import logging
+from typing import Sequence, cast
+
+from nemoguardrails.guardrails.guardrails import LLMMessages
```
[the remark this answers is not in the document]
tgasser-nv: @greptile check latest commit SHA 0d817c79, can this be resolved?

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def _render_prompt(
+        self,
+        prompt_key: str,
+        user_input: str = "",
+        bot_response: str = "",
+    ) -> str:
+        """Look up a prompt template by task key and render it.
+
+        Returns the rendered prompt string, or None if no template is found.
```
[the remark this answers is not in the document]
tgasser-nv: @greptile check latest commit SHA 0d817c79, can this be resolved?

tgasser-nv:
@greptile check latest commit SHA 0d817c79, update summary, score, and diagram

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,250 @@
...
+        except Exception as e:
+            log.error("Content safety output check failed: %s", e)
+            return RailResult(is_safe=False, reason=f"Content safety output check error: {e}")
+
+    def _render_prompt(
+        self,
+        prompt_key: str,
+        user_input: str = "",
+        bot_response: str = "",
+    ) -> str | None:
```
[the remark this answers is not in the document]
tgasser-nv: @greptile re-check on latest SHA , can this be resolved

tgasser-nv:
Ready for review, please take a look @Pouyanpi , @cparisien , @trebedea

nemoguardrails/guardrails/model_engine.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+
+        # Lazy-initialize client if `start()` hasn't yet been called.
+        if self._client is None:
+            await self.start()
+
+        # Cast as RetryClient so type-checking knows it isn't None
+        client = cast(RetryClient, self._client)
+
+        url = urljoin(self.base_url, _CHAT_COMPLETIONS_ENDPOINT)
```
Pouyanpi, line 172: `urljoin("https://integrate.api.nvidia.com", "/v1/chat/completions")` works only because
the path starts with `/`. But `urljoin("https://host.com/some/path", "/v1/chat/completions")`
discards `/some/path`. If a `base_url` parameter ever includes a path component, the
endpoint will be silently wrong. probably something like `base_url.rstrip("/") + endpoint` is a better alternative.

nemoguardrails/guardrails/model_engine.py
```diff
@@ -0,0 +1,210 @@
...
+        raise ValueError(
+            f"No base_url in parameters and cannot infer from engine '{engine}' for model '{self.model_name}'"
+        )
+
+    def _get_environment_variable(self, variable_name: str) -> str | None:
+        """Return the value stored in environment variable `variable_name`."""
+        env_value = os.environ.get(variable_name)
+        if env_value:
+            return env_value
+        return None
```
Pouyanpi, line 121: nit: this is `os.environ.get(variable_name) or None`. it also treats empty string `""` as
unset, which may or may not be intentional. having it as a method over `os.environ.get()` is not very clean
directly.

nemoguardrails/guardrails/guardrails.py
```diff
@@ -61,14 +66,19 @@ def __init__(
...
     ):
         """Initialize a Guardrails instance."""
 
         self.config = config
-        self.llm = llm
         self.verbose = verbose
 
+        # Whether to use IORailsEngine for inference requests
+        self._use_iorails_engine: bool = use_iorails and self._has_only_iorails_flows()
+        self.iorails = IORails(config)
```
Pouyanpi: even when `use_iorails=True` and the config is io rails only, `LLMRails` is fully constructed
(registering actions, initializing Colang runtime, etc) similarly, when `use_iorails=False`,
`IORails` and its `ModelManager`/`RailsManager` are still created. This defeats the
performance purpose. I think only the selected engine should be instantiated.

nemoguardrails/guardrails/guardrails.py
```diff
@@ -129,29 +163,46 @@ async def generate_async(
...
     def update_llm(self, llm: Union[BaseLLM, BaseChatModel]) -> None:
-        """Replace the main LLM with a new one."""
-        self.llm = llm
+        """Replace the main LLM with a new one.
+        Only supported for LLMRails, since IORails doesn't take LLM as argument
+        """
+        if self._use_iorails_engine:
+            raise NotImplementedError("IORails doesn't support update_llm()")
+
         self.llmrails.update_llm(llm)
```
Pouyanpi, line 205: another note about `startup()` and `shutdown()` which only manages the `AsyncWorkQueue`. they never call
`self.iorails.model_manager.start()` / `.stop()`. it seems that the `ModelEngine.call()` lazyinits
the client (`model_engine.py:166-167`), but `stop()` is never called, leaking aiohttp
essions. `ModelManager` has `start()`/`stop()` and `__aenter__`/`__aexit__` that are
ever used.

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,246 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class RailsManager:
+    """Orchestrates input and output safety checks for IORails.
+
+    Reads the rails configuration to determine which checks are enabled,
+    then runs them using ModelManager for all LLM/safety calls.
+    Input rails run concurrently via asyncio.gather().
```
Pouyanpi, line 40: Incorrect claim saying: "Input rails run concurrently via asyncio.gather()". The implementation is
sequential:

see `nemoguardrails/guardrails/rails_manager.py:68-71`

```python
for flow in self.input_flows:
    result = await self._run_input_rail(flow, messages)
    if not result.is_safe:
        return result
```

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,246 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Rails manager for IORails engine.
+
+Orchestrates input/output safety checks by calling ModelManager.
+Input rails run concurrently via asyncio.gather().
```
Pouyanpi, line 19: Incorrect claim saying: "Input rails run concurrently via asyncio.gather()". The implementation is
sequential:

see `nemoguardrails/guardrails/rails_manager.py:68-71`

```python
for flow in self.input_flows:
    result = await self._run_input_rail(flow, messages)
    if not result.is_safe:
        return result

```
same as RailsManager docstring below

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,246 @@
...
+        bot_response: str = "",
+    ) -> str:
+        """Look up a prompt template by task key and render the prompt."""
+        prompt_template = self.prompts.get(prompt_key)
+        if not prompt_template or not prompt_template.content:
+            raise RuntimeError(f"No prompt template found for key {prompt_key}")
+
+        content = prompt_template.content
+        content = content.replace("{{ user_input }}", user_input)
+        content = content.replace("{{ bot_response }}", bot_response)
```
Pouyanpi: we use Jinja2 why not using the real engine? naive template rendering is vulnerable to injection

nemoguardrails/guardrails/iorails.py
```diff
@@ -0,0 +1,66 @@
...
+import logging
+
+from nemoguardrails.guardrails.guardrails_types import LLMMessage, LLMMessages
+from nemoguardrails.guardrails.model_manager import ModelManager
+from nemoguardrails.guardrails.rails_manager import RailsManager
+from nemoguardrails.rails.llm.config import RailsConfig
+
+log = logging.getLogger(__name__)
+
+REFUSAL_MESSAGE = "I'm sorry, I can't help you with that."
```
Pouyanpi: nit: not configurable. better to make it same as "bot refuse to respond" ie. "I'm sorry I cannot respond to that"

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,246 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Pouyanpi, line 1: `nemoguardrails/guardrails/rails_manager.py:130-132` : errors caught, returned as `RailResult(is_safe=False)` (fail closed)
`nemoguardrails/guardrails/rails_manager.py:102-103` : unsupported flow raises `RuntimeError` (crashes the request = fail open)

The strategy should be consistent and documented.

pyproject.toml
```diff
@@ -9,28 +9,28 @@ packages = [{ include = "nemoguardrails" }]
...
-  "LICENCES-3rd-party.txt",
-  "chat-ui/**/*",
-  "examples/**/*",
-  "eval/data/**/*",
-  "**/*.yml",
-  "**/*.co",
-  "**/*.txt",
-  "**/*.json",
-  "**/*.lark",
+    "LICENSE.md",
```
Pouyanpi: `pyproject.toml` we have ~80 lines of pure whitespace reformatting  unrelated
to the feature. Should be reverted.

Pouyanpi (commented):
The review is still in progress.

nemoguardrails/guardrails/guardrails.py
tgasser-nv: Yes, that's addressed in #1649 which is stacked on top of this PR

nemoguardrails/guardrails/guardrails.py
tgasser-nv: Yes, that's addressed in #1649

nemoguardrails/guardrails/iorails.py
tgasser-nv: Updated this to be consistent. In a later PR I'll make this configurable.

nemoguardrails/guardrails/model_engine.py
tgasser-nv: The empty string is intentional, if you set an env var that's an empty string that should be the same as not setting the environment variable for our usecase (API Keys)

nemoguardrails/guardrails/model_engine.py
tgasser-nv: Yes I fixed this in #1649 (see https://github.com/NVIDIA-NeMo/Guardrails/pull/1649/changes#diff-669bb5c2b285ae4e557fde5bd30cad4bfcc884111c81ebd692c6e2e93d9bc5d2R181)

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: These strategies should not be consistent because they're catching different cases.

* Returning RailResult(is_safe=False) is failing closed intentionally. If we get a dynamic exception thrown from  a model (e.g. due to a network issue) then we have to assume the input content-safety failed (unless we have evidence to the contrary).
* The Unsupported flow is a static configuration issue from the `config.yml` file, not a dynamic issue like an API call failing. Raising an exception here gives the user early feedback with the offending flow name rather than silently swallowing it and giving a vague "Internal Server Error" response at inference-time

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: Fixed in #1649 (see https://github.com/NVIDIA-NeMo/Guardrails/pull/1649/changes#diff-d3c77a18cfd98fcea40a677ee9be08c6cc6649ce11b7a4b4a7279a35020e7d66R19)

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: Fixed in #1649  (this line was removed)

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: Updated to use jinja2 in latest commit

pyproject.toml
tgasser-nv: Reverted it in latest commit

nemoguardrails/guardrails/rails_manager.py
Pouyanpi, line 1: this module does 6 different things:

1. flow routing (`_run_input_rail`, `_run_output_rail`)
2. prompt rendering (`_render_prompt`)
3. response parsing (_parse_content_safety_*)
4. flow name parsing (`_flow_to_prompt_key`, `_flow_name`, `_flow_model_type`)
5. message extraction (`_last_user_content`, `_last_content_by_role`)
6. LLM invocation (`generate_async` calls in `_check_content_safety_*`)

When topic safety and jailbreak detection are added, this class will balloon with more
`elif` branches and more parsers and more prompt templates.

**one way** to improve it is to extract a `Rail` strategy per rail type (I cannot think of a better name right now). each rail (content safety, topic safety, jailbreak) owns its prompt, parser, and model type. `RailsManager` becomes a simple loop over registered rails. 

Another approach would be the idea of LLMTaskManager that we currently have in the codebase. 

Anyhow these responsibilities do not seem to belong to the RailsManager.
(more context below)

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+class RailsManager:
+    """Orchestrates input and output safety checks for IORails.
+
+    Reads the rails configuration to determine which checks are enabled,
+    then runs them using ModelManager for all LLM/safety calls.
+    Input rails run concurrently via asyncio.gather().
+    """
+
+    def __init__(self, config: RailsConfig, model_manager: ModelManager) -> None:
+        self.config = config
```
Pouyanpi, line 44: related to the other comments below:

here RailsManager receives config and model_manager and we are doing config parsing in a runtime class. We can see it as config parsing in constructor coupling 

`RailsManager` receives `RailsConfig` and immediately tears it apart in `__init__` and affter `__init__`, `self.config` is never referenced again. The class stores the entire config but only uses the 3 extracted fields. 

To me this looks like lazy initialization of the wrong thing, it should receive the extracted data, not the container. I have noticed similar pattern in other classes where RailsConfig or part of it is passed and stored as instance variable. In general I recommend not to couple the new architecture on the details of RailsConfig. For Guardrails and to some extent IORails it makes sense but for the rest better to avoid it.

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+            raise RuntimeError(f"Input rail flow `{base_flow}` not supported")
+
+    async def _run_output_rail(self, flow: str, messages: list[dict], response: str) -> RailResult:
+        """Run an output rail flow if it's supported. If not raise an exception"""
+        base_flow = flow.split("$")[0].strip()
+
+        if base_flow == "content safety check output":
+            return await self._check_content_safety_output(flow, messages, response)
+        else:
+            raise RuntimeError(f"Output rail flow `{base_flow}` not supported")
```
Pouyanpi, line 117: Today this is 1 branch each. When topic-safety and jailbreak-detection are added:

```python
if base_flow == "content safety check input":
    return await self._check_content_safety_input(flow, messages)
elif base_flow == "topic safety check input":
    return await self._check_topic_safety_input(flow, messages)
elif base_flow == "jailbreak detection model":
    return await self._check_jailbreak_detection_input(flow, messages)
else:
    raise RuntimeError(...)
```
and in addition there are going to be more methods like  `_check_content_safety_input` etc.

Every branch has the same shape: extract data from flow string, render prompt,
call model, parse response, return RailResult. The only things that vary are:

- which prompt template
- which model type
- which response parser

that's a "strategy".

`RailsManager` (or just `IORails` directly) becomes:

```python
for rail in self._input_rails:
    result = await rail.check(messages)
    if not result.is_safe:
        return result
```

no if/else no string matching, and adding a new rail type is equivalent to adding a new class, touching
zero existing code. 

Again this is only one way that might avoid RailsManager to have multiple non-cohesive responsibilities.

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+        self.model_manager = model_manager
+
+        # Store prompts keyed by task name for easy lookup
+        self.prompts: dict[str, TaskPrompt] = {}
+        if config.prompts:
+            self.prompts = {prompt.task: prompt for prompt in config.prompts}
+
+        # Determine which input/output rails are enabled
+        self.input_flows: list[str] = list(config.rails.input.flows)
+        self.output_flows: list[str] = list(config.rails.output.flows)
```
Pouyanpi, line 54: better not to depend on the primitives that the new architecture intend to remove in future. I think having a Rail strategy should resolve these sort of problems (see https://github.com/NVIDIA-NeMo/Guardrails/pull/1638/changes#r2833872213)

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+        """Convert a flow name to the corresponding prompt task key.
+
+        Flow names use spaces, prompt task keys use underscores:
+          'content safety check input $model=content_safety'
+          -> 'content_safety_check_input $model=content_safety'
+        """
+        if "$" in flow:
+            base, param = flow.split("$", 1)
+            return base.strip().replace(" ", "_") + " $" + param
+        return flow.replace(" ", "_")
```
Pouyanpi, line 184: as mentioned above: this is a very detailed knowledge/assumption that `RailsManager` knows about and I think it should not. If strategy pattern with `Rail` protocol is not an option then something like a `FlowSpec` might come handy. It also avoids the fragile nature of such stringified primitives that current `LLMRails` counterpart also suffer from.

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+    @staticmethod
+    def _flow_name(flow: str) -> str:
+        """Extract the model name from a flow.
+        For example:
+           'content safety check input $model=content_safety' -> 'content safety check input'
+           'self check input' -> 'self check input'
+        """
+        if "$model=" in flow:
+            return flow.split("$model=")[0].strip()
+        return flow
```
Pouyanpi, line 195: it is a dead code , please remove it.

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+
+    @staticmethod
```
Pouyanpi, line 197: this is very specific to one class (pattern) of flows (rails), I think such detail should not be evident to RailsManager. RailsManager can return it if needed but the responsibility should be delegated to a lower level entity.

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+        for message in reversed(messages):
+            message_role = message.get("role")
+            if message_role and message_role == role:
+                message_content = message.get("content")
+                if message_content:
+                    return message_content
+
+        raise RuntimeError(f"No {role}-role content in messages: {messages}")
+
+    def _last_user_content(self, messages: LLMMessages) -> str:
```
Pouyanpi, line 216: shouldn't be an instance method

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+        """Return the last entry in messages list with role set to `user`"""
+        return self._last_content_by_role(messages, "user")
+
+    def _parse_content_safety_input_response(self, response: str) -> RailResult:
```
Pouyanpi, line 220: shouldn't be an instance method

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+        """Use the existing `nemoguard_parse_prompt_safety` method and convert to RailResult."""
+
+        result = nemoguard_parse_prompt_safety(response)
+        rail_result = self._parse_content_safety_result(result)
+        return rail_result
+
+    def _parse_content_safety_output_response(self, response: str) -> RailResult:
```
Pouyanpi, line 227: shouldn't be an instance method

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -0,0 +1,251 @@
...
+        """Use the existing `nemoguard_parse_response_safety` method and convert to RailResult."""
+
+        result = nemoguard_parse_response_safety(response)
+        rail_result = self._parse_content_safety_result(result)
+        return rail_result
+
+    def _parse_content_safety_result(self, result: Sequence[bool | str]) -> RailResult:
```
Pouyanpi, line 234: shouldn't be an instance method

nemoguardrails/guardrails/iorails.py
```diff
@@ -0,0 +1,66 @@
...
+log = logging.getLogger(__name__)
+
+REFUSAL_MESSAGE = "I'm sorry, I can't respond to that."
+
+
+class IORails:
+    """Workflow engine for accelerated Input/Output rails inference."""
+
+    def __init__(self, config: RailsConfig) -> None:
+        self.config = config
```
Pouyanpi: we have passthrough coupling here (data that passes through a layer without being used.). `IORails` receives `RailsConfig` solely to pass it to `ModelManager` and `RailsManager` (on a different note, I think those two should not be coupled to RailsConfig).
It stores `self.config` but never reads from it.

nemoguardrails/guardrails/iorails.py
```diff
@@ -0,0 +1,66 @@
...
+
+        # Model Manager has one or more ModelEngine inside. Each ModelEngine calls a single model or API
+        self.model_manager = ModelManager(config)
+
+        # Rails Manager is responsible for running rails by making calls to Model Manager
+        self.rails_manager = RailsManager(config, self.model_manager)
```
Pouyanpi, line 44: preferably (a more DIP aligned design) `IORails` should `ModelManager` and `RailsManager` as constructor parameters with default, or receive only the narrow data it needs. It should not know about `RailsConfig` at all.

nemoguardrails/guardrails/iorails.py
```diff
@@ -0,0 +1,66 @@
...
+
+    async def generate_async(self, messages: LLMMessages, **kwargs) -> LLMMessage:
```
Pouyanpi, line 46: nice, reads like a pipeline.

nemoguardrails/guardrails/iorails.py
```diff
@@ -0,0 +1,66 @@
...
+        """Run input rails, generation, and output rails. Return response if safe."""
+
+        # Step 1: Check input rails
+        input_result = await self.rails_manager.is_input_safe(messages)
+        if not input_result.is_safe:
+            log.info("Input blocked: %s", input_result.reason)
+            return {"role": "assistant", "content": REFUSAL_MESSAGE}
```
Pouyanpi, line 53: nit: potential refactoring opportunity the same two field structure appears three times. a small factory function like `assistant_message(content: str) -> LLMMessage` would reduce duplication.

nemoguardrails/guardrails/guardrails.py
Pouyanpi, line 1: `Guardrails` routes between `IORails` and `LLMRails` but they share no protocol. This forces `isinstance` checks (in #1649) or boolean flags (in #1638) at every method, plus `NotImplementedError` guards on 4 of 5 public methods.

I recommend to define a `RailsEngine` **Protocol** with `generate_async` (also `generate`?). Methods only available on `LLMRails` (`stream_async`,`explain`, `update_llm`) should not appear on the shared interface at all. (update_llm should've been pirvate in LLMRails). Then `Guardrails` holds a `RailsEngine` and only calls methods on the shared protocol. Methods unique to `LLMRails` (`stream_async`, `explain`, `update_llm`) are accessed via `guardrails.llmrails` directly and not through the polymorphic interface.

It should also help with the identical overloads

nemoguardrails/guardrails/guardrails.py
```diff
@@ -100,13 +110,37 @@ def _convert_to_messages(prompt: str | None = None, messages: LLMMessages | None
...
+            if flow_type not in IORAILS_INPUT_FLOWS:
+                return False
+
+        for flow in self.config.rails.output.flows:
+            flow_type = flow.split("$")[0].strip()
+            if flow_type not in IORAILS_OUTPUT_FLOWS:
+                return False
+
+        return True
+
```
Pouyanpi, line 130: ideally Guardrails class should not know these details. In other words, this `_has_only_iorails_flows` method encodes domain logic (understanding what flows means and which engine supports them) in the composition root. Better to move it to `IORails` itself (e.g., `IORails.can_handle(config) -> bool`).

nemoguardrails/guardrails/guardrails.py
```diff
@@ -50,7 +52,10 @@ class MessageRole(str, Enum):
     TOOL = "tool"
```
Pouyanpi: dead code or speculative generality.

nemoguardrails/guardrails/guardrails.py
```diff
@@ -61,14 +66,19 @@ def __init__(
...
 
         self.config = config
-        self.llm = llm
         self.verbose = verbose
 
+        # Whether to use IORailsEngine for inference requests
+        self._use_iorails_engine: bool = use_iorails and self._has_only_iorails_flows()
+        self.iorails = IORails(config)
         self.llmrails = LLMRails(config, llm, verbose)
+        self.rails_engine = self.iorails if self._use_iorails_engine else self.llmrails
```
Pouyanpi: the constructor sets `self.config`, `self.iorails`, `self.llmrails`, and `self.rails_engine` as public attributes.
these should preferably be immutable once created and these should be prefixed with underscore (so that we know their nature).

nemoguardrails/guardrails/model_manager.py
```diff
@@ -0,0 +1,92 @@
...
+
+class ModelManager:
+    """Manages ModelEngine instances for IORails.
+
+    Creates one ModelEngine per configured model, keyed by model type
+    (e.g. "main", "content_safety", "jailbreak_detection").
+    Each engine owns its own HTTP client with per-model retry and timeout settings.
+    """
+
+    def __init__(self, config: RailsConfig) -> None:
```
Pouyanpi: this has already been mentioned that should be avoided. shotgun surgery maybe where `ModelManager` receives `RailsConfig` (30+ fields) but reads only `.models`.

nemoguardrails/guardrails/model_manager.py
```diff
@@ -0,0 +1,92 @@
...
+
+        Raises:
+            KeyError: If no model with the given type is configured.
+        """
+        if model_type not in self._engines:
+            available = list(self._engines.keys())
+            raise KeyError(f"No model configured with type '{model_type}'. Available types: {available}")
+        return self._engines[model_type]
+
+    async def generate_async(self, model_type: str, messages: list[dict], **kwargs: Any) -> str:
```
Pouyanpi, line 72: is it supposed to follow similar protocol as LLMRails and IORails? or this shared method name is accidental?

nemoguardrails/guardrails/model_manager.py
```diff
@@ -0,0 +1,92 @@
...
+            model_type: The model type key (e.g. "main", "content_safety").
+            messages: List of message dicts in OpenAI format.
+            **kwargs: Additional LLM parameters.
+
+        Returns:
+            The content string from the model's response.
+        """
+        engine = self.get_engine(model_type)
+        response = await engine.call(messages, **kwargs)
+        return response["choices"][0]["message"]["content"]
```
Pouyanpi, line 85: OpenAI specific response parsing hardcoded in the manager/adapter layer. This parsing logic belongs in `ModelEngine` itself, which is already the adapter for the HTTP layer.

aditionally when moved to its proper class it should be wrapped in a try/except with an actionable error message.

nemoguardrails/guardrails/model_manager.py
Pouyanpi, line 1: this class looks more like a ModelPool or registry. If possible let's not use manager as it does not convey much about the responsibilities/intents. It currently acts as registry/proxy over `ModelEngine` instances. 
 
`ModelManager` and `RailsManager` are  [name smells](https://blog.codinghorror.com/i-shall-call-it-somethingmanager/ ) because of Manager (to some extent this holds for Engine).

Pouyanpi (commented):
more high level comments

nemoguardrails/guardrails/guardrails.py
tgasser-nv: I'd prefer to use a base class to enforce method signatures. But in any case this can be added later. The issue is that LLMRails has poor encapsulation and exposes several methods which are implementation-specific to Colang.

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: It's not dead code. Either runs depending on whether `$model` is in the flow

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: RailsManager is responsible for running rails, which contain one or more flows. How is parsing flows out-of-scope ?

nemoguardrails/guardrails/guardrails.py
Pouyanpi: Sure, we must fix those concerns in LLMRails 👍🏻

Pouyanpi (approved):
Thank you Tim, let's keep these comments as unresolved so that we can get back to them later. Feel free to merge if it is OK with you.

nemoguardrails/guardrails/model_manager.py
tgasser-nv: Yes originally the aim was to pass through the same method signature to show we're proxying the main LLM request. I don't have a strong preference on keeping it, and with the API-calls we also support it might be better to rename

nemoguardrails/guardrails/guardrails.py
tgasser-nv: Removed this code

nemoguardrails/guardrails/guardrails.py
tgasser-nv: Prefixed them with underscores, and created @property getters with the non-underscore prefix

nemoguardrails/guardrails/iorails.py
tgasser-nv: Removed self.config here

nemoguardrails/guardrails/model_manager.py
tgasser-nv: Yes I changed the signature to receive a `list[Models]` which is the only part of RailsConfig the model manager cares about

tgasser-nv:
@greptile review this PR

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/iorails.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/iorails.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/guardrails.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/guardrails.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/model_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/model_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/model_manager.py
tgasser-nv: @tgasser-nv Return to this when refactoring post-v0.21 release

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: #1762 addresses this by having each action inherit from a RailAction base-class. Common methods to all actions are implemented in RailAction. Action-specific steps override the abstract methods and allow per-action customization.

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: #1762 The RailAction Abstract Base Class addresses this. RailsManager just iterates over actions all the action-specific code is in derived classes from RailAction.

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: In #1762 I create an LLMTaskManager inside RailsManager which needs the top-level RailsConfig at __init__() time. The other sub-models inside RailsConfig are unpacked as needed.

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: We need RailsConfig to instantiate an LLMTaskManager inside RailsManager. Is that the primitive you mean? Or you're planning to remove flows? @Pouyanpi

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: This is handled in the RailAction base class in #1762

nemoguardrails/guardrails/model_manager.py
tgasser-nv: Fixed in #1778 (now have `model_call()` and `api_call()` in ModelEngine

nemoguardrails/guardrails/model_manager.py
tgasser-nv: Fixed in #1778 . Parsing is done inside ModelEngine

nemoguardrails/guardrails/model_manager.py
tgasser-nv: Fixed in #1778 . Renamed ModelManager to EngineRegistry

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: Resolved in #1778

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: Resolved in #1778

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: Resolved in #1778

nemoguardrails/guardrails/iorails.py
tgasser-nv: Resolved in #1778

nemoguardrails/guardrails/iorails.py
tgasser-nv: Resolved in #1778

nemoguardrails/guardrails/rails_manager.py
tgasser-nv: Resolved in #1778

nemoguardrails/guardrails/guardrails.py
tgasser-nv: Addressing in #1893

PR #1666, opened 2026-02-25
fix(logging): show cache hits in Stats log and fix duplicate metadata restore

Pouyanpi:
## Description

- add `cache_hits` counter to `LLMStats` so the processing summary shows how many calls were served from cache, e.g. `Stats: 5 total calls (3 from cache), ...`
- rename `LLM Stats:` label to `Stats:` in `generate_async` to align with `generate_events_async` and `process_events_async` which already use `Stats:`
- fix copy-paste bug in `get_from_cache_and_restore_stats` where `restore_llm_metadata_from_cache` was called 3 times instead of once

## Test plan

**Unit tests:**

Save the following script (cache_stats_display.py) and run it against the `nemoguards_cache` config:

```python
import asyncio
import logging

from nemoguardrails import LLMRails, RailsConfig

logging.basicConfig(level=logging.INFO)

config = RailsConfig.from_path("./examples/configs/nemoguards_cache")
rails = LLMRails(config=config, verbose=True)

caches = rails.runtime.registered_action_params.get("model_caches", {})
messages = [{"role": "user", "content": "What is the company policy on paid time off?"}]


async def main():
    print("\n========== REQUEST 1 (cold cache) ==========\n")
    await rails.generate_async(messages=messages)
    for name, cache in caches.items():
        print(f"  {name}: {cache.get_stats()}")

    print("\n========== REQUEST 2 (warm cache) ==========\n")
    await rails.generate_async(messages=messages)
    for name, cache in caches.items():
        print(f"  {name}: {cache.get_stats()}")


asyncio.run(main())
```

Then filter for the Stats line:
```
poetry run python cache_stats_display.py 2>&1 | grep "Stats:"
```
**Expected output:**
```
14:13:42.597 | Total processing took 6.35 seconds. Stats: 4 total calls, 6.08 total time, 2734 total tokens, 2333 total prompt tokens, 401 total completion tokens, [0.51, 0.29, 4.85, 0.43] as latencies
14:13:48.738 | Total processing took 6.14 seconds. Stats: 4 total calls (2 from cache), 6.07 total time, 2898 total tokens, 2415 total prompt tokens, 483 total completion tokens, [0.0, 0.0, 5.49, 0.59] as latencies
```

### DIFF
--- a/nemoguardrails/llm/cache/utils.py
+++ b/nemoguardrails/llm/cache/utils.py
@@ -107,6 +107,7 @@ def restore_llm_stats_from_cache(cached_stats: LLMStatsDict, cache_read_duration
         llm_stats_var.set(llm_stats)
 
     llm_stats.inc("total_calls")
+    llm_stats.inc("cache_hits")
     llm_stats.inc("total_time", cache_read_duration_s)
     llm_stats.inc("total_tokens", cached_stats.get("total_tokens", 0))
     llm_stats.inc("total_prompt_tokens", cached_stats.get("prompt_tokens", 0))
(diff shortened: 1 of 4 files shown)


tgasser-nv (approved):
Looks great, the new logging is far clearer about which responses are returned from cache

PR #1668, opened 2026-02-25
docs: add redirects

miyoungc:
## Description

Found increased number of traffic to "page not found" in adobe analytics. Hence, adding redirects.

## Checklist


tgasser-nv (approved):
Looks good, especially with the high number of "page not found" pageviews from our analytics. Is there any validation / linting step that ensures the destination we're redirecting to exists? Or did you have to check these manually?

miyoungc:
> Looks good, especially with the high number of "page not found" pageviews from our analytics. Is there any validation / linting step that ensures the destination we're redirecting to exists? Or did you have to check these manually?

I asked claude in Cursor to create scripts to correctly form the redirects. And also I passed the before and after the doc revamp URLs to make sure they map correctly. We need to activate the build break upon any warnings to prevent any broken link issues. We should be able to add linting rule too.

PR #1685, opened 2026-03-04
fix(streaming): handle None stop tokens in streaming handler

Pouyanpi:
## Description

- LangChain LLMs commonly default `stop=None`. `_stream_llm_call` assigned this `None` to `handler.stop`, causing `TypeError: 'NoneType' object is not iterable` in `StreamingHandler._process()` when iterating `self.stop`.
- Add a property setter on `StreamingHandler.stop` that coerces `None` to `[]`, so the handler owns its own invariant instead of relying on callers to sanitize.

Root cause: `dict.get(key, default)` and `getattr(obj, attr, default)` only use the default when the key/attr is missing, not when it's explicitly `None`.

Introduced in: 2231475 

Repro: `rails.stream_async(messages=[...])` with any default LangChain LLM (e.g. gpt-4o-mini) that has `stop=None`. e.g. "examples/bots/hello_world"

### DIFF
--- a/nemoguardrails/streaming.py
+++ b/nemoguardrails/streaming.py
@@ -16,7 +16,7 @@
 import asyncio
 import logging
 import warnings
-from typing import Any, AsyncIterator, Dict, Optional, Union
+from typing import Any, AsyncIterator, Dict, List, Optional, Union
 
 from nemoguardrails.utils import new_uuid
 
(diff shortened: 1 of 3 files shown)


trebedea (approved):
Looks good, I think this is the correct way to treat stop. 
In the previous implementation, we were defaulting to sending `stop=[]` which is not the actual default param value for `stop` for most LLMs in Langchain.

PR #1687, opened 2026-03-05
fix(streaming): handle dict chunks in RollingBuffer.format_chunks

Pouyanpi:
## Summary

- **Bug**: `RollingBuffer` crashes with `TypeError` when `LLMRails.stream_async(include_metadata=True)` is used with output rails streaming enabled. The `StreamingHandler` yields dict chunks (`{"text": ..., "metadata": ...}`) but `format_chunks()` calls `"".join()` which cannot join dicts. Affects both direct `LLMRails` and `RunnableRails` usage.
- **Fix**: Normalize dict chunks to strings in `process_stream()` at the input boundary, preserving `List[str]` typing throughout the buffer internals
- **Tests**: Add tests for `process_stream` with dict chunks and mixed string/dict chunks

### DIFF
--- a/nemoguardrails/rails/llm/buffer.py
+++ b/nemoguardrails/rails/llm/buffer.py
@@ -292,7 +292,7 @@ class RollingBuffer(BufferStrategy):
         total_chunks = 0
 
         async for chunk in streaming_handler:
-            buffer.append(chunk)
+            buffer.append(chunk["text"] if isinstance(chunk, dict) else chunk)
             total_chunks += 1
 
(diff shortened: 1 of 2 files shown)


tgasser-nv (approved):
Looks good, how did this bug surface recently ..  Is this the first time we ran `stream_async()` with include_metadata=True? Or because it only occurs when we instantiate LLMRails directly and not through the server API?

Before merging, can you run a local integration test which fails before applying the fix and passes after the fix is added?

PR #1688, opened 2026-03-05
fix(server): prioritize env var API key over forwarded client header

Pouyanpi:
- Fix auth key priority in `fetch_models()` so the server's env var API key (e.g. `OPENAI_API_KEY`) is used first, falling back to the forwarded client `Authorization` header only when no env var is set
- before this PR, for OpenAI-compatible providers, the client's `Authorization` header was forwarded directly to the upstream API, bypassing the server's configured credentials
- this caused `test_list_models_openai` and `test_list_models_openai_fields` to fail when `OPENAI_API_KEY` was set, because the test client's dummy key was sent upstream instead of the real key from the env var

Test plan: 

run
```bash
poetry run pytest tests/server/test_openai_integration.py
```

Fails before this fix (you must have server deps and OPENAI_API_KEY set)

### DIFF
--- a/nemoguardrails/server/schemas/utils.py
+++ b/nemoguardrails/server/schemas/utils.py
@@ -140,16 +140,11 @@ async def fetch_models(
     headers: Dict[str, str] = {}
     forwarded = request_headers.get("Authorization", "")
 
-    # Check if the auth header is a Bearer token
-    if forwarded and auth_header_name == "Authorization":
-        headers["Authorization"] = forwarded
-    else:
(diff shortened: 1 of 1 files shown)


Pouyanpi:
@christinaexyou would you please have a look at this PR and kindly add your review. Just want to make sure that it is not breaking any part of your implementation. Thanks!

christinaexyou (approved):

christinaexyou:
@Pouyanpi looks good to me ! Thank you :)

PR #1690, opened 2026-03-05
chore(ci): enable Greptile auto-review on every push

Pouyanpi:
## Description
- Enable `triggerOnUpdates` in `greptile.json` so code reviews are triggered automatically on every push, not just on PR creation
- Devs were having to manually comment `@greptileai` after each commit to get a re-review


tgasser-nv (approved):
LGTM

PR #1714, opened 2026-03-11
fix(middleware): handle MODIFIED status in GuardrailsMiddleware instead of silently dropping it

Pouyanpi:
…

## Description

- Fix `GuardrailsMiddleware` silently dropping content modifications when rails return `MODIFIED` status
- Input rails now replace the last user message with sanitized content via `_replace_last_human_message`
- Output rails now replace the last AI message with sanitized content (reuses existing `_replace_last_ai_message`)
- Add early return guard in `abefore_model` when no user message exists (mirrors `aafter_model` pattern)
- Update docs to reflect the new behavior (previously documented as "by design")

## Summary by CodeRabbit

* **New Features**
  * Message content replacement now supported when rails detect modifications, enabling use cases like PII redaction and content sanitization for both input and output messages.

* **Documentation**
  * Updated middleware documentation to specify exact content replacement behavior.

* **Tests**
  * Added comprehensive test suite covering input/output message replacement scenarios.


tgasser-nv (approved):
Looks good, thanks for the last-minute fix

PR #1745, opened 2026-03-25
feat(types): add framework-agnostic LLM type system

Pouyanpi:
1. **stack-1: canonical types** (this PR)

## Description

Introduce nemoguardrails/types.py with provider-independent data types and protocols that replace direct LangChain type dependencies in core code.

- Add `nemoguardrails/types.py` with framework-agnostic data types (`ChatMessage`, `Role`, `ToolCall`, `ToolCallFunction`, `LLMResponse`, `LLMResponseChunk`, `UsageInfo`, `FinishReason`) and protocols (`LLMModel`, `LLMFramework`)
- `ChatMessage.from_dict()` handles both OpenAI nested and legacy flat tool call formats, JSON string argument parsing, and role aliases (`bot`, `human`, `developer`)
- `ChatMessage.to_dict()` / `from_dict()` roundtrip preserves all fields including `provider_metadata`
- `LLMModel` protocol defines the `generate()` / `stream()` contract that all LLM adapters must implement
- `LLMFramework` protocol defines the `create_model()` factory contract for pluggable backends

### DIFF
--- a/nemoguardrails/types.py
+++ b/nemoguardrails/types.py
@@ -261,16 +261,14 @@ class LLMFramework(Protocol):
     Each framework (LangChain, LiteLLM, etc.) implements this protocol to
     provide a factory for creating ``LLMModel`` instances.
 
-    ``mode`` and ``model_kwargs`` are optional. Not all frameworks need them.
-    LangChain uses ``mode`` to choose between chat and text completion
-    models; other frameworks may ignore it. Implementations should handle
-    ``None`` values with sensible defaults.
+    ``model_kwargs`` carries all provider-specific configuration. Framework
+    implementations extract what they need (e.g. LangChain pops ``mode``
+    to choose between chat and text completion models).
     """
 
     def create_model(
         self,
         model_name: str,
         provider_name: str,
-        mode: Optional[str] = None,
         model_kwargs: Optional[Dict[str, Any]] = None,
     ) -> LLMModel: ...
--- a/nemoguardrails/types.py
+++ b/nemoguardrails/types.py
@@ -156,6 +156,10 @@ class ChatMessage:
                             f"Tool call arguments must be a JSON object, got {type(args_dict).__name__}: {raw_args!r}"
                         )
                 else:
+                    if not isinstance(raw_args, dict):
+                        raise ValueError(
+                            f"Tool call arguments must be a dict, got {type(raw_args).__name__}: {raw_args!r}"
+                        )
                     args_dict = raw_args
 
                 tool_calls.append(
--- a/nemoguardrails/types.py
+++ b/nemoguardrails/types.py
@@ -105,13 +105,13 @@ class ChatMessage:
         if self.content is not None:
             payload["content"] = self.content
 
-        if self.tool_calls:
+        if self.tool_calls is not None:
             payload["tool_calls"] = [tc.to_dict() for tc in self.tool_calls]
 
-        if self.tool_call_id:
+        if self.tool_call_id is not None:
             payload["tool_call_id"] = self.tool_call_id
 
-        if self.name:
+        if self.name is not None:
             payload["name"] = self.name
 
         if self.provider_metadata:
@@ -131,7 +131,9 @@ class ChatMessage:
         captured into ``provider_metadata``.
         """
 
-        raw_role = d.get("role", "user")
+        raw_role = d.get("role")
+        if raw_role is None:
+            raise ValueError("Missing required key: 'role'")
         role = _ROLE_ALIASES.get(raw_role)
         if role is None:
             raise ValueError(f"Unknown role: {raw_role}")
--- a/nemoguardrails/types.py
+++ b/nemoguardrails/types.py
@@ -61,6 +61,8 @@ class UsageInfo:
 FinishReason = Literal["stop", "length", "tool_calls", "content_filter", "error", "other"]
 
 
+_STANDARD_MESSAGE_KEYS = {"role", "content", "tool_calls", "tool_call_id", "name", "provider_metadata"}
+
 _ROLE_ALIASES = {
     "bot": Role.ASSISTANT,
     "assistant": Role.ASSISTANT,
@@ -173,8 +175,7 @@ class ChatMessage:
                     )
                 )
 
-        _standard_keys = {"role", "content", "tool_calls", "tool_call_id", "name", "provider_metadata"}
-        extra = {k: v for k, v in d.items() if k not in _standard_keys}
+        extra = {k: v for k, v in d.items() if k not in _STANDARD_MESSAGE_KEYS}
         provider_metadata = {**d.get("provider_metadata", {}), **extra}
 
         return cls(
@@ -239,7 +240,9 @@ class LLMModel(Protocol):
         *,
         stop: Optional[List[str]] = None,
         **kwargs,
-    ) -> AsyncIterator["LLMResponseChunk"]: ...
+    ) -> AsyncIterator["LLMResponseChunk"]:
+        """Implementations must be async generator functions (use ``yield``)."""
+        ...
 
     @property
     def model_name(self) -> str: ...
@@ -258,7 +261,7 @@ class LLMFramework(Protocol):
     Each framework (LangChain, LiteLLM, etc.) implements this protocol to
     provide a factory for creating ``LLMModel`` instances.
 
-    ``mode`` and ``kwargs`` are optional. Not all frameworks need them.
+    ``mode`` and ``model_kwargs`` are optional. Not all frameworks need them.
     LangChain uses ``mode`` to choose between chat and text completion
     models; other frameworks may ignore it. Implementations should handle
     ``None`` values with sensible defaults.
@@ -269,5 +272,5 @@ class LLMFramework(Protocol):
         model_name: str,
         provider_name: str,
         mode: Optional[str] = None,
-        kwargs: Optional[Dict[str, Any]] = None,
+        model_kwargs: Optional[Dict[str, Any]] = None,
     ) -> LLMModel: ...
--- a/nemoguardrails/types.py
+++ b/nemoguardrails/types.py
@@ -140,7 +140,7 @@ class ChatMessage:
 
         tool_calls = None
         raw_tool_calls = d.get("tool_calls")
-        if raw_tool_calls:
+        if raw_tool_calls is not None:
             tool_calls = []
             for tc in raw_tool_calls:
                 func_data = tc.get("function")
--- a/nemoguardrails/types.py
+++ b/nemoguardrails/types.py
@@ -151,6 +151,10 @@ class ChatMessage:
                         args_dict = json.loads(raw_args)
                     except json.JSONDecodeError:
                         raise ValueError(f"Tool call arguments are not valid JSON: {raw_args!r}")
+                    if not isinstance(args_dict, dict):
+                        raise ValueError(
+                            f"Tool call arguments must be a JSON object, got {type(args_dict).__name__}: {raw_args!r}"
+                        )
                 else:
                     args_dict = raw_args
 
@@ -165,13 +169,17 @@ class ChatMessage:
                     )
                 )
 
+        _standard_keys = {"role", "content", "tool_calls", "tool_call_id", "name", "provider_metadata"}
+        extra = {k: v for k, v in d.items() if k not in _standard_keys}
+        provider_metadata = {**d.get("provider_metadata", {}), **extra}
+
         return cls(
             role=role,
             content=d.get("content"),
             tool_calls=tool_calls,
             tool_call_id=d.get("tool_call_id"),
             name=d.get("name"),
-            provider_metadata=d.get("provider_metadata", {}),
+            provider_metadata=provider_metadata,
         )
 
 
@@ -221,7 +229,7 @@ class LLMModel(Protocol):
         **kwargs,
     ) -> "LLMResponse": ...
 
-    async def stream(
+    def stream(
         self,
         prompt: Union[str, List["ChatMessage"]],
         *,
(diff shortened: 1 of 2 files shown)


nemoguardrails/types.py
```diff
@@ -0,0 +1,261 @@
...
+            content=d.get("content"),
+            tool_calls=tool_calls,
+            tool_call_id=d.get("tool_call_id"),
+            name=d.get("name"),
+            provider_metadata=d.get("provider_metadata", {}),
+        )
+
+
+@dataclass
+class LLMResponse:
```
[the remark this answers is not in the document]
Pouyanpi: Valid point semantically, but the codebase currently relies on content always being a str. Multiple call sites do string operations directly on it without null checks (e.g., "<think>" in response.content, response.content.strip(), re.sub(..., response.content)). Changing to Optional[str] would require auditing and updating all downstream consumers.
For now, content="" serves as the "no content" sentinel and is consistent with how the rest of the pipeline treats it.

nemoguardrails/types.py
```diff
@@ -0,0 +1,276 @@
...
+        self,
+        prompt: Union[str, List["ChatMessage"]],
+        *,
+        stop: Optional[List[str]] = None,
+        **kwargs,
+    ) -> AsyncIterator["LLMResponseChunk"]:
+        """Implementations must be async generator functions (use ``yield``)."""
+        ...
+
+    @property
```
tgasser-nv, line 249: Curious why these properties are different to the [Model](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/nemoguardrails/rails/llm/config.py#L103) `type`, `engine`, `model`?

nemoguardrails/types.py
```diff
@@ -0,0 +1,276 @@
...
+    Adapters wrap provider-specific SDKs (LangChain, LiteLLM, OpenAI, etc.)
+    behind this interface so the core pipeline remains framework-agnostic.
+
+    ``prompt`` accepts either a plain string or a list of ``ChatMessage``
+    objects. Adapters convert ``ChatMessage`` to whatever their SDK expects.
+    ``**kwargs`` are forwarded to the underlying SDK (e.g. temperature,
+    max_tokens).
+    """
+
+    async def generate(
```
tgasser-nv, line 231: Will we replace [`generate_async()`](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/nemoguardrails/rails/llm/llmrails.py#L733) on LLMRails with this signature?

nemoguardrails/types.py
```diff
@@ -0,0 +1,276 @@
...
+        self,
+        prompt: Union[str, List["ChatMessage"]],
+        *,
+        stop: Optional[List[str]] = None,
+        **kwargs,
+    ) -> "LLMResponse": ...
+
+    def stream(
```
tgasser-nv, line 239: Same as above, will the `stream_async()` on LLMRails be replaced with this?

tests/test_types.py
```diff
@@ -0,0 +1,467 @@
...
+        assert chunk.usage.total_tokens == 5
+
+
+class TestLLMModelProtocol:
+    def test_mock_satisfies_protocol(self):
+        class MockLLM:
+            async def generate(self, prompt, *, stop=None, **kwargs):
+                return LLMResponse(content="response")
+
+            async def stream(self, prompt, *, stop=None, **kwargs):
```
tgasser-nv, line 421: Shouldn't this be a sync method to match the LLMModel? With a Protocol it'll still work but this should match

nemoguardrails/types.py
```diff
@@ -0,0 +1,276 @@
...
+                else:
+                    if not isinstance(raw_args, dict):
+                        raise ValueError(
+                            f"Tool call arguments must be a dict, got {type(raw_args).__name__}: {raw_args!r}"
+                        )
+                    args_dict = raw_args
+
+                tool_calls.append(
+                    ToolCall(
+                        id=tc.get("id", ""),
```
tgasser-nv, line 171: Should we create a UUID if none is included?

nemoguardrails/types.py
```diff
@@ -0,0 +1,276 @@
...
+        raw_role = d.get("role", "user")
+        role = _ROLE_ALIASES.get(raw_role)
+        if role is None:
+            raise ValueError(f"Unknown role: {raw_role}")
+
+        tool_calls = None
+        raw_tool_calls = d.get("tool_calls")
+        if raw_tool_calls:
+            tool_calls = []
+            for tc in raw_tool_calls:
```
tgasser-nv, line 145: Nit: Rename to raw_tool_call rather than tc?

nemoguardrails/types.py
Pouyanpi: No, this shouldn’t be conflated with that. this will be used in `llm_call`. Once the logging callback handler refactor is merged, I’ll open the remaining PRs in this stack to make the overall direction clearer.

In the meantime, feel free to consult the RFC where the usage is more clear there. I’ll also rename this to generate_async for consistency. If you have better naming suggestions, I’d really appreciate you sharing them here.

nemoguardrails/types.py
Pouyanpi: no ☝🏻

tests/test_types.py
Pouyanpi: This one is a bit tricky 


in a protocol, async def and def tell the caller how to use the method:

```python
# Protocol says async def -> caller must await
async def generate(...) -> LLMResponse: ...
# and in the caller side: response = await llm.generate(prompt)

# Protocol uses def -> caller gets the result directly from calling it
def stream(...) -> AsyncIterator[LLMResponseChunk]: ...
#  caller side: async for chunk in llm.stream(prompt):
```

now, an async generator (which is how stream gets implemented):

```python
async def stream(self, prompt, **kwargs):
    yield LLMResponseChunk(...)
    yield LLMResponseChunk(...)

```
when we call llm.stream(prompt), it does not return a coroutine. It returns an `AsyncIterator` directly. we never await it. and that's the key: even though the implementation uses async def, the calling convention is not await. So the Protocol must use `def` because from the caller's pov, we call it and get an iterator back, no await involved.

nemoguardrails/types.py
Pouyanpi: Good idea, but I am not still 100 about making it a required field. Let's keep it unresolved I'll get back to it.

nemoguardrails/types.py
Pouyanpi: I found these cleaner and closer to the domain so people can easily understand and implement them. Also type is not relevant here.
TBH, I never liked engine and model in ModelConfig but that is a different concern.

nemoguardrails/types.py
Pouyanpi: Ok now it is more clear, we should make it required and raise ValueError if missing, rather than generating a UUID. The adapter layer generates UUIDs for provider responses that omit IDs, but from_dict() receives user/upstream data that should already have them 👍🏻 I've opened a PR to make this change to avoid rebase effect.

Pouyanpi:
thanks @tgasser-nv, all the comments are addressed/resolved. For the tool.id i've opened this PR: https://github.com/NVIDIA-NeMo/Guardrails/pull/1775

nemoguardrails/types.py
tgasser-nv: We'll have to maintain the Model fields forever to maintain backwards compatibility. How do you map between those and the new fields? I'd rather have a consistent approach throughout than lookups from one to the other

tgasser-nv (commented):
Looks good. Only concern is the impedance mismatch between the `Model` pydantic object and `provider_name`, `prodivder_url`. Why do we need a second slightly different way of representing the same thing? Are you intending to deprecate Model in the future?

Pouyanpi:
Thanks @tgasser-nv . re your question I see it this way: `Model` is the user facing config contract. LLMModel is the adapter author contract (Python protocol which we use internally and adapter authors could implement it). 

provider_name/model_name/provider_url are explicit for someone implementing an adapter and conform to our existing internals: llm_call() parameters (model_name, model_provider), LLMCallInfo fields (llm_model_name, llm_provider_name), LLMCallException error context (model=, provider=, endpoint=), and OpenTelemetry tracing spans (GenAIAttributes.GEN_AI_PROVIDER_NAME). The protocol surfaces what the codebase already names this way.

**The mapping happens in exactly one place (llmrails.py, `init_llm_model(model_name=main_model.model, provider_name=main_model.engine, kwargs=kwargs)`). There's no ongoing translation or lookup. config is read once at startup, the adapter is created, and from that point only LLMModel is used.**

When you review the other PRs in the stack it becomes clearer how these fit together, but if it's still a point of debate we can rename in a follow-up PR before the release.

nemoguardrails/types.py
Pouyanpi: ref: https://github.com/NVIDIA-NeMo/Guardrails/pull/1745#issuecomment-4222721431

nemoguardrails/types.py
```diff
@@ -0,0 +1,274 @@
...
+        return cls(role=Role.SYSTEM, content=content, **kwargs)
+
+    @classmethod
+    def from_tool(cls, content: str, tool_call_id: str, **kwargs) -> "ChatMessage":
+        return cls(role=Role.TOOL, content=content, tool_call_id=tool_call_id, **kwargs)
+
+    def to_dict(self) -> Dict[str, Any]:
+        payload: Dict[str, Any] = {"role": self.role.value}
+
+        if self.content is not None:
```
tgasser-nv, line 105: Recommend using `is not None` for all the attributes in `to_dict()` (content, tool_calls, tool_call_id, name, and provider_metadata). If not, a key with an empty string or list will be silently dropped. This means if we do a roundtrip to dict (`to_dict()` then `from_dict()`) we won't get back what we started with.

For example with a truthiness check (e.g. `if self.tool_calls`) rather than `if self.tool_calls is not None` an empty string will drop `tool_calls` from the output dict. When we run `from_dict()` on the returned dict, the `tool_calls` attribute key will disappear. This doesn't match the ChatMessage we called `to_dict()` on.

nemoguardrails/types.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Accepts both the canonical nested tool call format
+        (``{"function": {"name": ..., "arguments": ...}}``) and the legacy
+        flat format (``{"name": ..., "args": ...}``). JSON string arguments
+        are parsed automatically. Role aliases like "bot", "human", and
+        "developer" are mapped to canonical Role values. Unknown keys are
+        captured into ``provider_metadata``.
+        """
+
+        raw_role = d.get("role", "user")
```
tgasser-nv: Recommend removing the `user` fallback value. Otherwise dicts with no `role` key automatically default to user rather than raising an exception. Invalid roles will fire the ValueError on line 137.

```python
raw_role = d.get("role")
if raw_role is None:
    raise ValueError("ChatMessage dict missing required 'role' key")
```

nemoguardrails/types.py
```diff
@@ -0,0 +1,274 @@
...
+    delta_tool_calls: Optional[List[ToolCall]] = None
+    model: Optional[str] = None
+    finish_reason: Optional[FinishReason] = None
+    request_id: Optional[str] = None
+    usage: Optional[UsageInfo] = None
+    provider_metadata: Optional[Dict[str, Any]] = None
+
+
+@runtime_checkable
+class LLMModel(Protocol):
```
tgasser-nv, line 219: Do you think we'd ever have an LLMModel that's non-streaming only, or streaming-only? I'm thinking of Nemoguard models where streaming doesn't make sense for their output. They'd have to raise a NotImplementedError for `stream()`. I don't know if the overhead of having two Protocols is worth it though.

tests/test_types.py
```diff
@@ -0,0 +1,467 @@
...
+        resp = LLMResponse(content="hello")
+        assert resp.content == "hello"
+        assert resp.tool_calls is None
+        assert resp.reasoning is None
+        assert resp.model is None
+        assert resp.finish_reason is None
+        assert resp.usage is None
+        assert resp.provider_metadata is None
+
+    def test_creation_full(self):
```
tgasser-nv, line 376: nit: The `stop_sequence` and `request_id` fields are missing if the intent is to check object creation with all fields.

nemoguardrails/types.py
```diff
@@ -0,0 +1,274 @@
...
+            content=d.get("content"),
+            tool_calls=tool_calls,
+            tool_call_id=d.get("tool_call_id"),
+            name=d.get("name"),
+            provider_metadata=provider_metadata,
+        )
+
+
+@dataclass
+class LLMResponse:
```
tgasser-nv, line 194: Not blocking for this PR, but do we need `to-dict()` and `from_dict()` for this dataclass (similar to ChatMessage) for logging/serialization?

tgasser-nv (approved):
Looks good! Please take a look at the feedback, only a few small cleanups needed before meging

nemoguardrails/types.py
Pouyanpi: don't think so

tests/test_types.py
Pouyanpi: added 👍🏻

nemoguardrails/types.py
Pouyanpi: good catch 👍🏻

nemoguardrails/types.py
Pouyanpi: Not really. currently `llm_call` decides which method to invoke. one Protocol, implementors should decide what to do for unsupported method(s), raise error etc.

Pouyanpi:
a final note about engine vs provider: we can have openai `engine` with a `nim` , `vllm` or a litellm proxy server.

PR #1755, opened 2026-04-01
fix(embeddings): persist in-memory embedding cache instance across calls

Pouyanpi:
## Description

The @cache_embeddings decorator created a new EmbeddingsCache via from_config() on every invocation. For InMemoryCacheStore this meant a fresh empty dict each time, so the cache never had hits. Cache the EmbeddingsCache instance on self so it persists across calls.


tgasser-nv (approved):
LGTM, approving

PR #1759, opened 2026-04-02
feat(llm): add LangChain adapter and framework registry

Pouyanpi:
2. **stack-2: adapter and framework registry** (this PR)

## Description

- `LangChainLLMAdapter` wrapping `BaseChatModel` behind `LLMModel` protocol
- `LangChainFramework` implementing `create_model()`
- Framework registry with `register_framework()` / `get_framework()` / `set_default_framework()`
- `chatmessage_to_langchain_message()` and `chatmessages_to_langchain_messages()` conversion utils
- `provider_url` extracts endpoint from LangChain LLM attributes

Pure additive. No existing files modified except `message_utils.py`.

## Release Notes

* **New Features**
  * Added LangChain integration enabling the use of LangChain-based language models
  * Support for both streaming and generation workflows
  * Introduced framework registry system for managing multiple LLM frameworks
  * Enhanced support for reasoning models with automatic parameter optimization

* **Tests**
  * Added comprehensive test coverage for new LangChain integration and framework features

### DIFF
--- a/nemoguardrails/integrations/langchain/llm_adapter.py
+++ b/nemoguardrails/integrations/langchain/llm_adapter.py
@@ -31,6 +31,16 @@ from nemoguardrails.types import (
 log = logging.getLogger(__name__)
 
 
+def _flatten_content(raw: Any) -> Optional[str]:
+    if raw is None:
+        return None
+    if isinstance(raw, str):
+        return raw
+    if isinstance(raw, list):
+        return "".join(block.get("text", "") if isinstance(block, dict) else str(block) for block in raw)
+    return str(raw)
+
+
 def _infer_model_name(llm: Any):
     """Helper to infer the model name based from an LLM instance.
 
@@ -362,7 +372,7 @@ def _build_provider_metadata(
 
 
 def _langchain_response_to_llm_response(response: Any) -> LLMResponse:
-    content = getattr(response, "content", None)
+    content = _flatten_content(getattr(response, "content", None))
     if content is None:
         content = str(response)
 
@@ -384,9 +394,9 @@ def _langchain_response_to_llm_response(response: Any) -> LLMResponse:
 
 
 def _langchain_chunk_to_llm_response_chunk(chunk: Any) -> LLMResponseChunk:
-    content = getattr(chunk, "content", None)
+    content = _flatten_content(getattr(chunk, "content", None))
     if content is None:
-        content = getattr(chunk, "text", None)
+        content = _flatten_content(getattr(chunk, "text", None))
     if content is None:
         content = str(chunk)
 
--- a/nemoguardrails/integrations/langchain/llm_adapter.py
+++ b/nemoguardrails/integrations/langchain/llm_adapter.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import uuid
 from typing import Any, AsyncIterator, Dict, List, Optional, Union
 
@@ -26,6 +27,8 @@ from nemoguardrails.types import (
     ToolCallFunction,
     UsageInfo,
 )
+
+log = logging.getLogger(__name__)
 
 
 def _infer_model_name(llm: Any):
@@ -41,7 +44,7 @@ def _infer_model_name(llm: Any):
                 return val
 
     model_kwargs = getattr(llm, "model_kwargs", None)
-    if model_kwargs and isinstance(model_kwargs, Dict):
+    if model_kwargs and isinstance(model_kwargs, dict):
         for attr in ["model", "model_name", "name"]:
             val = model_kwargs.get(attr)
             if isinstance(val, str):
@@ -77,9 +80,7 @@ def _infer_provider_from_module(llm: Any) -> Optional[str]:
 
         if provider == "community":
             parts = module.split(".")
-            if len(parts) >= 3:
-                provider = parts[-1]
-                return provider
+            return parts[-1] if len(parts) >= 3 else "community"
         else:
             return provider
 
@@ -91,9 +92,7 @@ def _infer_provider_from_module(llm: Any) -> Optional[str]:
 
             if provider == "community":
                 parts = base_module.split(".")
-                if len(parts) >= 3:
-                    provider = parts[-1]
-                    return provider
+                return parts[-1] if len(parts) >= 3 else "community"
             else:
                 return provider
 
@@ -156,6 +155,7 @@ class LangChainLLMAdapter:
         if is_openai_reasoning_model:
             filtered = params.copy()
             filtered.pop("temperature", None)
+            log.debug("Stripped 'temperature' for reasoning model '%s'", model_name)
             return filtered
 
         return params
@@ -251,10 +251,13 @@ def _build_usage_info(raw: Any) -> Optional[UsageInfo]:
             return None
     if not raw:
         return None
+    input_tokens = raw.get("input_tokens", raw.get("prompt_tokens", 0))
+    output_tokens = raw.get("output_tokens", raw.get("completion_tokens", 0))
+    total_tokens = raw.get("total_tokens") or (input_tokens + output_tokens)
     return UsageInfo(
-        input_tokens=raw.get("input_tokens", raw.get("prompt_tokens", 0)),
-        output_tokens=raw.get("output_tokens", raw.get("completion_tokens", 0)),
-        total_tokens=raw.get("total_tokens", 0),
+        input_tokens=input_tokens,
+        output_tokens=output_tokens,
+        total_tokens=total_tokens,
         reasoning_tokens=raw.get("reasoning_tokens"),
         cached_tokens=raw.get("cached_tokens", raw.get("cache_read_input_tokens")),
     )
--- a/nemoguardrails/integrations/langchain/llm_adapter.py
+++ b/nemoguardrails/integrations/langchain/llm_adapter.py
@@ -263,7 +263,24 @@ def _build_usage_info(raw: Any) -> Optional[UsageInfo]:
     )
 
 
-def _extract_reasoning(response) -> Optional[str]:
+_EXTRACTED_METADATA_KEYS = frozenset(
+    {
+        "model_name",
+        "model",
+        "finish_reason",
+        "stop_reason",
+        "stop_sequence",
+        "id",
+        "request_id",
+        "token_usage",
+        "usage",
+    }
+)
+
+_REASONING_KEYS = frozenset({"reasoning_content"})
+
+
+def _extract_reasoning(response: Any) -> Optional[str]:
     content_blocks = getattr(response, "content_blocks", None)
     if content_blocks:
         for block in content_blocks:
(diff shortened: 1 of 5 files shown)


nemoguardrails/integrations/langchain/llm_adapter.py
```diff
@@ -0,0 +1,403 @@
...
+    LLMResponseChunk,
+    ToolCall,
+    ToolCallFunction,
+    UsageInfo,
+)
+
+log = logging.getLogger(__name__)
+
+
+def _infer_model_name(llm: Any):
```
Pouyanpi, line 44: this is moved from actions/llm/utils.py

nemoguardrails/integrations/langchain/llm_adapter.py
```diff
@@ -0,0 +1,403 @@
...
+class LangChainFramework:
+    def create_model(
+        self,
+        model_name: str,
+        provider_name: str,
+        model_kwargs: Optional[Dict[str, Any]] = None,
+    ) -> LLMModel:
+        from nemoguardrails.llm.models.langchain_initializer import (
+            init_langchain_model,
+        )
```
Pouyanpi, line 223: should we move langchain_initializer from models?

nemoguardrails/integrations/langchain/llm_adapter.py
```diff
@@ -0,0 +1,394 @@
...
+
+    response_metadata = getattr(chunk, "response_metadata", None) or {}
+    generation_info = getattr(chunk, "generation_info", None) or {}
+    merged_metadata = {**response_metadata, **generation_info}
+
+    return LLMResponseChunk(
+        delta_content=content,
+        usage=_extract_usage(chunk),
+        provider_metadata=merged_metadata or None,
+    )
```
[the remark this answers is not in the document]
tgasser-nv: This looks blocking to merging, and the helpers already exist to implement this

nemoguardrails/integrations/langchain/llm_adapter.py
```diff
@@ -0,0 +1,394 @@
...
+    def _filter_reasoning_model_params(self, params: Optional[dict]) -> Optional[dict]:
+        if not params or "temperature" not in params:
+            return params
+
+        model_name = _infer_model_name(self._llm).lower()
+
+        is_openai_reasoning_model = (
+            model_name.startswith("o1")
+            or model_name.startswith("o3")
+            or (model_name.startswith("gpt-5") and "chat" not in model_name)
```
tgasser-nv, line 162: nit: I don't think OpenAI are using chat suffixes on their models. They seem to use sizes instead (mini, nano, etc). This is brittle but with no clear convention from OpenAI on model-naming for reasoning models it's the best we can do

tgasser-nv (approved):
Looks good, please check the comments. Also can you review the code coverage and add more unit-tests to get the patch coverage up to 100%?

nemoguardrails/integrations/langchain/message_utils.py
```diff
@@ -147,6 +147,36 @@ def is_base_message(obj: Any) -> bool:
     return isinstance(obj, BaseMessage)
 
 
+def chatmessage_to_langchain_message(msg: "ChatMessage") -> BaseMessage:
```
tgasser-nv, line 150: Should the ChatMessage `provider_metadata` be passed through to relevant fields on Langchain messages? It's dropped currently

nemoguardrails/integrations/langchain/message_utils.py
```diff
@@ -147,6 +147,36 @@ def is_base_message(obj: Any) -> bool:
...
+    from nemoguardrails.types import Role
+
+    content = msg.content or ""
```
tgasser-nv, line 153: Are there any issues with converting msg.content == None to "" here? Do we need to treat a missing value differently to an empty string

nemoguardrails/integrations/langchain/llm_adapter.py
```diff
@@ -0,0 +1,394 @@
...
+    return None
+
+
+def _extract_model_info(response_metadata: Dict[str, Any]) -> tuple:
+    model = response_metadata.get("model_name") or response_metadata.get("model")
+    raw_finish = response_metadata.get("finish_reason") or response_metadata.get("stop_reason")
+    finish_reason = _map_finish_reason(raw_finish)
+    stop_sequence = response_metadata.get("stop_sequence")
+    request_id = response_metadata.get("id") or response_metadata.get("request_id")
+    return model, finish_reason, stop_sequence, request_id
```
tgasser-nv: This function is doing too much, returning a 4-tuple whose ordering is fragile and prone to mistakes at the call site. Either refactor into separate functions or use a NamedTuple as return type

nemoguardrails/llm/frameworks.py
```diff
@@ -0,0 +1,60 @@
...
+        raise ValueError(f"Framework '{name}' is already registered.")
+    _frameworks[name] = framework
+
+
+def get_framework(name: str) -> LLMFramework:
+    if name not in _frameworks:
+        if name == "langchain":
+            from nemoguardrails.integrations.langchain.llm_adapter import LangChainFramework
+
+            _frameworks["langchain"] = LangChainFramework()
```
tgasser-nv, line 36: Maybe not needed with Python's asyncio, but two threads could check and then create LangChainFrameworks here at the same time. The fix would be to wrap in a threading.Lock

tests/test_langchain_llm_adapter.py
```diff
@@ -0,0 +1,348 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tgasser-nv, line 1: Can you add unit-tests that test `chatmessage_to_langchain_message` and `chatmessages_to_langchain_messages` in messages_utils.py? They're tested indirectly here (test_langchain_llm_adapter.py) but may be used elsewhere in future

nemoguardrails/integrations/langchain/llm_adapter.py
```diff
@@ -0,0 +1,394 @@
...
+    async def generate(
+        self,
+        prompt: Union[str, List[ChatMessage]],
+        *,
+        stop: Optional[List[str]] = None,
+        **kwargs,
+    ) -> LLMResponse:
+        llm = self._prepare_llm(kwargs)
+        messages = self._to_langchain_input(prompt)
+        response = await llm.ainvoke(messages, stop=stop)
```
tgasser-nv, line 198: No exception catching here (is it implemented in a later PR?)

nemoguardrails/integrations/langchain/llm_adapter.py
```diff
@@ -0,0 +1,394 @@
...
+    async def stream(
+        self,
+        prompt: Union[str, List[ChatMessage]],
+        *,
+        stop: Optional[List[str]] = None,
+        **kwargs,
+    ) -> AsyncIterator[LLMResponseChunk]:
+        llm = self._prepare_llm(kwargs)
+        messages = self._to_langchain_input(prompt)
+        async for chunk in llm.astream(messages, stop=stop):
```
tgasser-nv, line 210: No exception handling here, is it implemented in a later PR?

tgasser-nv (commented):
Added feedback to be addressed before merging

nemoguardrails/integrations/langchain/llm_adapter.py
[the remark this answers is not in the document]
Pouyanpi: Done

nemoguardrails/integrations/langchain/message_utils.py
Pouyanpi: `chatmessage_to_langchain_message` is only used on the input side (preparing messages to send to the llm). provider_metadata is populated on the response side, where it's already captured in ▎ `_langchain_response_to_llm_response` and `_langchain_chunk_to_llm_response_chunk` via `_build_provider_metadata`. no round trip needed here.

nemoguardrails/integrations/langchain/message_utils.py
Pouyanpi: no, langchain's `BaseMessage(content=...)` expects str, not None.  "" fallback is intentional and it covers tool call assistant messages where content is absent.

nemoguardrails/integrations/langchain/llm_adapter.py
Pouyanpi: You're right :+1 made the change

nemoguardrails/llm/frameworks.py
Pouyanpi: I agree get_framework runs in asyncio context  which is single threaded. I think in worst case scenario with threads is going to be a harmless duplicate instantiation of the stateless LangChainFramework. skipping the Lock for now

nemoguardrails/integrations/langchain/llm_adapter.py
Pouyanpi: yes exception handling is done in the caller (`llm_call` / `_stream_llm_call` in utils.py), not in the adapter. stack-3 already has this. the adapter intentionally stays thin and lets exceptions propagate.

nemoguardrails/integrations/langchain/llm_adapter.py
Pouyanpi: yes exception handling is done in the caller (`llm_call` / `_stream_llm_call` in utils.py), not in the adapter. stack-3 already has this. the adapter intentionally stays thin and lets exceptions propagate.

nemoguardrails/integrations/langchain/llm_adapter.py
Pouyanpi: Yes please see https://github.com/NVIDIA-NeMo/Guardrails/pull/1526#discussion_r2661936332 for reference to this issue.

nemoguardrails/llm/frameworks.py
```diff
@@ -0,0 +1,60 @@
...
+# limitations under the License.
+
+import os
+from typing import Dict
+
+from nemoguardrails.types import LLMFramework
+
+_frameworks: Dict[str, LLMFramework] = {}
+_default_framework: str = os.environ.get("NEMOGUARDRAILS_LLM_FRAMEWORK", "langchain")
+
```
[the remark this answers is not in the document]
Pouyanpi: The env var sets _default_framework directly. But _reset_frameworks() does the same. If someone sets NEMOGUARDRAILS_LLM_FRAMEWORK=garbage, get_framework() will fail later. the env var is meant for legitimate custom frameworks that get registered at startup. validating eagerly at import time would break that flow since the framework isn't registered yet. I'd skip this one.

nemoguardrails/integrations/langchain/llm_adapter.py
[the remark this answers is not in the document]
Pouyanpi: done, only happens under some conditions for ChatAnthropic that requires user passing parameters in the config.

Pouyanpi:
Thanks @tgasser-nv . I addressed all the comments 👍🏻

PR #1760, opened 2026-04-02
refactor(llm)!: atomic switch to LLMModel protocol

Pouyanpi:
3. **stack-3: pipeline rewrite + caller migration** (this PR)

## Summary

Atomic switch from LangChain to the LLMModel protocol. These changes are tightly coupled (you cannot change what the initializer returns without also changing how utils.py calls it unless bridge stacks are introduced which is a mess, and we cannot change the return type without updating all callers), so they land as one PR.

## How to review

This PR is split into 5 scoped commits for easier review. Please review commit-by-commit rather than the full diff:

1. **`rewrite initializer and llmrails for LLMModel protocol`** 
`init_llm_model()` delegates to framework registry, `LLMRails` types `self.llm` as `LLMModel` with legacy wrapping
2. **`rewrite llm_call to use LLMModel.generate/stream`**
   the core pipeline change, removes all LangChain extraction logic
3. **`update integration boundary for canonical tool calls`**
   `tool_calls_to_langchain_format()` at output boundary
4. **`migrate all callers to LLMResponse.content`** 
   mechanical one-line changes across 35 call sites
5. **`restore endpoint URL in LLMCallException via provider_url`** -
   uses `LLMModel.provider_url` instead of inspecting LLM attributes

Individual commits do not pass tests on their own since the changes are interdependent. The PR as a whole passes the full test suite.

## Summary by CodeRabbit

* **New Features**
  * Introduced framework abstraction layer enabling support for multiple LLM providers beyond LangChain, with new public APIs: `get_default_framework()`, `register_framework()`, and `set_default_framework()`.

* **Refactor**
  * Replaced LangChain-specific LLM types with unified `LLMModel` interface throughout the codebase. Added `LangChainLLMAdapter` for backward compatibility with existing LangChain implementations.

### DIFF
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -277,12 +277,15 @@ def _raise_llm_call_exception(
     llm_call_info = llm_call_info_var.get()
     model_name = llm_call_info.llm_model_name if llm_call_info else model.model_name
     provider_name = llm_call_info.llm_provider_name if llm_call_info else model.provider_name
+    endpoint_url = model.provider_url
 
     context_parts = []
     if model_name:
         context_parts.append(f"model={model_name}")
     if provider_name:
         context_parts.append(f"provider={provider_name}")
+    if endpoint_url:
+        context_parts.append(f"endpoint={endpoint_url}")
 
     if context_parts:
         detail = f"Error invoking LLM ({', '.join(context_parts)})"
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -233,9 +233,9 @@ def _update_token_stats(response: LLMResponse) -> None:
         llm_call_info.completion_tokens = 0
 
     if response.usage:
-        llm_call_info.total_tokens += response.usage.total_tokens
-        llm_call_info.prompt_tokens += response.usage.input_tokens
-        llm_call_info.completion_tokens += response.usage.output_tokens
+        llm_call_info.total_tokens = response.usage.total_tokens
+        llm_call_info.prompt_tokens = response.usage.input_tokens
+        llm_call_info.completion_tokens = response.usage.output_tokens
 
         if llm_stats:
             llm_stats.inc("total_tokens", response.usage.total_tokens)
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -82,9 +82,9 @@ async def llm_call(
     except Exception as e:
         _raise_llm_call_exception(e, model)
 
+    _store_reasoning_traces(response)
     _log_completion(response)
     _update_token_stats(response)
-    _store_reasoning_traces(response)
     _store_tool_calls(response)
     _store_response_metadata(response)
     return response
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -15,10 +15,7 @@
 
 import logging
 import re
-from typing import TYPE_CHECKING, Any, Dict, List, NoReturn, Optional, Union
-
-from langchain_core.language_models import BaseLanguageModel
-from langchain_core.runnables.base import Runnable
+from typing import TYPE_CHECKING, Any, Dict, List, NoReturn, Optional, Union, cast
 
 from nemoguardrails.colang.v2_x.lang.colang_ast import Flow
 from nemoguardrails.colang.v2_x.runtime.flows import InternalEvent, InternalEvents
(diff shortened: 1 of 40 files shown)


nemoguardrails/actions/llm/utils.py
```diff
@@ -30,249 +27,84 @@
...
+        response: LLMResponse = await model.generate(chat_prompt, stop=stop, **(llm_params or {}))
+    except Exception as e:
+        _raise_llm_call_exception(e, model)
+
+    _log_completion(response)
+    _update_token_stats(response)
+    _store_reasoning_traces(response)
+    _store_tool_calls(response)
+    _store_response_metadata(response)
+    return response
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai this is already fixed! _store_reasoning_traces is at line 85, before _log_completion at line 86. This was done in the review feedback commit (393943023).

nemoguardrails/rails/llm/llmrails.py
```diff
@@ -118,17 +118,34 @@
...
+            "Either install langchain or pass an LLMModel instance."
+        )
+    warnings.warn(
+        "Passing a raw LangChain LLM is deprecated. "
+        "Use LangChainLLMAdapter(llm) explicitly or pass an LLMModel instance.",
+        DeprecationWarning,
+        stacklevel=3,
+    )
+    return LangChainLLMAdapter(llm)
+
```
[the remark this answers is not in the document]
Pouyanpi: pre-existing issue

tests/test_tool_output_rails.py
```diff
@@ -32,7 +32,8 @@ async def validate_tool_parameters(tool_calls, context=None, **kwargs):
     dangerous_patterns = ["eval", "exec", "system", "../", "rm -", "DROP", "DELETE"]
 
     for tool_call in tool_calls:
-        args = tool_call.get("args", {})
+        func = tool_call.get("function", {})
+        args = func.get("arguments", {})
```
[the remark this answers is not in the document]
Pouyanpi: pre-existing and out of scope

nemoguardrails/actions/llm/utils.py
```diff
@@ -292,30 +124,38 @@ async def _stream_llm_call(
         if last_chunk is not None:
             _update_token_stats_from_chunk(last_chunk)
 
-        return handler.completion
+        return LLMResponse(
```
tgasser-nv, line 127: Should this be `LLMResponseChunk` rather than `LLMResponse`? And the fields should be completed as below?

```
@dataclass
class LLMResponseChunk:
    delta_content: Optional[str] = None
    delta_reasoning: Optional[str] = None
    delta_tool_calls: Optional[List[ToolCall]] = None
    model: Optional[str] = None
    finish_reason: Optional[FinishReason] = None
    request_id: Optional[str] = None
    usage: Optional[UsageInfo] = None
    provider_metadata: Optional[Dict[str, Any]] = None
```

nemoguardrails/actions/llm/utils.py
```diff
@@ -398,51 +232,20 @@ def _update_token_stats(response) -> None:
...
-        token_stats_found = True
-        usage = response.usage_metadata
-        total = usage.get("total_tokens", 0)
-        input_tokens = usage.get("input_tokens", 0)
-        output_tokens = usage.get("output_tokens", 0)
-
-        llm_call_info.total_tokens = total
-        llm_call_info.prompt_tokens = input_tokens
-        llm_call_info.completion_tokens = output_tokens
+    if response.usage:
```
tgasser-nv, line 235: We're incrementing here by total/input/output tokens whereas before we set the values directly (i.e. no accumulation over chunks). Was this an uncaught bug previous to this PR?

nemoguardrails/actions/llm/utils.py
```diff
@@ -685,49 +339,20 @@ def _extract_and_remove_think_tags(response) -> Optional[str]:
...
-        return tool_calls if tool_calls else None
-    return None
-
-
-def _extract_tool_calls_from_attribute(response) -> List | None:
-    return getattr(response, "tool_calls", None)
-
+def _store_tool_calls(response: LLMResponse) -> None:
+    if response.tool_calls:
+        tool_calls_var.set([tc.to_dict() for tc in response.tool_calls])
```
tgasser-nv, line 344: nit: `tool_call` rather than `tc`?

nemoguardrails/actions/llm/utils.py
```diff
@@ -30,249 +27,84 @@
...
-        model_name: Optional model name for tracking
-        model_provider: Optional model provider for tracking
-        stop: Optional list of stop tokens
-        llm_params: Optional configuration dictionary to pass to the LLM (e.g., temperature, max_tokens)
-        streaming_handler: Optional StreamingHandler to receive streaming chunks
-
-    Returns:
-        The generated text response
-    """
+) -> LLMResponse:
```
tgasser-nv, line 60: Moving from str to LLMResponse return-type means every call-site has to have a trailing `.content` added to remain compatible. This looks is a wide-ranging change and someone will forget to add it at some point. 

Recommend renaming this function to `llm_call_response()` or another name, and keeping llm_call() as a wrapper which calls the `llm_call_response()` function and pulls out the `content` field. It's backwards-compatible and call-sites can be migrated to use LLMResponse if it makes sense, or the content string directly

Pouyanpi: like your suggestion but this is intentional, all call sites must respect the new contract. It also helps with eliminating different hacks like context vars etc in different part of generation modules (something to do in future). Users will get a clear error and the change is straightforward. 

I'll emphasize this breaking change in changelog and release notes. To me this is a nice breaking change where we can easily inform users about the new contract that we've introduced with a straightforward migration.

nemoguardrails/actions/llm/utils.py
Pouyanpi: It should be LLMResponse for historic reasons, it is compatible with `handler.completion` before this change 😄 

Because _stream_llm_call is called from actions in generation modules. Those actions need the complete result to continue the pipeline for example run the next Colang flow step, store the bot response, execute tool calls, update context variables etc.

`StreamingHandler` txt out as it arrives, but it doesn't return a structured result to the caller. It's a push mechanism (fire and forget chunks). The action code needs a pull mechanism something like give me thefinal answer so I can decide what to do next.

I have a WIP that populates other fields in LLMResponse

nemoguardrails/actions/llm/utils.py
Pouyanpi: Good catch 👍🏻 thanks! the += is not needed, functionally += and = are equivalent here but unnecessary . historically, the `LoggingCallbackHandler` (removed in #1616) used += because it was iterating over response.generations and accumulated token stats across multiple generations within a single langchain response and not needed anymore. will change to = to match the streaming counterpart.

nemoguardrails/actions/llm/utils.py
Pouyanpi: @tgasser-nv made the change in https://github.com/NVIDIA-NeMo/Guardrails/pull/1760/commits/0135fd101c7ed231c8b8be5915a98ac692857582

nemoguardrails/actions/llm/utils.py
Pouyanpi: @tgasser-nv I've opened https://github.com/NVIDIA-NeMo/Guardrails/pull/1789 to address this issue. would you approve it with this PR? 

(if you approve it I'll merge after stack-7 is merged)

nemoguardrails/actions/llm/utils.py
Pouyanpi: I'll do these nits at the end 👍🏻

tgasser-nv (approved):
Thanks for fixing the accumulation change, the rest all looks good

PR #1769, opened 2026-04-07
refactor(llm): rename generate/stream to generate_async/stream_async

Pouyanpi:
4. **stack-4: rename generate/stream to generate_async/stream_async** (this PR)

## Description

Rename LLMModel protocol methods for clarity and consistency. Updates protocol, LangChainLLMAdapter, llm_call callers, and tests.

## Summary by CodeRabbit

* **Refactor**
  * Asynchronous LLM model methods renamed to explicitly indicate their async nature: `generate_async()` and `stream_async()` for improved API clarity.

### DIFF
--- a/nemoguardrails/actions/llm/utils.py
+++ b/nemoguardrails/actions/llm/utils.py
@@ -78,7 +78,7 @@ async def llm_call(
         return await _stream_llm_call(model, chat_prompt, streaming_handler, stop, llm_params)
 
     try:
-        response: LLMResponse = await model.generate(chat_prompt, stop=stop, **(llm_params or {}))
+        response: LLMResponse = await model.generate_async(chat_prompt, stop=stop, **(llm_params or {}))
     except Exception as e:
         _raise_llm_call_exception(e, model)
 
@@ -102,7 +102,7 @@ async def _stream_llm_call(
     last_chunk: Optional[LLMResponseChunk] = None
 
     try:
-        async for chunk in model.stream(prompt, stop=stop, **(llm_params or {})):
+        async for chunk in model.stream_async(prompt, stop=stop, **(llm_params or {})):
             last_chunk = chunk
             content = chunk.delta_content or ""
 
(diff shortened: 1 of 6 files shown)


tgasser-nv (approved):
It's a straightforward change and looks good. Updating the Protocol with different method names means existing code will silently stop matching it, but I don't think the Protocol is intended for external use.

I also checked some clients to see how they handle sync vs async versions for inference, none of them use a suffix

##  Anthropic

  - Sync: Anthropic() → client.messages.create(...)
  - Async: AsyncAnthropic() → await client.messages.create(...)
  - Same method names. No suffix.
  - github.com/anthropics/anthropic-sdk-python

##  OpenAI

  - Sync: OpenAI() → client.chat.completions.create(...)
  - Async: AsyncOpenAI() → await client.chat.completions.create(...)
  - Same method names. No suffix.
  - github.com/openai/openai-python

##  Google GenAI

  - Sync: client.models.generate_content(...)
  - Async: await client.aio.models.generate_content(...)
  - Same method names. Async accessed via .aio property.
  - github.com/googleapis/python-genai

Pouyanpi:
Thanks @tgasser-nv . I agree, I think we don't need to have sync alternatives in our public contract (as in https://github.com/NVIDIA-NeMo/Guardrails/pull/1605#discussion_r2764034002) then as an async only API we can drop all these async suffixes. I later changed my mind to be consistent with the rest of codebase (see https://github.com/NVIDIA-NeMo/Guardrails/pull/1745#discussion_r3015180039 for background). 

re use of suffix, those clients have both sync and async versions which makes polymorphism a clear winner, I think our API has adapted to langchain sync vs async distinction at method level, like ainvoke vs invoke etc.  
What do you think we should do forward?

tgasser-nv:
> Thanks @tgasser-nv . I agree, I think we don't need to have sync alternatives in our public contract (as in [#1605 (comment)](https://github.com/NVIDIA-NeMo/Guardrails/pull/1605#discussion_r2764034002)) then as an async only API we can drop all these async suffixes. I later changed my mind to be consistent with the rest of codebase (see [#1745 (comment)](https://github.com/NVIDIA-NeMo/Guardrails/pull/1745#discussion_r3015180039) for background).
> 
> re use of suffix, those clients have both sync and async versions which makes polymorphism a clear winner, I think our API has adapted to langchain sync vs async distinction at method level, like ainvoke vs invoke etc. What do you think we should do forward?

I think let's stick with what we've got. There would be a lot of disruption if we change the public entry-points for little gain

PR #1770, opened 2026-04-07
refactor(llm): remove LangChain imports from core modules

Pouyanpi:
Part of the LangChain decoupling stack:
1. stack-1: canonical types (#1745)
2. stack-2: adapter and framework registry (#1759)
3. stack-3: pipeline rewrite + caller migration (#1760)
4. stack-4: rename generate/stream to generate_async/stream_async (#1769)
5. **stack-5: remove LangChain imports from core modules** (this PR)
6. stack-6: move LangChain implementations into integrations/langchain/ (#1772)
7. stack-7: framework-owned provider registry (#1773)
8. stack-8: framework-agnostic test infrastructure (TODO)

## Description

- Remove PromptTemplate from hallucination actions (was a no-op)
- Replace Runnable isinstance check with duck-type ainvoke check in action_dispatcher
- Remove PromptTemplate and llm.bind().invoke() from evaluate_factcheck, use LLMModel.generate_async via event loop
- Make eval CLI LangChain cache a lazy optional import

## Summary by CodeRabbit

* **Bug Fixes**
  * LLM caching no longer causes startup failure if LangChain dependencies are unavailable; users receive a warning instead.

* **Refactor**
  * Reduced direct dependency on specific LangChain APIs for improved flexibility and maintainability.

PR #1772, opened 2026-04-07
refactor(llm): move LangChain implementations into integrations/langchain/

Pouyanpi:
6. **stack-6: move LangChain implementations into integrations/langchain/** (this PR)

## Description

Move providers, helpers, langchain_initializer, and safetools into integrations/langchain/. Keep llm/providers/__init__.py as a backwards compatible shim for public API (register_chat_provider, etc.). Update all test imports and patch paths to new locations.

The shim is temporary. Stack 7 will introduce a framework-agnostic provider registry in core where each framework (LangChain, etc.) registers its own providers. At that point the shim is replaced with the real generic API.

## Summary by CodeRabbit

* **Documentation**
  * Updated streaming configuration examples with corrected import paths.

* **Refactor**
  * Reorganized internal module structure for better code organization. Old import paths now display clear migration guidance.

* **Chores**
  * Updated copyright year information.


Pouyanpi:
@greptileai

these are pre-existing issues and not related to this PR and thus out of scope:



> P1 — _parse_version crashes on pre-release installs: int() is applied to every dot-split segment; version strings like \"1.0.0rc1\" or \"2.0.0.post1\" raise an uncaught ValueError that propagates out of init_langchain_model, blocking all model initialization for users on pre-release LangChain packages.
> P2 — shared kwargs dict mutated across initializer attempts in _update_model_kwargs; if _init_community_chat_models mutates and then fails, _init_text_completion_model sees the polluted dict.

nemoguardrails/integrations/langchain/actions/__init__.py
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tgasser-nv, line 1: nit: How did the copyright text end up changing from 2023 - 2026 to just 2026, I think we need to change it back

nemoguardrails/integrations/langchain/providers/huggingface/__init__.py
tgasser-nv, line 14: Somehow the contents of this file were lost in the move/rename:

```python
from .pipeline import HuggingFacePipelineCompatible
from .streamers import AsyncTextIteratorStreamer

__all__ = [
    "HuggingFacePipelineCompatible",
    "AsyncTextIteratorStreamer",
]
```

Without this, imports like this won't work:

```
from nemoguardrails.integrations.langchain.providers.huggingface import HuggingFacePipelineCompatible
```

nemoguardrails/llm/providers/__init__.py
```diff
@@ -13,7 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .providers import (
+# Re-export from new location for backwards compatibility.
```
tgasser-nv, line 16: Do you need to do the same shim approach for the other subdirectories?

1. nemoguardrails/llm/helpers.py (get_llm_instance_wrapper)-> nemoguardrails.integrations.langchain.helpers.
2. nemoguardrails/actions/langchain/__init__.py-> nemoguardrails.integrations.langchain.actions
3. nemoguardrails/actions/langchain/actions.py -> nemoguardrails.integrations.langchain.actions.actions
4. nemoguardrails/actions/langchain/safetools.py -> nemoguardrails.integrations.langchain.actions.safetools

tests/llm/models/test_langchain_init_scenarios.py
```diff
@@ -51,13 +51,13 @@
 
 import pytest
 
-from nemoguardrails.llm.models.langchain_initializer import (
+from nemoguardrails.integrations.langchain.langchain_initializer import (
```
tgasser-nv, line 54: Can you add a few tests that import from the shim to cover the shim imports? 

`from nemoguardrails.integrations.langchain.providers import` should be `from nemoguardrails.llm.providers import ...` to make sure the shim works

tgasser-nv (approved):
Looks good! Can you address the feedback before merging? There are issues with the header copyright, missing backwards-compatibility shims for llm.helpers, etc missing shim unit-tests and empty huggingface __init__..py

nemoguardrails/integrations/langchain/actions/__init__.py
Pouyanpi: for the pre-commit hook this is a new file and it auto updates the header. I'll see if we can avoid it but I don't think so.

Pouyanpi:
Thank you @tgasser-nv for catching these, all valid and great points 🚀 

  fixed:
  - restored __init__.py re-exports for providers/huggingface and providers/trtllm
  - added import error stubs at the old paths (llm/helpers.py, llm/providers/huggingface, llm/providers/trtllm) will raise ImportError with the new path instead of silently breaking. and no LangChain imported into core.
  - added providers shim import test
  - each stub is tagged TODO(0.24.0) for removal next release

nemoguardrails/integrations/langchain/actions/__init__.py
Pouyanpi: I'll take care of it

PR #1773, opened 2026-04-07
refactor(llm): framework-owned provider registry

Pouyanpi:
7. **stack-7: framework-owned provider registry** (this PR)

## Description

Add register_provider/get_provider_names to LLMFramework protocol. 

our public API delegates to the active framework instead of importing from LangChain internals directly. LangChainFramework implements the new methods, routing to its internal chat/llm provider registries. Backward compat aliases register_chat_provider/register_llm_provider still work.

## Summary by CodeRabbit

* **New Features**
  * Added provider registration and discovery APIs to the framework, enabling dynamic registration and enumeration of available providers.

* **Deprecations**
  * Marked LLM-specific provider functions as deprecated; they will be removed in version 0.23.0. Use the new generic provider APIs instead.

### DIFF
--- a/nemoguardrails/llm/providers/__init__.py
+++ b/nemoguardrails/llm/providers/__init__.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import warnings
 from typing import Any, List
 
 from nemoguardrails.llm.frameworks import get_default_framework, get_framework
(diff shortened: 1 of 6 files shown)


nemoguardrails/llm/providers/__init__.py
```diff
@@ -13,20 +13,68 @@
...
+def _active_framework():
+    return get_framework(get_default_framework())
+
+
+def register_provider(name: str, provider_cls: Any) -> None:
+    _active_framework().register_provider(name, provider_cls)
+
+
+def get_provider_names() -> List[str]:
+    return _active_framework().get_provider_names()
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai these methods are part of the LLMFramework protocol. Any implementation must provide them. the `hasattr` guards on the deprecated functions exist because those are optional langchain specific extensions, not protocol requirements.

nemoguardrails/llm/providers/__init__.py
```diff
@@ -13,20 +13,68 @@
...
+        return fw.get_llm_provider_names()  # type: ignore[attr-defined]
+    return fw.get_provider_names()
+
 
 __all__ = [
-    "get_chat_provider_names",
-    "get_community_chat_provider_names",
-    "get_llm_provider_names",
+    "register_provider",
+    "get_provider_names",
```
tgasser-nv, line 75: Do we still need to export `get_chat_provider_names` and `get_community_chat_provider_names` with a deprecation warning before we move over to `get_provider_names` which provides both?

tgasser-nv (approved):

nemoguardrails/llm/providers/__init__.py
Pouyanpi: Good point. get_chat_provider_names doesn't have a deprecation warning but should, since the chat/text distinction is going away in 0.23.0. right now it's kept for backwards compat with lc's framework which still tracks chat and text providers separately. Once text completion is removed in 0.23.0, get_chat_provider_names becomes redundant and can be deprecated together with get_llm_provider_names. get_community_chat_provider_names isn't exported from this module at all because it's internal to the langchain framework. Will take care of it 👍🏻

PR #1797, opened 2026-04-17
feat(llm): add default framework with OpenAI-compatible client

Pouyanpi:
## Description

Adds a native, HTTP client for OpenAI compatible endpoints (OpenAI, NVIDIA NIM, vLLM, local Ollama), a chat-model adapter implementing the `LLMModel` protocol, and `DefaultFramework` that wires this as the default LLM framework in place of LangChain.

End state:

- minimal dependency path from user input to any OpenAI-compatible provider
- Typed error handling
- deterministic streaming with explicit cleanup on consumer abort (PEP 533 war applied)
- fixture backed offline tests for CI determinism, plus opt-in live tests against real APIs
- `DefaultFramework` is now the default; LangChain stays reachable via `NEMOGUARDRAILS_LLM_FRAMEWORK=langchain`

## How to review

The 7 commits are self-contained and build on each other. Review in order, top-down
**Note:**
 +4317 additions belongs to a6a3fd5bfb7c21dae410514ecc10f30689eda1e1 were fixture files are created (tests/llm/clients/fixtures/*.json)

## Testing

Client subset:
```
poetry run pytest tests/llm/clients/
```

Opt-in live tests (real API calls):
```
LIVE_TEST_MODE=1 OPENAI_API_KEY=... NVIDIA_API_KEY=... poetry run pytest tests/llm/clients/test_openai_compatible_live.py
```

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI-compatible LLM client with support for multiple providers (OpenAI, Azure, NVIDIA NIM, local endpoints).
  * Introduced streaming chat completion responses with Server-Sent Event parsing.
  * Added new default LLM framework for managing provider registration and model instantiation.

* **Bug Fixes & Improvements**
  * Enhanced error handling with structured exception hierarchy for authentication, rate limits, context window, and network failures.
  * Implemented intelligent retry logic with exponential backoff and provider-specific metadata enrichment.

* **Configuration**
  * Updated default OpenAI model from `gpt-3.5-turbo-instruct` to `gpt-4`.


nemoguardrails/llm/frameworks.py
```diff
@@ -56,5 +61,10 @@ def get_default_framework() -> str:
 
 def _reset_frameworks() -> None:
     global _default_framework
+    for fw in list(_frameworks.values()):
+        try:
+            asyncio.run(fw.reset())
+        except RuntimeError:
+            pass
```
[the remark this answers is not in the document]
Pouyanpi: _reset_frameworks is a test-only hook (tests/conftest.py, tests/llm/test_frameworks.py). sync pytest fixtures run outside any loop thus asyncio.run() works. The edge case (fixture triggered while a loop is running) is unlikely and the cleanup is just closing httpx clients. 
  
  Also your solution looks like a nightmare

nemoguardrails/llm/clients/openai_chat_model.py
```diff
@@ -0,0 +1,295 @@
...
+            args_dict = raw_args
+
+        return ToolCall(
+            id=tc["id"],
+            type=tc.get("type", "function"),
+            function=ToolCallFunction(
+                name=func.get("name", ""),
+                arguments=args_dict,
+            ),
+        )
```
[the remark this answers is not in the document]
Pouyanpi: fair enough, will push the fix

nemoguardrails/llm/default_framework.py
```diff
@@ -0,0 +1,132 @@
...
+        return sorted({*_DEFAULT_BASE_URLS, *self._providers})
+
+    async def reset(self) -> None:
+        for client in list(self._clients.values()):
+            try:
+                await client.close()
+            except Exception:
+                pass
+        self._clients.clear()
+        self._providers.clear()
```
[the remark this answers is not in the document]
Pouyanpi: The behavior is deliberate. reset is kind of a nuke state hook. will add a docstring

nemoguardrails/llm/clients/_errors.py
```diff
@@ -0,0 +1,188 @@
...
+
+_UNSUPPORTED_PARAMS_KEYWORDS = [
+    "unsupported parameter",
+    "is not supported",
+    "not allowed",
+    "unknown parameter",
+    "unrecognized parameter",
+]
+
+_SECRET_PATTERN = re.compile(r"(sk-|nvapi-|key-|bearer\s+)\S+", re.IGNORECASE)
```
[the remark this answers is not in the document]
Pouyanpi: will push the fix

nemoguardrails/llm/clients/openai_chat_model.py
```diff
@@ -0,0 +1,295 @@
...
+    "tool_calls": "tool_calls",
+    "content_filter": "content_filter",
+}
+
+_STANDARD_RESPONSE_KEYS = frozenset({"model", "choices", "usage", "id", "object", "created", "_response_headers"})
+
+
+def _is_reasoning_model(model_name: str) -> bool:
+    name = model_name.lower()
+    return name.startswith("o1") or name.startswith("o3") or (name.startswith("gpt-5") and "chat" not in name)
```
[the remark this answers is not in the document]
Pouyanpi: tightened it a bit more, check the last commit

nemoguardrails/llm/clients/openai_compatible.py
```diff
@@ -0,0 +1,87 @@
...
+        payload: Dict[str, Any] = {
+            "model": model,
+            "messages": messages,
+        }
+        if stop:
+            payload["stop"] = stop
+        if stream:
+            payload["stream"] = True
+            payload["stream_options"] = {"include_usage": True}
+        payload.update(kwargs)
```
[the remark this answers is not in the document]
Pouyanpi: it's fixed now

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,250 @@
...
+            try:
+                delay = float(retry_after)
+                if 0 < delay <= MAX_RETRY_AFTER:
+                    return delay
+                log.debug("Ignoring Retry-After=%r (out of range, max=%s)", retry_after, MAX_RETRY_AFTER)
+            except ValueError:
+                log.debug("Ignoring unparseable Retry-After=%r", retry_after)
+
+        sleep = min(INITIAL_RETRY_DELAY * (2.0**retries_attempted), MAX_RETRY_DELAY)
+        jitter = 1 - 0.25 * random.random()
```
tgasser-nv: The jitter subtracted here will exacerbate the thundering-herd problem, each jitter value will be 1 or less. Recommend making random component symmetric about 0, or better to use [Full Jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/).

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,250 @@
...
+    MAX_RETRY_AFTER,
+    MAX_RETRY_DELAY,
+    RETRYABLE_STATUS_CODES,
+)
+
+log = logging.getLogger(__name__)
+
+
+class BaseClient:
+    _client: httpx.AsyncClient
```
tgasser-nv, line 48: Why httpx rather than asyncio? Are there any performance considerations between the two? I chose asyncio (somewhat randomly) for the IORails clients

nemoguardrails/llm/default_framework.py
```diff
@@ -0,0 +1,138 @@
...
+        connect_timeout = kwargs.pop("connect_timeout", None)
+        max_retries = kwargs.pop("max_retries", None)
+        default_headers = kwargs.pop("default_headers", None)
+        default_query = kwargs.pop("default_query", None)
+
+        client = self._get_or_create_client(
+            base_url, api_key, timeout, connect_timeout, max_retries, default_headers, default_query
+        )
+
+        return OpenAIChatModel(client=client, model=model_name, **kwargs)
```
tgasser-nv: At what level are connections pooled? Do we have a single pool of connections handling multiple inference requests ? Or do we create new connections on each inference request (likely inefficient)

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,250 @@
...
+                retries_attempted += 1
+                continue
+
+            if response.status_code >= 400:
+                raise_for_status(response.status_code, response.text, response.headers, ctx)
+            data = response.json()
+            data["_response_headers"] = dict(response.headers)
+            return data
+
+    async def _apost_stream(self, path: str, payload: Dict[str, Any]) -> AsyncGenerator[Dict[str, Any], None]:
```
tgasser-nv, line 198: Need to add a inter-chunk-timeout as well as the non-streaming timeout (in constants.py)

nemoguardrails/llm/frameworks.py
```diff
@@ -13,13 +13,14 @@
...
 
+import asyncio
 import os
 from typing import Dict
 
 from nemoguardrails.types import LLMFramework
 
 _frameworks: Dict[str, LLMFramework] = {}
-_default_framework: str = os.environ.get("NEMOGUARDRAILS_LLM_FRAMEWORK", "langchain")
+_default_framework: str = os.environ.get("NEMOGUARDRAILS_LLM_FRAMEWORK", "default")
```
tgasser-nv, line 26: Why not use a bool here, are you expecting multiple default frameworks in future rather than Langchain vs non-Langchain default?

nemoguardrails/llm/clients/_errors.py
```diff
@@ -0,0 +1,188 @@
...
+        if not error_message:
+            error_message = parsed_body.get("message") or parsed_body.get("detail") or ""
+    return error_message, error_type, error_code, param
+
+
+def _build_error_fields(parsed_body: Any, raw_body: str, headers: Any, ctx: ErrorContext) -> Tuple[str, Dict[str, Any]]:
+    error_message, error_type, error_code, param = _extract_from_parsed_body(parsed_body)
+    if not error_message:
+        error_message = raw_body or ""
+    error_message = _redact_secrets(error_message)
```
tgasser-nv, line 120: We need to redact body and response headers as well, in case they contain API keys, confidential info, or anything else sensitive

nemoguardrails/llm/clients/_errors.py
```diff
@@ -0,0 +1,188 @@
...
+    LLMBadRequestError,
+    LLMClientError,
+    LLMContextWindowError,
+    LLMRateLimitError,
+    LLMServerError,
+    LLMTimeoutError,
+    LLMUnsupportedParamsError,
+)
+
+_CONTEXT_WINDOW_KEYWORDS = [
```
tgasser-nv, line 34: Where did you get this list, how do you know whether it's complete? Same for `_UNSUPPORTED_PARAMS_KEYWORDS`, and `_SECRET_PATTERN`

nemoguardrails/exceptions.py
```diff
@@ -19,6 +19,16 @@
     "InvalidModelConfigurationError",
     "InvalidRailsConfigurationError",
     "LLMCallException",
+    "LLMClientError",
```
tgasser-nv, line 22: How can we ensure the same errors are thrown for both Langchain and Default framework for upper-level callers (i.e. FastAPI service) to have consistent error-handling and client error-returning code?

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,250 @@
...
+
+            if self._should_retry(response.status_code, response.headers) and retries_remaining > 0:
+                await self._sleep_for_retry(retries_attempted, response.headers)
+                retries_remaining -= 1
+                retries_attempted += 1
+                continue
+
+            if response.status_code >= 400:
+                raise_for_status(response.status_code, response.text, response.headers, ctx)
+            data = response.json()
```
tgasser-nv: Recommend wrapping this in a `try .. except` to catch a `JSONDecodeError and raise a corresponding `LLMResponseValidationError`. Something like:

```
  try:
      data = response.json()
  except json.JSONDecodeError as err:
      raise LLMResponseValidationError(
          f"Provider returned non-JSON response (status={response.status_code}, "
          f"content-type={response.headers.get('content-type')!r}): {err}",
          response_data=None,
          model_name=ctx.model_name,
          provider_name=ctx.provider_name,
          base_url=ctx.base_url,
      ) from err
```

nemoguardrails/llm/frameworks.py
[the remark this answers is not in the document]
tgasser-nv: @Pouyanpi This is a valid concern, even in tests. It raises the possibility of connection pools leaking between tests and accumulating over the course of a pytest run (giving flaky results)

nemoguardrails/llm/clients/openai_chat_model.py
```diff
@@ -0,0 +1,303 @@
...
+    "stop": "stop",
+    "length": "length",
+    "tool_calls": "tool_calls",
+    "content_filter": "content_filter",
+}
+
+_STANDARD_RESPONSE_KEYS = frozenset({"model", "choices", "usage", "id", "object", "created", "_response_headers"})
+
+
+def _is_reasoning_model(model_name: str) -> bool:
```
tgasser-nv: This is a brittle way to detect reasoning models, but I don't know a better way today. The gpt5 branch will miss `gpt-5.4` and `gpt-5.5` .

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,250 @@
...
+        return ErrorContext(
+            model_name=model_name,
+            provider_name=self.provider_name,
+            base_url=self.provider_url,
+        )
+
+    def _build_headers(self) -> Dict[str, str]:
+        headers = {"Content-Type": "application/json"}
+        if self._api_key:
+            headers["Authorization"] = f"Bearer {self._api_key}"
```
tgasser-nv, line 113: Should raise a `ValueError` here if the user provided a `custom_headers` `Authorization` field and we try to override it via `self.api_key`. Need to document that custom headers overrides the api key env value

nemoguardrails/llm/clients/openai_compatible.py
```diff
@@ -0,0 +1,88 @@
...
+    ) -> Dict[str, Any]:
+        payload: Dict[str, Any] = {
+            "model": model,
+            "messages": messages,
+        }
+        if stop:
+            payload["stop"] = stop
+        if stream:
+            payload["stream"] = True
+            if self.provider_name in {"openai", "nim", "azure"}:
```
tgasser-nv: Is this too restrictive? We have Ollama in our supported engines list, and how about if we use a locally running model but still want token counts (e.g. for ITL, tokens/sec measurements)

nemoguardrails/llm/clients/openai_compatible.py
```diff
@@ -0,0 +1,88 @@
...
+from typing import Any, AsyncGenerator, Dict, List, Optional
+
+from nemoguardrails.llm.clients.base import BaseClient
+
+
+class OpenAICompatibleClient(BaseClient):
+    _ROUTE = "/chat/completions"
+
+    @property
+    def provider_name(self) -> Optional[str]:
```
tgasser-nv: Why do a lookup on the base URL when we have the provider from `DefaultFramework.create_model()`?

nemoguardrails/llm/clients/_sse.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class SSEDecoder:
+    def __init__(self) -> None:
+        self._event = ""
+        self._data: List[str] = []
+        self._last_event_id = ""
+        self._retry: Optional[int] = None
+
+    def decode(self, line: str) -> Optional[ServerSentEvent]:
```
tgasser-nv, line 76: If a provider sends `\n\n\n` after getting a valid `id` (setting `self._last_event_id`) this would send an non-empty chunk with the most recent ID rather than None

nemoguardrails/llm/default_framework.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def get_provider_names(self) -> List[str]:
+        return sorted({*_DEFAULT_BASE_URLS, *self._providers})
+
+    async def reset(self) -> None:
+        """Close all pooled HTTP clients and clear registered providers.
+
+        Destructive: clears both the client pool and any providers registered
+        via ``register_provider``. Callers expecting custom providers to
+        survive must re-register them after ``reset``.
```
tgasser-nv, line 134: Why conflate both the custom providers metadata and connection pool dataplane in the same method? I'd expect a reset to apply to the connection pool rather than clearing custom providers

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,250 @@
...
+            if self._should_retry(response.status_code, response.headers) and retries_remaining > 0:
+                await self._sleep_for_retry(retries_attempted, response.headers)
+                retries_remaining -= 1
+                retries_attempted += 1
+                continue
+
+            if response.status_code >= 400:
+                raise_for_status(response.status_code, response.text, response.headers, ctx)
+            data = response.json()
+            data["_response_headers"] = dict(response.headers)
```
tgasser-nv, line 195: Rather than smuggling the headers our in the same dict as the HTTP body, recommend using a namedTuple / dataclass here with headers and body separated

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,250 @@
...
+    def _build_headers(self) -> Dict[str, str]:
+        headers = {"Content-Type": "application/json"}
+        if self._api_key:
+            headers["Authorization"] = f"Bearer {self._api_key}"
+        headers.update(self._custom_headers)
+        return headers
+
+    @staticmethod
+    def _should_retry(status_code: int, headers: Any) -> bool:
+        should = headers.get("x-should-retry")
```
tgasser-nv: Header comparisons need to be case-insensitive ([RFC 9110 section 5.1](https://www.rfc-editor.org/rfc/rfc9110.html)), need a `lower()` here

tgasser-nv (commented):
A high-level comment, this PR is way too large. It could easily have been stacked into 4 (or more) PRs. 

1. New client implementation.
2. Client fixture JSON and code to re-generate them later.
3. Moving tests (without edits).
4. New / modified tests.

Non-blocking feedback / comments:

* What is the customer-benefit of this change? Lightweight dependencies compared with Langchain? Better performance? More specific customizations.
* Can you measure performance difference between Langchain and Default LLM clients to make sure there are no regressions?

nemoguardrails/llm/clients/base.py
Pouyanpi: Nice catch, didn't pay attention to it. the current formula is lifted verbatim from the openai and anthropic:
https://github.com/openai/openai-python/blob/main/src/openai/_base_client.py#L754-L776
https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/_base_client.py#L779-L801

it just does 25% shave and not full Jitter.

nemoguardrails/llm/clients/base.py
Pouyanpi: I think you are referring to `aiohttp`?

reasons for httpx:

1. already a transitive dependency
2. SDK and ecosystem convention: openai, anthropic, `langchain-openai`, and the Google SDKs all use `httpx`. aligning with the majority keeps retry/timeout/error semantics consistent with what users coming from those clients already expect.
3. `MockTransport` for testing, the fixture-backed client tests intercept requests with `httpx.MockTransport` to assert on the actual wire payload. `aiohttp` has no comparable first-party hook.
4. public API stability: `LLMClient` accepts `http_client: httpx.AsyncClient` for injection. Switching transports would break that contract. this is mainly used so that users can supply a fully pre-configured client. it's required for real deployments for example corporate proxies, CA bundles etc.

nemoguardrails/llm/clients/base.py
Pouyanpi: Good point!
yes openai and anthropic do same. will fix in follow-up. we have to keep the default loose for reasoning models.

nemoguardrails/llm/clients/openai_chat_model.py
Pouyanpi: I'm going to update this based on:

https://github.com/NVIDIA-NeMo/Guardrails/pull/1811
https://github.com/NVIDIA-NeMo/Guardrails/pull/1814 (there is an important TODO).

I'll bring the logic here, then in a follow-up will share it as a helper.

nemoguardrails/llm/frameworks.py
Pouyanpi: it is a registry not toggle. register_framework() is public api. users plug in custom (litellm, in-house) and pick it as the default via `set_default_framework("custom_name")` or the `NEMOGUARDRAILS_LLM_FRAMEWORK` env var.  bool closes that door. symmetric with _frameworks: Dict[str, ...].

nemoguardrails/exceptions.py
Pouyanpi: typed `LLMClientError` is only from default framework. langchain raises sdk exceptions (openai.RateLimitError, etc.). both catchable as LLMCallException. Each framework has its own exceptions and we don't intend nor should we to consolidate all. 

Currently these are all wrapped in the opaque LLMCaLLException which we need to improve in follow-up PRs (not related to this PR though)

nemoguardrails/llm/clients/_errors.py
Pouyanpi: mostly from DataDesigner:

https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/packages/data-designer-engine/src/data_designer/engine/models/clients/errors.py#L259

nemoguardrails/llm/clients/_errors.py
Pouyanpi: response headers and bodies from openai/anthropic/nim don't contain api keys, those live in request headers, not response. error bodies are sanitized too (just {"error": {"message", "type", "code"}}). so this specific concern doesn't apply for the providers we ship clients for. 

if a provider leaks secrets in responses that's a provider bug, not our gate to hold. once we start redacting upstream content, we own that responsibility forever, across every provider and every error shape.

nemoguardrails/llm/clients/openai_chat_model.py
Pouyanpi: we can improve it using real-fixture tests. if there are cases that we are missing we should update these constants but there is no straight-forward way to catch all possible patterns right now.

nemoguardrails/llm/clients/base.py
Pouyanpi: this is a widely used pattern, current code already has custom_headers win over api_key in `headers.update(custom_headers) `runs after the api_key Authorization is set. so users passing Authorization in custom_headers get their value on the wire. matches openai sdk default_headers precedence (+ langchain-nvidia-ai-endpoints, langchain-openai, our own NMP). raising value error would break legitimate custom-auth flows (basic auth, raw jwt, pre-signed tokens). will add docstring note documenting the precedence 👍🏻

nemoguardrails/llm/default_framework.py
Pouyanpi: agreed. dataplane (close pools) and registry (clear providers) are different lifecycles. Will push the fix:+1

nemoguardrails/llm/clients/base.py
Pouyanpi: header name lookup is fine (httpx.Headers is CaseInsensitiveDict) but I agree on value side 👍🏻  I'll push the fix

nemoguardrails/llm/clients/base.py
Pouyanpi: agreed, this is a wider refactor (touches every `_apost` callsite + parsing layer), so I'd rather do it as a focused follow-up rather than wedge it into this PR. tracking it separately.

Is it ok with you?

nemoguardrails/llm/clients/openai_compatible.py
Pouyanpi: Yes, should've removed this fragile part already.

nemoguardrails/llm/clients/openai_compatible.py
Pouyanpi: yes, the allow list is too tight! I think nim doesn't allow it. I'll check, but i think it is better to enable it by default.

nemoguardrails/llm/clients/base.py
Pouyanpi: agreed 👍🏻

nemoguardrails/llm/default_framework.py
Pouyanpi: Two layers. 
one client per (base_url, api_key, ...) cached in DefaultFramework. each client owns one httpx.AsyncClient with pool: 1000 max, 100 keepalive (constants.py). 

Subsequent inference requests share (reuse) warm connections. no per-call handshake. two providers = two pools

tests/test_task_specific_model.py
```diff
@@ -89,4 +89,4 @@ def test_task_specific_model_for_generate_user_intent_and_generate_next_steps():
 
     other_calls = [call for call in res.log.llm_calls if call.task not in task_specific_tasks]
     for call in other_calls:
-        assert call.llm_model_name == "gpt-3.5-turbo-instruct"
+        assert call.llm_model_name == "gpt-4o"
```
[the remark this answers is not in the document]
Pouyanpi: done 👍🏻

tests/llm/clients/fixtures/openai_generate_tool_call.json
```diff
@@ -0,0 +1,82 @@
...
+    "openai-version": "2020-10-01",
+    "x-openai-proxy-wasm": "v0.1",
+    "x-ratelimit-limit-requests": "30000",
+    "x-ratelimit-limit-tokens": "150000000",
+    "x-ratelimit-remaining-requests": "29999",
+    "x-ratelimit-remaining-tokens": "149999987",
+    "x-ratelimit-reset-requests": "2ms",
+    "x-ratelimit-reset-tokens": "0s",
+    "x-request-id": "req_bac1cd4443a94d999e024070cff3715d",
+    "set-cookie": "__cf_bm=BpZ5pGWy1Pyh9XtTU1AueryapQwHwm6ckUgwsLt7GBk-1776675703-1.0.1.1-dOpm4_csUqQQ8PnYGox4WcDHih8hs15rk9nLYD.2sgSfDSuEVke28hK7RYHXanb1mG9Wiz6LjUXA_wwvjnPfJdooNV9ZU.jSGI9dH00y5dU; path=/; expires=Mon, 20-Apr-26 09:31:43 GMT; domain=.api.openai.com; HttpOnly; Secure; SameSite=None, _cfuvid=50XcnamXKFUmnlHGtBnPQq.X11hn2qWH8bHFJYHBzcc-1776675703363-0.0.1.1-604800000; path=/; domain=.api.openai.com; HttpOnly; Secure; SameSite=None",
```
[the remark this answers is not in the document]
Pouyanpi: done

tests/llm/clients/fixtures/openai_generate_refusal.json
```diff
@@ -0,0 +1,62 @@
...
+    "openai-version": "2020-10-01",
+    "x-openai-proxy-wasm": "v0.1",
+    "x-ratelimit-limit-requests": "30000",
+    "x-ratelimit-limit-tokens": "150000000",
+    "x-ratelimit-remaining-requests": "29999",
+    "x-ratelimit-remaining-tokens": "149999980",
+    "x-ratelimit-reset-requests": "2ms",
+    "x-ratelimit-reset-tokens": "0s",
+    "x-request-id": "req_b60ab91779fd4c4884b1aa7ecac9bda6",
+    "content-encoding": "gzip",
```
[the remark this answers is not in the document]
Pouyanpi: done

tests/llm/clients/fixtures/openai_error_401.json
```diff
@@ -0,0 +1,34 @@
...
+    "strict-transport-security": "max-age=31536000; includeSubDomains; preload",
+    "x-content-type-options": "nosniff",
+    "x-error-json": "ewogICJlcnJvciI6IHsKICAgICJtZXNzYWdlIjogIkluY29ycmVjdCBBUEkga2V5IHByb3ZpZGVkOiBzay1pbnZhbCoqKioqKioqKioqKioqdHVyZS4gWW91IGNhbiBmaW5kIHlvdXIgQVBJIGtleSBhdCBodHRwczovL3BsYXRmb3JtLm9wZW5haS5jb20vYWNjb3VudC9hcGkta2V5cy4iLAogICAgInR5cGUiOiAiaW52YWxpZF9yZXF1ZXN0X2Vycm9yIiwKICAgICJjb2RlIjogImludmFsaWRfYXBpX2tleSIsCiAgICAicGFyYW0iOiBudWxsCiAgfSwKICAic3RhdHVzIjogNDAxCn0=",
+    "x-openai-authorization-error": "401",
+    "x-openai-ide-error-code": "invalid_api_key",
+    "x-openai-internal-caller": "unknown_through_ide",
+    "x-openai-proxy-wasm": "v0.1",
+    "x-request-id": "req_29a348eec909427f9ef65e0969ffd451",
+    "alt-svc": "h3=\":443\"; ma=86400"
+  },
```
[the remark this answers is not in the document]
Pouyanpi: done, the auto record fixture also redact these now

nemoguardrails/llm/clients/openai_chat_model.py
```diff
@@ -0,0 +1,303 @@
...
+        raw_finish = choice.get("finish_reason")
+        finish_reason = _FINISH_REASON_MAP.get(raw_finish, "other") if raw_finish else None
+
+        return LLMResponseChunk(
+            delta_content=content,
+            delta_reasoning=reasoning,
+            model=data.get("model"),
+            finish_reason=finish_reason,
+            request_id=data.get("id"),
+            provider_metadata=provider_metadata,
```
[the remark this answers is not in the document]
Pouyanpi: Nice catch. Fixed and pushed 👍🏻

tests/llm/clients/test_client_config.py
```diff
@@ -0,0 +1,266 @@
...
+    @pytest.mark.asyncio
+    async def test_close_closes_owned_client(self):
+        client = OpenAICompatibleClient(base_url="https://api.openai.com/v1", api_key="sk")
+        owned = client._client
+        await client.close()
+        assert owned.is_closed
+
+    def test_invalid_type_raises(self):
+        with pytest.raises(TypeError, match="httpx.AsyncClient"):
+            OpenAICompatibleClient(base_url="https://api.openai.com/v1", http_client="not a client")
```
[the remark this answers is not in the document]
Pouyanpi: Yes, fix pushed

nemoguardrails/integrations/langchain/llm_adapter.py
```diff
@@ -262,6 +262,9 @@ def get_llm_provider_names(self) -> List[str]:
 
         return _get_llm()
 
+    async def reset(self) -> None:
+        return
```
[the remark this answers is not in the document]
Pouyanpi: we don't own it. Calling `client.close()` / `provider.reset()` on the wrapped LLM could break shared instances elsewhere in the user's process, and the lc llm lifecycle isn't part of our contract. The empty body is intentional: there are no framework-owned pools or pending tasks on this side (those live inside the user's LLM, managed by them or by langchain itself). the `DefaultFramework` path that does own pools has a real `reset()` that closes them.

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,250 @@
...
+    def _calculate_retry_delay(headers: Any, retries_attempted: int) -> float:
+        retry_after = headers.get("retry-after")
+        if retry_after:
+            try:
+                delay = float(retry_after)
+                if 0 < delay <= MAX_RETRY_AFTER:
+                    return delay
+                log.debug("Ignoring Retry-After=%r (out of range, max=%s)", retry_after, MAX_RETRY_AFTER)
+            except ValueError:
+                log.debug("Ignoring unparseable Retry-After=%r", retry_after)
```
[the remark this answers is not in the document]
Pouyanpi: good point, fixed it and pushed. However, anthropic and openai SDK doesn't do this

nemoguardrails/llm/default_framework.py
```diff
@@ -0,0 +1,138 @@
...
+    ) -> OpenAICompatibleClient:
+        key = (
+            base_url,
+            api_key or "",
+            timeout,
+            connect_timeout,
+            max_retries,
+            tuple(sorted((default_headers or {}).items())),
+            tuple(sorted((default_query or {}).items())),
+        )
```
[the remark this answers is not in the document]
Pouyanpi: Cache key lives inside code. Not shown to outside. API key is already in memory. Hash key does not help. If we want to hide API key, we must erase after use. We should not change cache key. Skip for now.

nemoguardrails/llm/clients/openai_chat_model.py
```diff
@@ -0,0 +1,303 @@
...
+                            if idx not in tool_call_acc:
+                                tool_call_acc[idx] = {
+                                    "id": tc_delta.get("id", ""),
+                                    "type": tc_delta.get("type", "function"),
+                                    "function_name": tc_delta.get("function", {}).get("name", ""),
+                                    "arguments_buffer": "",
+                                }
+                            arg_fragment = tc_delta.get("function", {}).get("arguments", "")
+                            if arg_fragment:
+                                tool_call_acc[idx]["arguments_buffer"] += arg_fragment
```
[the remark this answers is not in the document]
Pouyanpi: per OpenAI's streaming tool-call spec, `id`, `type`, and `function.name` are emitted in the first delta with a given `index`; subsequent deltas at the same `index` carry only `function.arguments` fragments. Keeping current behavior.

nemoguardrails/llm/default_framework.py
```diff
@@ -0,0 +1,138 @@
...
+
+        Destructive: clears both the client pool and any providers registered
+        via ``register_provider``. Callers expecting custom providers to
+        survive must re-register them after ``reset``.
+        """
+        for client in list(self._clients.values()):
+            try:
+                await client.close()
+            except Exception:
+                pass
```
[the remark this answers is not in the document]
Pouyanpi: fixed it 👍🏻

nemoguardrails/llm/frameworks.py
```diff
@@ -56,5 +61,10 @@ def get_default_framework() -> str:
...
 def _reset_frameworks() -> None:
     global _default_framework
+    for fw in list(_frameworks.values()):
+        try:
+            asyncio.run(fw.reset())
+        except RuntimeError:
+            pass
     _frameworks.clear()
-    _default_framework = os.environ.get("NEMOGUARDRAILS_LLM_FRAMEWORK", "langchain")
+    _default_framework = os.environ.get("NEMOGUARDRAILS_LLM_FRAMEWORK", "default")
```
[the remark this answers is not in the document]
Pouyanpi: Already pushed the fix

tests/llm/clients/_helpers.py
```diff
@@ -0,0 +1,280 @@
...
+
+OPENAI_BASE_URL = "https://api.openai.com/v1"
+OPENAI_DEFAULT_MODEL = "gpt-4o-mini"
+NIM_BASE_URL = "https://integrate.api.nvidia.com/v1"
+NIM_DEFAULT_MODEL = "nvidia/nemotron-3-nano-30b-a3b"
+
+
+def make_client(**kwargs):
+    return OpenAICompatibleClient(base_url="https://api.openai.com/v1", api_key="sk-test", **kwargs)
+
```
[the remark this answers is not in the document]
Pouyanpi: real concern but bounded in practice tests that replace client._client with mocks never call any real http method on the original AsyncClient so its connection pool is empty httpx is lazy connections only open on first request the leak is an idle python object referencing an empty pool until gc not open file descriptors  
what I did fix every test in test_client_config.py and TestProviderName in test_openai_compatible.py that constructed clients without closing them is now @pytest.mark.asyncio with async with or try finally await client.close so the AsyncClient is properly torn down that covers the visible leakers  
what I left for follow-up refactoring make_client itself to inject a MockTransport-based AsyncClient by default so tests dont need to override _client at all thats a wider refactor of the mock-heavy tests doing it in a focused PR

tests/llm/clients/test_openai_compatible.py
```diff
@@ -0,0 +1,1079 @@
...
+        c = OpenAICompatibleClient(base_url="https://integrate.api.nvidia.com/v1")
+        assert c.provider_name == "nim"
+
+    def test_local(self):
+        c = OpenAICompatibleClient(base_url="http://localhost:11434/v1")
+        assert c.provider_name == "local"
+
+    def test_azure(self):
+        c = OpenAICompatibleClient(base_url="https://mycompany.openai.azure.com/v1")
+        assert c.provider_name == "azure"
```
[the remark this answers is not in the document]
Pouyanpi: Converted `test_openai`, `test_nim`, `test_local`, `test_azure` to `@pytest.mark.asyncio` with `async with OpenAICompatibleClient(...) as c:` so the internal `httpx.AsyncClient` is properly closed.

nemoguardrails/llm/frameworks.py
```diff
@@ -56,5 +62,44 @@ def get_default_framework() -> str:
...
 def _reset_frameworks() -> None:
     global _default_framework
+
+    frameworks_to_close = list(_frameworks.values())
+    if frameworks_to_close:
+        _run_async_from_sync(_close_frameworks(frameworks_to_close))
+
     _frameworks.clear()
-    _default_framework = os.environ.get("NEMOGUARDRAILS_LLM_FRAMEWORK", "langchain")
+    _default_framework = os.environ.get("NEMOGUARDRAILS_LLM_FRAMEWORK", "default")
```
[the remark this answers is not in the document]
Pouyanpi: kept the sync `_reset_frameworks` API but split out the awaitable work into `_close_frameworks(frameworks)` and added `_run_async_from_sync(coro)` which detects a running loop with `asyncio.get_running_loop()`. added tests too

nemoguardrails/llm/frameworks.py
```diff
@@ -56,5 +62,44 @@ def get_default_framework() -> str:
...
+
+
+async def _close_frameworks(frameworks) -> None:
+    for fw in frameworks:
+        reset = getattr(fw, "reset", None)
+        if reset is None:
+            continue
+        await reset()
```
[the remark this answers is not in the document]
Pouyanpi: you're right, fix pushed 👍🏻

nemoguardrails/llm/frameworks.py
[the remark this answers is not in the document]
Pouyanpi: Sure, made the change 👍🏻

nemoguardrails/llm/frameworks.py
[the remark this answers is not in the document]
Pouyanpi: fixed it

Pouyanpi:
Thanks @tgasser-nv for the review :+1 I know this was a painful one given the size.

> A high-level comment, this PR is way too large. It could easily have been stacked into 4 (or more) PRs.
> 

Agreed. I went with self-contained commits hoping that would make the review easier/more managble and also my life easier. Excluding the ~4700 fixture related json files, considering AI code reviewers and your experience I totally agree that splitting it would have been the right call :+1 .  

> Non-blocking feedback / comments:
> 
> * What is the customer-benefit of this change? Lightweight dependencies compared with Langchain? Better performance? More specific customizations.

we're dropping langchain as a hard dependency on the default path. 

* lighter install footprint, no LangChain version-compat surface, smaller supply-chain attack surface. langchain-core + langchain-openai pull a deep transitive dependency tree (pydantic-v1 shims,
deprecated submodules, etc.). Each one is something we have to track for CVEs and version churn. Dropping them on the default path shrinks both the install size and the attack surface.

* beyond the size win: this PR is the structural prerequisite for making LangChain optional at all. Without a default framework that doesn't depend on LangChain, "make LangChain optional" is impossible since every code path needs some framework, and today every path goes through LangChain. The default framework gives us the alternative path so removing langchain-core from pyproject.toml becomes possible as in https://github.com/NVIDIA-NeMo/Guardrails/pull/1806. Users who do want LangChain still get it through LangChainFramework; nothing forces a migration.


> * Can you measure performance difference between Langchain and Default LLM clients to make sure there are no regressions?

I've done it already, no regression at either layer and we'll open the PR once this is merged :+1

nemoguardrails/llm/frameworks.py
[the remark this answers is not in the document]
Pouyanpi: fixed

Pouyanpi:
@tgasser-nv three of your items are already implemented as follow-up branches, will open as PRs after this lands:

- inter-chunk timeout: feat/llm/stream-read-timeout
- split reset(): refactor/llm/framework-aclose
- HTTPResponse dataclass: refactor/llm/http-response-dataclass

nemoguardrails/llm/default_framework.py
tgasser-nv: Is it in a different PR? I don't see any changes below

nemoguardrails/llm/clients/base.py
tgasser-nv: Sounds good, is there an Issue or draft PR with the fix?

nemoguardrails/llm/clients/openai_chat_model.py
```diff
@@ -0,0 +1,333 @@
...
+        result = []
+        for idx in sorted(acc.keys()):
+            entry = acc[idx]
+            raw_args = entry["arguments_buffer"]
+            # Graceful degrade: truncated (max_tokens) or malformed args => empty dict.
+            # The tool will surface the real error when invoked with no arguments.
+            try:
+                args_dict = json.loads(raw_args) if raw_args else {}
+            except json.JSONDecodeError:
+                args_dict = {}
```
tgasser-nv, line 187: Should we raise a `LLMResponseValidationError` here, same as in `_apost_stream()`?

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,279 @@
...
+        """Initialize the HTTP-backed LLM client.
+
+        custom_headers takes precedence over api_key-derived headers.
+        Passing custom_headers={"Authorization": "..."} overrides the
+        Bearer token built from api_key, allowing custom auth schemes
+        (Basic, raw JWT, pre-signed tokens, multi-scheme) to be used
+        alongside the env-var-driven api_key default.
+        """
+        self._base_url = base_url.rstrip("/")
+        self._api_key = api_key
```
tgasser-nv, line 70: nit: Could you add a warning if an API Key is provided and the base_url is `http://` rather than `https://`? Something saying the API Key will be sent in plaintext, not recommended for production use

tgasser-nv (approved):
Thanks for making the updates, just a few cleanups needed before merging

nemoguardrails/llm/clients/openai_chat_model.py
Pouyanpi: Different layers. sse.json() failing in _apost_stream is a server protocol violation so here the provider sent bytes that aren't valid JSON, so we have to raise. _finalize_tool_calls parses the model's own generated arguments string, which fails for ordinary reasons (max_tokens truncation, partial JSON). Graceful degrade to {} keeps the stream yielding so the consumer sees finish_reason=length with empty args and can handle it. The inline comment marks this as deliberate.

nemoguardrails/llm/clients/base.py
Pouyanpi: Done

nemoguardrails/llm/clients/base.py
Pouyanpi: Yes will share in the last comment

nemoguardrails/llm/clients/_sse.py
Pouyanpi: It's fixed

nemoguardrails/llm/default_framework.py
Pouyanpi: Yes, will share it in the last comment 👍🏻

nemoguardrails/llm/default_framework.py
```diff
@@ -0,0 +1,145 @@
...
+        default_query: Optional[Dict[str, Any]],
+    ) -> OpenAICompatibleClient:
+        key = (
+            base_url,
+            api_key or "",
+            timeout,
+            connect_timeout,
+            max_retries,
+            tuple(sorted((default_headers or {}).items())),
+            tuple(sorted((default_query or {}).items())),
```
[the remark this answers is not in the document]
Pouyanpi: why didn't you mention it in previous iterations. This is very annoying

nemoguardrails/llm/clients/base.py
```diff
@@ -0,0 +1,290 @@
...
+
+        if api_key and base_url.startswith("http://"):
+            host = (urlparse(base_url).hostname or "").lower()
+            if host not in ("localhost", "127.0.0.1", "::1") and not host.endswith(".local"):
+                warnings.warn(
+                    f"API key will be sent over plaintext HTTP to {base_url}; use https:// for production deployments.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+
```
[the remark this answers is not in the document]
Pouyanpi: our test show that this comment is not valid.

urlparse("http://[::1]:...").hostname returns "::1" without brackets

nemoguardrails/llm/default_framework.py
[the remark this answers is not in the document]
Pouyanpi: pushed the fix

Pouyanpi:
https://github.com/NVIDIA-NeMo/Guardrails/pull/1828
resolves https://github.com/NVIDIA-NeMo/Guardrails/pull/1797#discussion_r3140331094

https://github.com/NVIDIA-NeMo/Guardrails/pull/1829
resolves https://github.com/NVIDIA-NeMo/Guardrails/pull/1797#discussion_r3150960821

https://github.com/NVIDIA-NeMo/Guardrails/pull/1830
resolves https://github.com/NVIDIA-NeMo/Guardrails/pull/1797#discussion_r3140548145
cc. @tgasser-nv

PR #1808, opened 2026-04-22
docs(iorails): tech-writer review fixes for OTEL Logging page

miyoungc:
## Summary

Follow-up edits for #1807 based on a tech-writer review (accuracy audit + NVIDIA style + polish). Target is the PR branch so @tgasser-nv can merge these into #1807 before it lands on `develop`.

All changes are confined to `docs/observability/tracing/opentelemetry-logs.md`.

## Changes

**Accuracy**

- **Logger namespace scope.** The original wording said every submodule logs under `nemoguardrails`. In practice, `nemoguardrails/logging/verbose.py` uses the root logger (verified at `nemoguardrails/logging/verbose.py:195,208`), so a handler attached to `"nemoguardrails"` will not capture verbose-mode output. The bullet now calls this out and tells readers to attach the handler to the root logger as well if they need verbose records forwarded.
- **`LoggingHandler` caching.** The previous callout said `LoggingHandler()` "resolves the active `LoggerProvider` at emit time" — correct in spirit, but the SDK actually caches the logger on first emit. Calling `set_logger_provider(...)` afterward does not recover a handler that has already pinned a no-op logger. The callout now leads with "configure the provider first, then attach the handler" and explains the caching footgun.
- **`configure_logging()` propagation flag.** `configure_logging` only sets `propagate=False` on the first call, when the logger has no handlers yet (`nemoguardrails/guardrails/__init__.py:42-55`). Added a parenthetical so readers know re-running it against a pre-configured logger does not flip the flag.

**Structure**

- Moved the prerequisite `{important}` callout **above** the three-line code block in _Minimal Setup_ — it is a precondition, not a footnote, and readers will otherwise copy the snippet and skip the warning.

**NVIDIA style**

- `see` → `refer to` (two body-text instances; the in-code comment is unchanged).
- Bolded lead-ins in _Considerations_ end with a colon rather than a period.
- `via` → `through` in the reworded callout.

## Deliberately not changed

These came up in the review but are opinionated structural calls best left to the original author:

- Converting _Minimal / Full_ into a tab-set.
- Wrapping the full example in a dropdown.
- Converting _What the Exported Records Contain_ to a table.

Happy to open a follow-up if any of the above sound good.

## Test plan

- [ ] `make docs` / local MyST build succeeds with the updated callout placement and admonition syntax.
- [ ] Links (`../logging/index.md`, intra-page anchor to `#full-example-traces-and-logs-together`, external `OpenTelemetry Registry`) resolve in the built site.
- [ ] Rendered Considerations list shows bold labels with colons, not periods.
- [ ] Callout in Minimal Setup renders above the code block.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/observability/tracing/opentelemetry-logs.md
+++ b/docs/observability/tracing/opentelemetry-logs.md
@@ -27,7 +27,7 @@ This page covers the setup. For plain Python logging (verbose mode, explain, gen
 
 ## API and SDK Responsibilities
 
-The NeMo Guardrails library follows the OpenTelemetry library-instrumentation pattern:
+The NeMo Guardrails library follows the OpenTelemetry library-instrumentation pattern.
 
 - **The library depends on the OpenTelemetry API only.** It creates spans, emits log records, and otherwise participates in whatever OTEL pipeline the host application provides.
 - **The host application owns the SDK.** Configuring a `TracerProvider`, a `LoggerProvider`, exporters, and attaching handlers to Python's `logging` tree are all the application's responsibility.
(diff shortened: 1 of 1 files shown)


tgasser-nv (approved):
Looks good! Thanks

PR #1810, opened 2026-04-22
fix(actions): guard bot message extraction against composite specs

Pouyanpi:
## Description

`_extract_bot_message_example` unconditionally accessed `spec["name"]`, which raises `KeyError` for SpecOr / SpecAnd composite dicts produced by Colang v2.x or / and operators on bot flows. This crashed `LLMRails.__init__` for any config using that standard v2.x syntax, including the official interaction_loop and multi_modal tutorials.

Restore the pre-#1361 safety: return early for composite `_type` and revert the identifier from the copy-pasted `UtteranceUserActionFinished` back to `UtteranceBotAction`.

Regression introduced in 67de94723 (#1361).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bot message extraction to correctly handle composite specifications using `or` and `and` operators in bot flows.

* **Tests**
  * Added test coverage for composite bot flow specifications to ensure proper validation and handling.

### DIFF
--- a/nemoguardrails/actions/llm/generation.py
+++ b/nemoguardrails/actions/llm/generation.py
@@ -177,7 +177,10 @@ class LLMGenerationActions:
             else cast(Dict, spec_op.spec)
         )
 
-        if not spec["name"] or spec["name"] != "UtteranceUserActionFinished" or "script" not in spec["arguments"]:
+        if spec.get("_type") in ("spec_or", "spec_and"):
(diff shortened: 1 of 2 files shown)


tgasser-nv (approved):
Thanks for fixing. I pulled the `generation.py` over from the `develop` branch and both the new tests failed so they isolate the regression.

PR #1824, opened 2026-04-26
feat(library): add peyeeye PII redaction & rehydration guardrail

tim-peyeeye:
## Summary

Adds an NeMo Guardrails community integration for the [peyeeye.ai](https://peyeeye.ai) PII redaction & rehydration API. The integration is modeled after the existing `privateai` library but adds a *rehydration* output rail so that the LLM only ever sees stable placeholders (e.g. `[EMAIL_1]`), while the user still sees the original values in the bot's response.

### Pre/post hooks

- **Input rail (`peyeeye redact input`)** — calls `POST /v1/redact`, swaps each detected entity for a placeholder, and stashes the returned session id (`ses_…` for stateful, `skey_…` for stateless) on the global context so the output rail can pick it up.
- **Output rail (`peyeeye rehydrate output`)** — calls `POST /v1/rehydrate` with the stashed session id and swaps the placeholders back to the originals before the response leaves the system. For stateful sessions, fires a best-effort `DELETE /v1/sessions/{id}` for cleanup.
- **Retrieval rail (`peyeeye redact retrieval`)** — same shape as the input rail, but for retrieved knowledge-base chunks.

The integration refuses to silently forward unredacted text: malformed responses, length mismatches between the request batch and the redacted response, and 4xx/5xx all raise typed exceptions (`PEyeEyeGuardrailMissingSecrets` for auth, `PEyeEyeGuardrailAPIError` for everything else).

### Config example

```yaml
rails:
  config:
    peyeeye:
      api_base: https://api.peyeeye.ai
      input:
        entities: [EMAIL, PHONE, CARD]
        locale: auto
        session_mode: stateful   # or "stateless" for AEAD-sealed blobs
  input:
    flows:
      - peyeeye redact input
  output:
    flows:
      - peyeeye rehydrate output
```

`PEYEEYE_API_KEY` is read from the environment.

### Tests added

`tests/test_peyeeye.py` (17 tests, all mock-only via `aioresponses`):

- redact happy path (stateful + stateless)
- length mismatch raises (load-bearing — no silent partial redaction)
- unexpected response shape raises
- missing API key raises
- 401 → `PEyeEyeGuardrailMissingSecrets`; 5xx → `PEyeEyeGuardrailAPIError`
- empty text short-circuits (no API call)
- entities/locale/session_mode are forwarded correctly in the request body
- session id is stashed onto the action `context` dict
- rehydrate replaces placeholders and triggers session DELETE
- rehydrate is a no-op on empty inputs
- rehydrate swallows API errors and returns the original text
- rehydrate skips DELETE for `skey_` (stateless) sessions
- end-to-end rail test using `RailsConfig.from_content` + `TestChat`

```
$ poetry run pytest tests/test_peyeeye.py -q
.................                                                        [100%]
17 passed in 1.44s
```

### CHANGELOG

Added an entry under `## [Unreleased]` -> `### 🚀 Features`.

### Docs

- New community page: `docs/configure-rails/guardrail-catalog/community/peyeeye.md`
- New section appended to `docs/configure-rails/guardrail-catalog/pii-detection.md`

## Test plan

- [x] `poetry run pytest tests/test_peyeeye.py -v` — 17/17 pass locally
- [x] `poetry run pytest -k peyeeye` — 17/17 pass
- [x] `poetry run pytest tests/test_config_validation.py tests/test_config_loading.py tests/test_privateai.py tests/test_activefence_rail.py` — adjacent suites still green
- [x] `poetry run ruff check` and `poetry run ruff format --check` clean on all touched files
- [x] DCO sign-off present on the commit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 >
 > The changes related to the Colang language and runtime have moved to [CHANGELOG-Colang](./CHANGELOG-Colang.md) file.
 
+## [Unreleased]
+
+### 🚀 Features
+
+- *(library)* Add Peyeeye PII redaction & rehydration guardrail. New input rail
+  redacts PII from the user message before it reaches the LLM and a paired
+  output rail rehydrates the model's response with the original values via
+  the [peyeeye.ai](https://peyeeye.ai) API. Supports both stateful (`ses_…`)
+  and stateless (`skey_…` AEAD-sealed) session modes.
+
 ## [0.21.0] - 2026-03-12
 
 ### 🚀 Features
--- a/docs/configure-rails/guardrail-catalog/community/peyeeye.md
+++ b/docs/configure-rails/guardrail-catalog/community/peyeeye.md
@@ -45,8 +45,11 @@ rails:
         locale: auto
         session_mode: stateful
       output:
-        # Output is rehydrated via the input session, so options here only
-        # apply if you also enable `peyeeye redact retrieval` etc.
+        # Output rehydration uses the session ids stashed by the input
+        # and retrieval rails, so the options here (locale / entities /
+        # session_mode) are only consulted if you call `peyeeye_redact`
+        # with `source="output"` directly from a custom flow. The
+        # built-in `peyeeye rehydrate output` flow does not read them.
         locale: auto
   input:
     flows:
--- a/docs/configure-rails/guardrail-catalog/community/peyeeye.md
+++ b/docs/configure-rails/guardrail-catalog/community/peyeeye.md
@@ -1,0 +1,106 @@
+# Peyeeye Integration
+
+[Peyeeye](https://peyeeye.ai) is a PII redaction & rehydration API. The
+integration lets NeMo Guardrails redact PII from a user's message before it is
+sent to the LLM, and then rehydrate the model's response so the user sees the
+original values — all without the LLM ever seeing the raw PII.
+
+## How it works
+
+1. **Input rail**: the user message is sent to `POST /v1/redact`, which returns
+   the message with each detected entity swapped for a stable placeholder
+   (e.g. `[EMAIL_1]`) plus a session id.
+2. **LLM call**: the redacted message goes to the LLM. The LLM's response
+   typically echoes some placeholders (`"I have emailed you at [EMAIL_1]"`).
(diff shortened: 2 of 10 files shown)


tim-peyeeye:
Thanks for the careful pass — pushed [`6cd6f4a`](https://github.com/peyeeye/Guardrails/commit/6cd6f4a36) addressing every actionable item.

**Greptile P1 — `request.py:127` null fallback**
Switched to `payload.get("text") or text`, so a `{"text": null}` response no longer propagates `None` into `$bot_message`. Covered by the new `test_rehydrate_falls_back_when_payload_text_is_null`.

**CodeRabbit P1 — `request.py:176-178` error body verbatim in exception**
Added `_safe_snippet()` (200-char cap with a truncation marker); both the 4xx/5xx and `ContentTypeError` paths now embed only a truncated body. Covered by `test_rehydrate_error_body_is_truncated`.

**CodeRabbit P1 — `request.py:179-181` non-dict payload guard**
`_post_json` now rejects non-object JSON with a `PEyeEyeGuardrailAPIError` instead of letting `payload.get(...)` blow up later with `AttributeError`. Covered by `test_redact_rejects_non_object_json_payload`.

**CodeRabbit nit — `actions.py:188` redundant `isinstance`**
Dropped; the type annotation plus the falsy guard above already cover it.

**Greptile P2 — `actions.py` blocking cleanup**
`peyeeye_delete_session` is now fired with `asyncio.create_task` so the DELETE doesn't add up to 10s to every rehydrated response. Falls back to inline `await` if no loop is running (defensive — shouldn't happen from an async action). Covered by `test_rehydrate_cleanup_runs_in_background`.

**Both bots — `flows.co` / `flows.v1.co` retrieval session unused**
`peyeeye rehydrate output` now rehydrates against `$peyeeye_retrieval_session_id` first, then `$peyeeye_input_session_id`, so retrieval-side redactions are restored alongside input-side ones.

**CodeRabbit nit — `tests/test_peyeeye.py:271-277` dead `_capture`**
Removed.

**CodeRabbit doc nit — `community/peyeeye.md` output options note**
Clarified that the built-in `peyeeye rehydrate output` flow does **not** read the `output.*` options (it threads stashed session ids); those only apply when a custom flow calls `peyeeye_redact` with `source="output"` directly.

Local: 21/21 tests pass, `ruff check` and `ruff format --check` clean. Commit is DCO-signed (`Signed-off-by: Tim <info@airskip.app>`).

PR #1826, opened 2026-04-26
Add speculative generation for non-streaming generate_async

hazai:
## Description

Race input rails against LLM generation using asyncio.wait(FIRST_COMPLETED), to overlap safety checks with model inference, for reducing latency.

  - Add speculative_generation config flag to InputRails
  - Split _do_generate into _do_generate_sequential and _do_generate_speculative
  - Add _race_rails_and_generation with proper task cancellation and cleanup
  - Add OTel span attributes for speculative mode telemetry
  - Warn when speculative_generation is enabled but streaming is used
  - Add 7 tests covering all race outcomes and error propagation

## Benchmarking results (non-streaming only)

The benchmarking results below were generated using the tooling in the `benchmarking` directory. The reference latency is set to 5,000ms for each request (0.5s content-safety input, 4s generation, 0.5s content-safety output). There are three runs: LLMRails (as a baseline), IORails without speculative generation, and IORails with speculative generation. 

Because speculative generation runs input content-safety and generation in-parallel, the end-to-end latency theoretically reduces by 500ms. However, the IORails engine adds some latency to buffer and process requests (e.g. at concurrency 32 p50 of 13.3ms). So the overall saving should be 500ms - 13.3ms = 486.7ms which is within 2.3ms of the measured 490ms value.

### LLMRails Baseline

image

### IORails without speculative generation

image

### IORails **with** speculative generation

image

## Related Issue(s)

[NGUARD-746](https://jirasw.nvidia.com/browse/NGUARD-746)

## Release Notes

* **New Features**
  * Introduced speculative generation capability: input safety checks and LLM generation now execute concurrently when enabled, reducing latency and improving throughput.
  * Enhanced tracing with metrics capturing execution order, timing savings, and rejection sources.
  * Added configuration option to control speculative generation behavior (defaults to disabled).

* **Tests**
  * Added comprehensive test suite validating concurrent execution, error handling, and sequential fallback behavior.

### DIFF
--- a/nemoguardrails/guardrails/iorails.py
+++ b/nemoguardrails/guardrails/iorails.py
@@ -24,6 +24,7 @@ import asyncio
 import json
 import logging
 import time
+import warnings
 from collections.abc import AsyncGenerator, AsyncIterator
 from contextlib import nullcontext, suppress
 from typing import TYPE_CHECKING, Optional, Union
@@ -51,6 +52,7 @@ from nemoguardrails.guardrails.telemetry import (
     record_stream_rejected,
     register_nonstream_saturation_gauges,
     request_metrics,
+    set_speculative_span_attrs,
     stream_active_metric,
     traced_request,
 )
@@ -114,7 +116,6 @@ class IORails:
             tracer=self._tracer,
         )
         self._speculative_generation = config.rails.input.speculative_generation or False
-        self._warned_speculative_streaming = False
 
         # Non-streaming admission queue + worker pool (owned by IORails so
         # all request-path concurrency controls sit under one roof).  The
@@ -348,19 +349,35 @@ class IORails:
         gen_task = asyncio.create_task(self.engine_registry.model_call("main", messages, **llm_kwargs))
 
         try:
-            response = await self._race_rails_and_generation(rails_task, gen_task, req_id, t0, request_span)
-        except BaseException:
+            response = await self._parallel_input_rail_and_response_generation(
+                rails_task, gen_task, req_id, t0, request_span
+            )
+        except BaseException as outer_exc:
             for t in (rails_task, gen_task):
                 if not t.done():
                     t.cancel()
-            # Drain all tasks (including done) to retrieve exceptions and avoid
-            # asyncio "Task exception was never retrieved" warnings.
-            await asyncio.gather(rails_task, gen_task, return_exceptions=True)
+            # Drain all tasks (including done) to retrieve their exceptions and
+            # avoid asyncio "Task exception was never retrieved" warnings, then
+            # log any genuine errors that get swallowed here (i.e. not the
+            # exception being re-raised and not cancellations from above).
+            rails_exc, gen_exc = await asyncio.gather(rails_task, gen_task, return_exceptions=True)
+            for name, exc in (("input_rails", rails_exc), ("generation", gen_exc)):
+                if (
+                    isinstance(exc, BaseException)
+                    and not isinstance(exc, asyncio.CancelledError)
+                    and exc is not outer_exc
+                ):
+                    log.warning(
+                        "[%s] %s task error discarded during cleanup: %r",
+                        req_id,
+                        name,
+                        exc,
+                    )
             raise
 
         return response
 
-    async def _race_rails_and_generation(
+    async def _parallel_input_rail_and_response_generation(
         self,
         rails_task: asyncio.Task,
         gen_task: asyncio.Task,
@@ -369,9 +386,13 @@ class IORails:
         request_span: Optional["Span"] = None,
     ) -> Optional[LLMResponse]:
         """Race input rails against LLM generation, return LLMResponse or None (rejected)."""
-        done, pending = await asyncio.wait({rails_task, gen_task}, return_when=asyncio.FIRST_COMPLETED)
+        done, _ = await asyncio.wait({rails_task, gen_task}, return_when=asyncio.FIRST_COMPLETED)
 
-        first_completed = "input_rails" if rails_task in done else "generation"
+        first_completed = (
+            GuardrailsAttributes.SPECULATIVE_FIRST_COMPLETED_INPUT_RAILS
+            if rails_task in done
+            else GuardrailsAttributes.SPECULATIVE_FIRST_COMPLETED_GENERATION
+        )
 
         if rails_task in done:
             input_result = rails_task.result()
@@ -384,16 +405,18 @@ class IORails:
                 # tasks finish simultaneously, gen_task may hold a stored exception that
                 # would leak through suppress(CancelledError). gather drains it safely.
                 gen_result = (await asyncio.gather(gen_task, return_exceptions=True))[0]
-                if isinstance(gen_result, Exception) and not isinstance(gen_result, asyncio.CancelledError):
+                if isinstance(gen_result, BaseException) and not isinstance(gen_result, asyncio.CancelledError):
                     log.warning("[%s] LLM generation error suppressed: %s", req_id, gen_result)
                 if self._metrics_enabled:
                     record_request_blocked(RailDirection.INPUT)
-                self._set_speculative_span_attrs(request_span, first_completed, "input_rails", 0.0)
+                set_speculative_span_attrs(
+                    request_span, first_completed, GuardrailsAttributes.SPECULATIVE_FIRST_COMPLETED_INPUT_RAILS, 0.0
+                )
                 return None
 
             # Rails passed — wait for generation to finish
             response = await gen_task
-            self._set_speculative_span_attrs(request_span, first_completed, "none", rails_elapsed * 1000)
+            set_speculative_span_attrs(request_span, first_completed, "none", rails_elapsed * 1000)
         else:
             # Generation finished first — wait for rails verdict
             response = gen_task.result()
@@ -405,24 +428,15 @@ class IORails:
                 log.info("[%s] Input blocked (speculative, gen-first): %s", req_id, input_result.reason)
                 if self._metrics_enabled:
                     record_request_blocked(RailDirection.INPUT)
-                self._set_speculative_span_attrs(request_span, first_completed, "input_rails", 0.0)
+                set_speculative_span_attrs(
+                    request_span, first_completed, GuardrailsAttributes.SPECULATIVE_FIRST_COMPLETED_INPUT_RAILS, 0.0
+                )
                 return None
 
-            self._set_speculative_span_attrs(request_span, first_completed, "none", gen_elapsed * 1000)
+            set_speculative_span_attrs(request_span, first_completed, "none", gen_elapsed * 1000)
 
         log.debug("[%s] Main LLM response: %s", req_id, truncate(response.content))
         return response
-
-    @staticmethod
-    def _set_speculative_span_attrs(
-        span: Optional["Span"], first_completed: str, first_rejector: str, time_saved_ms: float
-    ) -> None:
-        if span is None:
-            return
-        span.set_attribute(GuardrailsAttributes.SPECULATIVE_MODE_ACTIVE, True)
-        span.set_attribute(GuardrailsAttributes.SPECULATIVE_FIRST_COMPLETED, first_completed)
-        span.set_attribute(GuardrailsAttributes.SPECULATIVE_FIRST_REJECTOR, first_rejector)
-        span.set_attribute(GuardrailsAttributes.SPECULATIVE_TIME_SAVED_MS, time_saved_ms)
 
     def _validate_streaming_with_output_rails(self) -> None:
         """Raise if output rails exist but streaming is not enabled for them."""
@@ -466,9 +480,8 @@ class IORails:
             asyncio.QueueFull: If the streaming concurrency limit is
                 reached (load shedding).
         """
-        if self._speculative_generation and not self._warned_speculative_streaming:
-            self._warned_speculative_streaming = True
-            log.warning("speculative_generation is not supported for streaming; falling back to sequential")
+        if self._speculative_generation:
+            warnings.warn("speculative_generation is not supported for streaming; falling back to sequential")
         self._validate_streaming_with_output_rails()
 
         if include_metadata and self._has_streaming_output_rails:
--- a/nemoguardrails/guardrails/iorails.py
+++ b/nemoguardrails/guardrails/iorails.py
@@ -343,14 +343,13 @@ class IORails:
     ) -> Optional[LLMResponse]:
         """Speculative path: input rails and LLM generation race concurrently."""
         log.info("[%s] Speculative generation: launching input rails + LLM concurrently", req_id)
-        t0 = time.monotonic()
 
         rails_task = asyncio.create_task(self.rails_manager.is_input_safe(messages))
         gen_task = asyncio.create_task(self.engine_registry.model_call("main", messages, **llm_kwargs))
 
         try:
             response = await self._parallel_input_rail_and_response_generation(
-                rails_task, gen_task, req_id, t0, request_span
+                rails_task, gen_task, req_id, request_span
             )
         except BaseException as outer_exc:
             for t in (rails_task, gen_task):
@@ -382,7 +381,6 @@ class IORails:
         rails_task: asyncio.Task,
         gen_task: asyncio.Task,
         req_id: str,
-        t0: float,
         request_span: Optional["Span"] = None,
     ) -> Optional[LLMResponse]:
         """Race input rails against LLM generation, return LLMResponse or None (rejected)."""
--- a/nemoguardrails/guardrails/iorails.py
+++ b/nemoguardrails/guardrails/iorails.py
@@ -396,7 +396,6 @@ class IORails:
 
         if rails_task in done:
             input_result = rails_task.result()
-            rails_elapsed = time.monotonic() - t0
 
             if not input_result.is_safe:
                 log.info("[%s] Input blocked (speculative): %s", req_id, input_result.reason)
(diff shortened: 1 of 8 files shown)


nemoguardrails/guardrails/iorails.py
```diff
@@ -194,6 +195,104 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+            input_result = await rails_task
+
+            if not input_result.is_safe:
+                log.info("[%s] Input blocked (speculative, gen-first): %s", req_id, input_result.reason)
+                if self._metrics_enabled:
+                    record_request_blocked(RailDirection.INPUT)
+                self._set_speculative_span_attrs(request_span, first_completed, "input_rails", 0.0)
+                return None
+
+            self._set_speculative_span_attrs(request_span, first_completed, "none", gen_elapsed * 1000)
```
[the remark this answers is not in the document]
hazai: Yes, you still wait for rails, but without speculative mode you'd wait for rails and then wait for generation. The generation time was fully overlapped, so it is the saving.
`time_saved_ms` measures how much wall-clock time speculative mode saved vs. sequential execution `(rails_time + gen_time vs. max(rails_time, gen_time))`. When generation finishes first, the saved time is `gen_time` — that's the work that ran concurrently instead of sequentially. The current formula `(min(rails_time, gen_time))` is correct for both cases.

nemoguardrails/guardrails/iorails.py
```diff
@@ -194,6 +195,104 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+
+            if not input_result.is_safe:
+                log.info("[%s] Input blocked (speculative): %s", req_id, input_result.reason)
+                gen_task.cancel()
+                with suppress(asyncio.CancelledError):
+                    await gen_task
+                if self._metrics_enabled:
+                    record_request_blocked(RailDirection.INPUT)
+                self._set_speculative_span_attrs(request_span, first_completed, "input_rails", 0.0)
+                return None
```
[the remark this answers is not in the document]
hazai: @greptile review the changes in the latest commits, the issue should be fixed. Can we resolve this?

nemoguardrails/guardrails/iorails.py
```diff
@@ -236,6 +335,8 @@ def stream_async(
             asyncio.QueueFull: If the streaming concurrency limit is
                 reached (load shedding).
         """
+        if self._speculative_generation:
+            log.warning("speculative_generation is not supported for streaming; falling back to sequential")
```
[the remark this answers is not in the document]
hazai: @greptile review the changes in the latest commits, the issue should be fixed. Can we resolve this?

nemoguardrails/guardrails/iorails.py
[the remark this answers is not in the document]
hazai: @greptile pls resolve this issue

nemoguardrails/tracing/constants.py
```diff
@@ -127,6 +127,12 @@ class GuardrailsAttributes:
     # llm attributes (application-level, not provider-level)
     LLM_CACHE_HIT = "llm.cache.hit"
 
+    # speculative generation attributes
+    SPECULATIVE_MODE_ACTIVE = "speculative.mode_active"
```
tgasser-nv: nit: Can you rename these `speculative_generation` rather than just `speculative`?

nemoguardrails/guardrails/iorails.py
```diff
@@ -300,6 +301,108 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+                if not t.done():
+                    t.cancel()
+            # Drain all tasks (including done) to retrieve exceptions and avoid
+            # asyncio "Task exception was never retrieved" warnings.
+            await asyncio.gather(rails_task, gen_task, return_exceptions=True)
+            raise
+
+        return response_text
+
+    async def _race_rails_and_generation(
```
tgasser-nv: nit: Can you call this `_parallel_input_rail_and_response_generation()`? It sounds like we're creating a race condition intentionally

nemoguardrails/guardrails/iorails.py
```diff
@@ -300,6 +301,108 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+        rails_task: asyncio.Task,
+        gen_task: asyncio.Task,
+        req_id: str,
+        t0: float,
+        request_span: Optional["Span"] = None,
+    ) -> Optional[str]:
+        """Race input rails against LLM generation, return response text or None (rejected)."""
+        done, pending = await asyncio.wait({rails_task, gen_task}, return_when=asyncio.FIRST_COMPLETED)
+
+        first_completed = "input_rails" if rails_task in done else "generation"
```
tgasser-nv: If we're using strings like `input_rails` and `generation` that are reported out on span attributes, they should be defined in `nemoguardrails/tracing/constants.py`.

nemoguardrails/guardrails/iorails.py
```diff
@@ -300,6 +301,108 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+                self._set_speculative_span_attrs(request_span, first_completed, "input_rails", 0.0)
+                return None
+
+            self._set_speculative_span_attrs(request_span, first_completed, "none", gen_elapsed * 1000)
+
+        log.debug("[%s] Main LLM response: %s", req_id, truncate(response_text))
+        return response_text
+
+    @staticmethod
+    def _set_speculative_span_attrs(
```
tgasser-nv: This telemetry-related helper belongs in nemoguardrails/guardrails/telemetry.py rather than iorails.py. Could you move it?

nemoguardrails/guardrails/iorails.py
```diff
@@ -102,6 +106,8 @@ def __init__(self, config: RailsConfig) -> None:
             output_parallel=config.rails.output.parallel or False,
             tracer=self._tracer,
         )
+        self._speculative_generation = config.rails.input.speculative_generation or False
+        self._warned_speculative_streaming = False
```
tgasser-nv: Recommend removing this (see below line 449)

nemoguardrails/guardrails/iorails.py
```diff
@@ -342,6 +445,9 @@ def stream_async(
             asyncio.QueueFull: If the streaming concurrency limit is
                 reached (load shedding).
         """
+        if self._speculative_generation and not self._warned_speculative_streaming:
```
tgasser-nv: There's no need for the `self._warned_speculative_streaming` bool, the Python `warnings` does this already. The warning below will only fire once.

```python
  if self._speculative_generation:
      warnings.warn(
          "speculative_generation is not supported for streaming; falling back to sequential",
          stacklevel=2,
      )
```

nemoguardrails/guardrails/iorails.py
```diff
@@ -300,6 +301,108 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+    async def _race_rails_and_generation(
+        self,
+        rails_task: asyncio.Task,
+        gen_task: asyncio.Task,
+        req_id: str,
+        t0: float,
+        request_span: Optional["Span"] = None,
+    ) -> Optional[str]:
+        """Race input rails against LLM generation, return response text or None (rejected)."""
+        done, pending = await asyncio.wait({rails_task, gen_task}, return_when=asyncio.FIRST_COMPLETED)
```
tgasser-nv: Nit: I don't think `pending` is used, replace with underscore?

nemoguardrails/guardrails/iorails.py
```diff
@@ -300,6 +301,108 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+            rails_elapsed = time.monotonic() - t0
+
+            if not input_result.is_safe:
+                log.info("[%s] Input blocked (speculative): %s", req_id, input_result.reason)
+                gen_task.cancel()
+                # Use gather(return_exceptions=True) instead of bare await: when both
+                # tasks finish simultaneously, gen_task may hold a stored exception that
+                # would leak through suppress(CancelledError). gather drains it safely.
+                gen_result = (await asyncio.gather(gen_task, return_exceptions=True))[0]
+                if isinstance(gen_result, Exception) and not isinstance(gen_result, asyncio.CancelledError):
```
tgasser-nv: Need to check against `BaseException` to catch KeyboardInterrupt also

nemoguardrails/guardrails/iorails.py
```diff
@@ -300,6 +301,108 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+
+        try:
+            response_text = await self._race_rails_and_generation(rails_task, gen_task, req_id, t0, request_span)
+        except BaseException:
+            for t in (rails_task, gen_task):
+                if not t.done():
+                    t.cancel()
+            # Drain all tasks (including done) to retrieve exceptions and avoid
+            # asyncio "Task exception was never retrieved" warnings.
+            await asyncio.gather(rails_task, gen_task, return_exceptions=True)
```
tgasser-nv: This can return up to two exceptions if both rails_task and gen_task have an exception. The `raise` only applies to the first one. Recommend storing the exceptions and iterating over them, dropping the asyncio.CancelledError as on line 366

```python
  except BaseException:
      for t in (rails_task, gen_task):
          if not t.done():
              t.cancel()
      results = await asyncio.gather(rails_task, gen_task, return_exceptions=True)
      for r in results:
          if isinstance(r, BaseException) and not isinstance(r, asyncio.CancelledError):
              log.warning("[%s] Task exception suppressed during cleanup: %s", req_id, r)
      raise
```

nemoguardrails/guardrails/iorails.py
```diff
@@ -300,6 +301,108 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+    @staticmethod
+    def _set_speculative_span_attrs(
+        span: Optional["Span"], first_completed: str, first_rejector: str, time_saved_ms: float
+    ) -> None:
+        if span is None:
+            return
+        span.set_attribute(GuardrailsAttributes.SPECULATIVE_MODE_ACTIVE, True)
+        span.set_attribute(GuardrailsAttributes.SPECULATIVE_FIRST_COMPLETED, first_completed)
+        span.set_attribute(GuardrailsAttributes.SPECULATIVE_FIRST_REJECTOR, first_rejector)
+        span.set_attribute(GuardrailsAttributes.SPECULATIVE_TIME_SAVED_MS, time_saved_ms)
```
tgasser-nv: This should be an OTEL metric, not span attribute. Then we don't have cardinality in the order of each unique millisecond measurement and can aggregate into a histogram

tgasser-nv: Recommend removing in this PR `GuardrailsAttributes.SPECULATIVE_TIME_SAVED_MS` and add it in follow-on to add it to metrics.

tgasser-nv (commented):
Thanks Oren, this looks good. Most of the comments are nits and/or cleanups. High-level comments:

1. Unit-tests are broken, can you look into those?
2. Can you run with and without speculative generation and show:
  a) That both work scenarios work and produce the same results (within the non-deterministic LLM margins).
  b) The e2e latency improves by the time taken through the input rail?

There are also some gaps in test coverage:

  - Missing simultaneous-completion regression test. When asyncio.wait(FIRST_COMPLETED) returns with both tasks in done (rails reject + gen exception together), the cleanup must still drain gen_task correctly. This was the prior P1 scenario; a test would lock the fix down.
  - No span attribute assertions. Tests verify the response payload but not that speculative.mode_active, speculative.first_completed, speculative.first_rejector, and  speculative.time_saved_ms are set correctly across rails-first-pass, rails-first-reject, gen-first-pass, and gen-first-reject paths.
  - No warn-once streaming test. Behavior that stream_async warns on first call only (when speculative_generation=True) is unverified. Easy caplog or pytest.warns assertion.
  - test_rails_first_reject doesn't verify cancellation. Only checks the refusal response is returned. Should assert gen_task.cancelled() is true, or count LLM calls, to confirm
   the speculative LLM work was actually stopped rather than completing silently.

nemoguardrails/guardrails/iorails.py
hazai: Fixed

nemoguardrails/guardrails/iorails.py
hazai: Fixed

nemoguardrails/guardrails/iorails.py
hazai: Done

nemoguardrails/guardrails/iorails.py
hazai: Done

nemoguardrails/guardrails/iorails.py
hazai: Fixed

nemoguardrails/guardrails/iorails.py
hazai: Done

nemoguardrails/guardrails/iorails.py
hazai: Done, and added a comment to add the metric in next version

nemoguardrails/tracing/constants.py
hazai: Done

nemoguardrails/guardrails/iorails.py
```diff
@@ -194,6 +195,104 @@ async def _do_generate(self, messages: LLMMessages, req_id: str, **kwargs) -> LL
...
+
+        if rails_task in done:
+            input_result = rails_task.result()
+            rails_elapsed = time.monotonic() - t0
+
+            if not input_result.is_safe:
+                log.info("[%s] Input blocked (speculative): %s", req_id, input_result.reason)
+                gen_task.cancel()
+                with suppress(asyncio.CancelledError):
+                    await gen_task
```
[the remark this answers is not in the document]
hazai: @coderabbit This is irrelevant anymore. Pls check.

nemoguardrails/guardrails/iorails.py
hazai: Good point.
I changed it a bit differently, have a look.

nemoguardrails/guardrails/iorails.py
hazai: Done

hazai (commented):
All comments were attended to.
Now only need to add the benchmark part.

nemoguardrails/tracing/constants.py
```diff
@@ -127,6 +127,15 @@ class GuardrailsAttributes:
     # llm attributes (application-level, not provider-level)
     LLM_CACHE_HIT = "llm.cache.hit"
 
+    # speculative generation attributes
+    SPECULATIVE_MODE_ACTIVE = "speculative_generation.mode_active"
+    SPECULATIVE_FIRST_COMPLETED = "speculative_generation.first_completed"
+    SPECULATIVE_FIRST_REJECTOR = "speculative_generation.first_rejector"
+    SPECULATIVE_TIME_SAVED_MS = "speculative_generation.time_saved_ms"
```
tgasser-nv: nit: Could you remove this since it isn't used in the PR?

nemoguardrails/guardrails/telemetry.py
```diff
@@ -348,6 +348,30 @@ def mark_rail_stop(span: Optional["Span"], is_safe: bool) -> None:
     span.set_attribute(GuardrailsAttributes.RAIL_STOP, True)
 
 
+def set_speculative_span_attrs(
+    span: Optional["Span"],
+    first_completed: str,
+    first_rejector: str,
+    time_saved_ms: float,
```
tgasser-nv: Can you remove this argument as we're not storing this until a later PR?

tests/guardrails/test_speculative_generation.py
```diff
@@ -0,0 +1,174 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tgasser-nv, line 1: Could you add tests to check the OTEL span attributes are set correctly? This can be done by connecting up an in-memory collector and adding assertions for each of the tests below on(`first_completed`, `first_rejector`) permutations in the OTEL data

nemoguardrails/guardrails/iorails.py
```diff
@@ -342,6 +459,8 @@ def stream_async(
             asyncio.QueueFull: If the streaming concurrency limit is
                 reached (load shedding).
         """
+        if self._speculative_generation:
+            warnings.warn("speculative_generation is not supported for streaming; falling back to sequential")
```
tgasser-nv: Can you add a stacklevel=2 to the call?

tests/guardrails/test_speculative_generation.py
```diff
@@ -0,0 +1,174 @@
...
+# SPDX-License-Identifier: Apache-2.0
```
tgasser-nv, line 2: Can you also add a test to make sure the warning only fires once (`warnings.warn()`) ?

tgasser-nv (commented):
Looks much better! Can you add tests to improve patch line coverage to 100 % (see report below, and the benchmarking results in the PR with and without spec-gen enabled?

```
-------------
nemoguardrails/guardrails/iorails.py (93.2%): Missing lines 349,388,390,409,463
nemoguardrails/guardrails/telemetry.py (50.0%): Missing lines 368-370
nemoguardrails/rails/llm/config.py (100%)
nemoguardrails/tracing/constants.py (100%)
tests/guardrails/test_data.py (100%)
tests/guardrails/test_iorails_telemetry.py (100%)
tests/guardrails/test_speculative_generation.py (97.1%): Missing lines 74,130,144
-------------
Total:   194 lines
Missing: 11 lines
Coverage: 94%
-------------
```

nemoguardrails/guardrails/iorails.py
hazai: Done. Though it might be that stacklevel=3 is more appropriate here.

nemoguardrails/tracing/constants.py
hazai: Done

nemoguardrails/guardrails/telemetry.py
hazai: Done. Also fixed some tests that were using this arg

tests/guardrails/test_speculative_generation.py
hazai: Done

tests/guardrails/test_speculative_generation.py
hazai: Added

hazai (commented):
Attended to all comments

tgasser-nv (approved):
Benchmark results look good, approving!

PR #1836, opened 2026-04-29
refactor(llm): share OpenAI reasoning-model classifier across adapters

Pouyanpi:
## Description

Move `is_openai_reasoning_model` into nemoguardrails/llm/openai_reasoning.py so the `DefaultFramework` OpenAI client and the LangChain adapter can use a single source of truth. the only behavior change is the rename from `_is_openai_reasoning_model` (private) to `is_openai_reasoning_model` (public) since callers now reach across package boundaries.

Stack:
1. share OpenAI reasoning-model classifier (this PR)
2. #1837 
3. #1838 

## Summary by CodeRabbit

* **Refactor**
  * Consolidated internal OpenAI reasoning model detection logic across modules for improved code organization. No changes to user-facing behavior—reasoning models continue to have temperature and stop parameters appropriately managed.

PR #1837, opened 2026-04-29
fix(llm): drop temperature/stop and rename max_tokens for OpenAI reasoning models

Pouyanpi:
Add `apply_openai_reasoning_overrides()` in nemoguardrails/llm/openai_reasoning.py that drops the params OpenAI reasoning models reject and renames max_tokens to max_completion_tokens. OpenAIChatModel._prepare_params calls the helper for any model is_openai_reasoning_model() flags as reasoning.

The drop list is limited to the params the live probe in `tests/integrations/langchain/test_openai_param_filter.py` confirms are rejected by gpt-5 / gpt-5.x / o-series models AND that nemoguardrails sends from internal action code: temperature and stop. Other params that some reasoning models reject (top_p, presence_penalty, etc.) are accepted by others (gpt-5.1 / 5.2 / 5.4 non-chat variants) and are left alone. logprobs and top_logprobs return account-level 403s that also fire on non-reasoning chat models, so they are not reasoning-specific signals.

The rename is what actually unblocks reasoning-model configs in the DefaultFramework path: passing max_tokens to a reasoning model previously hit HTTP 400 ("Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead."). An explicit max_completion_tokens always wins over a renamed max_tokens.

The LangChain adapter is intentionally not touched here. langchain-openai 1.1.10 already does the max_tokens -> max_completion_tokens rename and the temperature drop on its own (verified by capturing the wire body); applying our override on top would be redundant or risk double-rename.

**Stack:**
1. #1836 
2. drop / remap params OpenAI rejects on reasoning models (this PR)
3. #1838
 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parameter handling for OpenAI reasoning models to ensure proper compatibility and configuration across integrations.

* **Tests**
  * Added comprehensive test coverage for OpenAI reasoning model parameter handling and edge cases.

### DIFF
--- a/nemoguardrails/llm/openai_reasoning.py
+++ b/nemoguardrails/llm/openai_reasoning.py
@@ -14,6 +14,11 @@
 # limitations under the License.
 
 
+from typing import Any, Dict
+
+_REASONING_DROP_PARAMS = frozenset({"temperature", "stop"})
+
+
 def is_openai_reasoning_model(model_name: str) -> bool:
     """True for OpenAI reasoning models (o1/o3/o4 series, gpt-5+).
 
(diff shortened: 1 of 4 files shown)


nemoguardrails/llm/openai_reasoning.py
```diff
@@ -30,3 +35,34 @@ def is_openai_reasoning_model(model_name: str) -> bool:
...
+    Returns a new dict; does not mutate the input.
+
+    Drops: ``temperature``, ``stop``.
+
+    Renames: ``max_tokens`` -> ``max_completion_tokens`` (only if
+    ``max_tokens`` is present and ``max_completion_tokens`` is not, so an
+    explicit ``max_completion_tokens`` always wins).
+    """
+    out = {key: value for key, value in params.items() if key not in _REASONING_DROP_PARAMS}
+    if "max_tokens" in out and "max_completion_tokens" not in out:
```
tgasser-nv: Should we additionally check that out["max_tokens"] isn't `""` before moving it to `max_completion_tokens`? Something like:

`if "max_tokens" in out and out["max_tokens"] and "max_completion_tokens" not in out:`

nemoguardrails/llm/openai_reasoning.py
```diff
@@ -30,3 +35,34 @@ def is_openai_reasoning_model(model_name: str) -> bool:
     if name.startswith(("gpt-5.", "gpt-6")):
         return True
     return False
+
+
+def apply_openai_reasoning_overrides(params: Dict[str, Any]) -> Dict[str, Any]:
```
tgasser-nv, line 40: Would it be better to move the `is_openai_reasoning_model()` check inside this function, so we guarantee non-reasoning models are never affected?

tgasser-nv (approved):
Looks good! Just a couple of comments

nemoguardrails/llm/openai_reasoning.py
Pouyanpi: Good one 👍🏻 Made the change

nemoguardrails/llm/openai_reasoning.py
Pouyanpi: It is better to keep the symmetry with langchain/llm_adapter.py both paths gate with is_openai_reasoning_model at the call site.

PR #1838, opened 2026-04-29
test(llm): expand reasoning-model param probe + regenerate baseline

Pouyanpi:
## Description

Probe stop, temperature, and max_tokens against every chat-candidate OpenAI model returned by /v1/models, not just stop and temperature. These are the params nemoguardrails actually passes from internal action code (self_check_*, content_safety, topic_safety) and from prompt-task configs, so the probe scope matches what we actually filter.

The baseline now records per-param accepted/rejected/other verdicts and is regenerated against the current OpenAI surface (70 chat candidates). The classifier check requires every PROBE field to be present in the baseline, forcing any future probe-set change through a regeneration.

Findings from the regen worth noting:
- max_tokens is universally rejected by OpenAI reasoning models (gpt-5, gpt-5.x, o-series); they require max_completion_tokens. This validates the rename in #1837  .
- temperature is rejected by gpt-5 / gpt-5-chat-latest / gpt-5.1-chat / gpt-5.5 / o-series, but accepted by gpt-5.1 / 5.2 / 5.4 (non-chat variants). The classifier still strips it defensively.
- stop is rejected by most reasoning models, accepted by o1 / o3-mini.

The probe deliberately omits top_p, presence_penalty, frequency_penalty, logprobs, top_logprobs, and logit_bias. nemoguardrails never sends those from internal code, and the logprobs rejection in particular turned out to be account-level (also fires on gpt-5-chat-latest, a non-reasoning model) rather than reasoning-specific.

**Stack:**
1. #1836 
2. #1837 
3. expand the live param-filter probe (this PR)

## Summary by CodeRabbit

* **Tests**
  * Enhanced OpenAI parameter validation testing to probe a broader set of chat-completions parameters.
  * Improved error classification logic for better detection and handling of parameter rejection messages.
  * Updated test data to reflect expanded parameter coverage and validation.

### DIFF
--- a/tests/integrations/langchain/test_openai_param_filter.py
+++ b/tests/integrations/langchain/test_openai_param_filter.py
@@ -206,21 +206,22 @@ def _classify(resp: httpx.Response, param: str) -> str:
     message = err.get("message", "") or ""
     if err.get("param") == param or f"'{param}'" in message:
         return "rejected"
-    # 403 / param=None case: rejection mentions the param by bare name with no
-    # quotes, e.g. "You are not allowed to request logprobs from this model".
+    # Bare-keyword case: rejection mentions the param without single quotes
+    # and ``err.param`` is None, e.g. 403 "You are not allowed to request
+    # logprobs from this model" or 400 "specify max_completion_tokens rather
+    # than max_tokens". We require both the keyword and a rejection-shaped
+    # phrase so requirement-shaped messages ("max_tokens parameter is
+    # required") fall through to accepted-by-inference instead.
     msg_lower = message.lower()
-    if re.search(rf"\b{re.escape(param)}\b", msg_lower) and (
-        "not allowed" in msg_lower or "not supported" in msg_lower
-    ):
-        return "rejected"
-    # Guard: the accepted-by-inference fallback below treats any mention of
-    # max_tokens as a generation-limit error, which only makes sense for
-    # probes that send max_completion_tokens=1 in the body. For the
-    # max_tokens probe itself, a message naming max_tokens that did not
-    # match the earlier rejection patterns is safer to flag as rejected
-    # than to silently accept (e.g. "Please specify max_completion_tokens
-    # rather than max_tokens").
-    if param == "max_tokens" and re.search(r"\bmax_tokens\b", msg_lower):
+    rejection_indicators = (
+        "not allowed",
+        "not supported",
+        "rather than",
+        "instead",
+        "unsupported",
+        "deprecated",
+    )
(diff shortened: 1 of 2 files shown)


tests/integrations/langchain/test_openai_param_filter.py
```diff
@@ -169,13 +203,20 @@ def _classify(resp: httpx.Response, param: str) -> str:
...
+    msg_lower = message.lower()
+    if re.search(rf"\b{re.escape(param)}\b", msg_lower) and (
+        "not allowed" in msg_lower or "not supported" in msg_lower
+    ):
+        return "rejected"
+    # OpenAI validates params before generation, so a max_completion_tokens /
+    # max_tokens / output-limit error on a 1-token probe means the param was
+    # accepted and generation started. Treat as accepted-by-inference.
+    if "max_completion_tokens" in message or "max_tokens" in message or "output limit" in message:
         return "accepted"
```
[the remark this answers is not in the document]
Pouyanpi: fixed

Pouyanpi:
@greptileai re your latest comment (code suggestion) I don't think OpenAI does this today. Their unsupported_parameter errors historically name both the offending param and the canonical replacement, almost always quoted. looking at common occurrences: every reported gpt-5 / o-series rejection on the API given our test probe seen names max_tokens explicitly

tests/integrations/langchain/test_openai_param_filter.py
```diff
@@ -169,13 +203,29 @@ def _classify(resp: httpx.Response, param: str) -> str:
...
+    ):
+        return "rejected"
+    # Guard: the accepted-by-inference fallback below treats any mention of
+    # max_tokens as a generation-limit error, which only makes sense for
+    # probes that send max_completion_tokens=1 in the body. For the
+    # max_tokens probe itself, a message naming max_tokens that did not
+    # match the earlier rejection patterns is safer to flag as rejected
+    # than to silently accept (e.g. "Please specify max_completion_tokens
+    # rather than max_tokens").
+    if param == "max_tokens" and re.search(r"\bmax_tokens\b", msg_lower):
```
tgasser-nv: Would this also pick up non-rejection messages like "max_tokens parameter is required"?

tgasser-nv (approved):
Looks good. Would be great to have this run in CI regularly to pick up models as they get released. A small nit: The `openai_reasoning_probe_baseline.json` has more than reasoning models in it, it's an exhaustive list of OpenaI models. Maybe rename to just `openai_probe_baseline.json`

tests/integrations/langchain/test_openai_param_filter.py
Pouyanpi: I expanded the list, should cover such cases now 👍🏻

PR #1840, opened 2026-04-30
fix(llm/clients): retry on stale event loop binding

Pouyanpi:
## Description

`httpx.AsyncClient` binds its connection pool to the first event loop on first I/O. Any client reused on a new loop (pytest-asyncio function-scope, repeated `asyncio.run` on the same `LLMRails`) crashes with `RuntimeError: Event loop is closed`.

The bug is not unique to NeMo Guardrails. The exact same defect exists in:
- raw `httpx.AsyncClient`
- `openai.AsyncOpenAI` with `max_retries=0`
- `langchain` `ChatOpenAI` with `max_retries=0`

It is masked in production by the openai SDK's default `max_retries=2` plus its broad `except Exception:` retry. Our `BaseClient` had `max_retries=2` too but only caught `httpx.TimeoutException` and `httpx.NetworkError`, so the `RuntimeError` propagated unmasked.

The fix mirrors the openai SDK's behavior: catch the stale-loop `RuntimeError` (gated on the message text) inside the existing retry loop. Six lines of code per request method, plus a `log.warning` on the recovery branch.

### 1. Raw httpx — the root cause

```python
# repro_httpx.py
import asyncio, httpx

client = httpx.AsyncClient(timeout=httpx.Timeout(10.0, connect=5.0))

async def fetch(label):
    try:
        r = await client.get(
            "https://api.openai.com/v1/models",
            headers={"Authorization": "Bearer fake"},
        )
        print(f"{label}: status={r.status_code}")
    except Exception as exc:
        print(f"{label}: FAIL {type(exc).__name__}: {exc}")

asyncio.run(fetch("call 1"))
asyncio.run(fetch("call 2"))
```

```
call 1: status=401
call 2: FAIL RuntimeError: Event loop is closed
```

### 2. openai SDK with retries=0

```python
# repro_openai.py
import asyncio
from openai import AsyncOpenAI

client = AsyncOpenAI(max_retries=0)

async def call(label):
    try:
        r = await client.chat.completions.create(
            model="gpt-4o-mini",
            messages=[{"role": "user", "content": "hi"}],
            max_completion_tokens=3,
        )
        print(f"{label}: OK -> {r.choices[0].message.content!r}")
    except Exception as exc:
        print(f"{label}: FAIL -> {type(exc).__name__}: {exc}")
        if exc.__cause__:
            print(f"  caused by: {type(exc.__cause__).__name__}: {exc.__cause__}")

asyncio.run(call("call 1"))
asyncio.run(call("call 2"))
```

```
call 1: OK -> 'Hello!'
call 2: FAIL -> APIConnectionError: Connection error.
  caused by: RuntimeError: Event loop is closed
```

### 3. LangChain ChatOpenAI with retries=0

```python
# repro_langchain.py
import asyncio
from langchain_openai import ChatOpenAI
from langchain_core.messages import HumanMessage

llm = ChatOpenAI(model="gpt-4o-mini", max_retries=0)

async def call(label):
    try:
        r = await llm.ainvoke([HumanMessage(content="say ok")])
        print(f"{label}: OK -> {r.content!r}")
    except Exception as exc:
        print(f"{label}: FAIL -> {type(exc).__name__}: {exc}")
        if exc.__cause__:
            print(f"  caused by: {type(exc.__cause__).__name__}: {exc.__cause__}")

asyncio.run(call("call 1"))
asyncio.run(call("call 2"))
```

```
call 1: OK -> 'Ok'
call 2: FAIL -> APIConnectionError: Connection error.
  caused by: RuntimeError: Event loop is closed
```

The same `RuntimeError: Event loop is closed` is the underlying cause in all three cases. The wrapping is cosmetic.

## Why the bug is invisible at default settings

Same scripts as above, but at the library defaults:

```python
client = AsyncOpenAI()                       # default max_retries=2
llm = ChatOpenAI(model="gpt-4o-mini")        # default max_retries=2
```

Both produce:

```
call 1: OK
call 2: OK
```

The openai SDK's `_request` method catches **`Exception`** (not just httpx errors) inside its retry loop:

```python
# openai/_base_client.py (around line 1623)
try:
    response = await self._client.send(request, ...)
except httpx.TimeoutException as err:
    if remaining_retries > 0:
        await self._sleep_for_retry(...); continue
    raise APITimeoutError(...) from err
except Exception as err:                # this catches RuntimeError too
    if remaining_retries > 0:
        await self._sleep_for_retry(...); continue
    raise APIConnectionError(...) from err
```

When the first attempt raises `RuntimeError: Event loop is closed`, the broad catch retries. httpx's connection pool detects the stale connection on retry and opens a fresh one in the running loop. The user sees a successful call.

LangChain `ChatOpenAI` delegates to `AsyncOpenAI`, inheriting the same retry behavior.

NeMo Guardrails' `DefaultFramework` previously caught only the two httpx-specific exceptions, so `RuntimeError` propagated past the retry loop and surfaced as a hard failure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of transient stale event loop errors with automatic retry logic and exponential backoff.
  * Enhanced error reporting for event loop-related connection failures.

* **Tests**
  * Added comprehensive test coverage for retry behavior on event loop errors across streaming and non-streaming scenarios.

### DIFF
--- a/nemoguardrails/llm/clients/base.py
+++ b/nemoguardrails/llm/clients/base.py
@@ -45,6 +45,33 @@ from nemoguardrails.llm.clients.constants import (
 log = logging.getLogger(__name__)
 
 
+_STALE_LOOP_MARKERS = (
+    "event loop is closed",
+    "different loop",
+    "different event loop",
+)
+
+
+def _is_stale_loop_error(err: BaseException) -> bool:
+    """Return True if ``err`` is a RuntimeError caused by a stale loop binding.
+
+    CPython raises one of several messages when an httpx transport (or any
+    asyncio primitive it owns) is reused on a loop other than the one it
+    was created on:
+      * 'Event loop is closed' (BaseEventLoop._check_closed) when the
+        original loop has been closed.
+      * '<asyncio.X object> is bound to a different event loop' (Lock /
+        Event / Semaphore / Queue) when the original loop is still alive.
+      * 'got Future attached to a different loop' / 'Task got Future ...'
+        when a Future created in another loop is awaited.
+
+    All three are transient: httpx invalidates the stale primitive on retry
+    and rebuilds it in the running loop.
+    """
+    text = str(err).lower()
+    return any(marker in text for marker in _STALE_LOOP_MARKERS)
+
+
 @dataclass(frozen=True)
 class HTTPResponse:
     body: Dict[str, Any]
@@ -182,6 +209,16 @@ class BaseClient:
                     retries_attempted += 1
                     continue
                 raise LLMConnectionError(0, f"Connection error: {err}", **ctx.as_kwargs()) from err
+            except RuntimeError as err:
+                if not _is_stale_loop_error(err):
+                    raise
+                if retries_remaining > 0:
+                    log.warning("Retrying after stale event loop binding: %s", err)
+                    await self._sleep_for_retry(retries_attempted)
+                    retries_remaining -= 1
+                    retries_attempted += 1
+                    continue
+                raise LLMConnectionError(0, f"Stale event loop: {err}", **ctx.as_kwargs()) from err
 
             if self._should_retry(response.status_code, response.headers) and retries_remaining > 0:
                 await self._sleep_for_retry(retries_attempted, response.headers)
@@ -272,6 +309,16 @@ class BaseClient:
             except httpx.NetworkError as err:
                 if first_yielded or retries_remaining <= 0:
                     raise LLMConnectionError(0, f"Connection error: {err}", **ctx.as_kwargs()) from err
+                await self._sleep_for_retry(retries_attempted)
+                retries_remaining -= 1
+                retries_attempted += 1
+                continue
+            except RuntimeError as err:
+                if not _is_stale_loop_error(err):
+                    raise
+                if first_yielded or retries_remaining <= 0:
+                    raise LLMConnectionError(0, f"Stale event loop: {err}", **ctx.as_kwargs()) from err
+                log.warning("Retrying stream after stale event loop binding: %s", err)
                 await self._sleep_for_retry(retries_attempted)
                 retries_remaining -= 1
                 retries_attempted += 1
--- a/tests/llm/clients/test_openai_compatible.py
+++ b/tests/llm/clients/test_openai_compatible.py
@@ -912,6 +912,28 @@ class TestNetworkExceptionRetry:
 
         assert post_calls == loops * 2
 
+    def test_recovers_with_real_httpx_against_refused_endpoint_across_asyncio_run(self):
+        """Real httpx.AsyncClient making real network attempts across asyncio.run().
+
+        httpx reliably fails on the second asyncio.run() with RuntimeError:
+        Event loop is closed. With our retry in place, the same client should surface
+        a NETWORK error on the second call (proving recovery happened and we reached the
+        connection-attempt phase), not a "Stale event loop" error (which
+        would mean retries exhausted on consecutive RuntimeErrors and
+        recovery never happened).
+        """
+        client = OpenAICompatibleClient(base_url="http://127.0.0.1:1", api_key="sk-test", max_retries=2)
+
+        with pytest.raises(LLMConnectionError) as exc1:
+            asyncio.run(client.chat_completion("gpt-4o", [{"role": "user", "content": "hi"}]))
+        assert "Stale event loop" not in str(exc1.value)
+        assert "Connection error" in str(exc1.value)
+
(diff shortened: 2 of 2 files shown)


nemoguardrails/llm/clients/base.py
```diff
@@ -182,6 +182,16 @@ async def _apost(self, path: str, payload: Dict[str, Any]) -> HTTPResponse:
                     retries_attempted += 1
                     continue
                 raise LLMConnectionError(0, f"Connection error: {err}", **ctx.as_kwargs()) from err
+            except RuntimeError as err:
+                if "event loop is closed" not in str(err).lower():
```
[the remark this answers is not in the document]
Pouyanpi: nice catch, i could repro. Pushed a "more complete" version 👍🏻

nemoguardrails/llm/clients/base.py
```diff
@@ -276,6 +286,16 @@ async def _apost_stream(self, path: str, payload: Dict[str, Any]) -> AsyncGenera
                 retries_remaining -= 1
                 retries_attempted += 1
                 continue
+            except RuntimeError as err:
+                if "event loop is closed" not in str(err).lower():
```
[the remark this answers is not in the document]
Pouyanpi: fixed it

nemoguardrails/llm/clients/base.py
```diff
@@ -45,6 +45,33 @@
...
+    asyncio primitive it owns) is reused on a loop other than the one it
+    was created on:
+      * 'Event loop is closed' (BaseEventLoop._check_closed) when the
+        original loop has been closed.
+      * '<asyncio.X object> is bound to a different event loop' (Lock /
+        Event / Semaphore / Queue) when the original loop is still alive.
+      * 'got Future attached to a different loop' / 'Task got Future ...'
+        when a Future created in another loop is awaited.
+
+    All three are transient: httpx invalidates the stale primitive on retry
```
tgasser-nv, line 68: Should we manually drop the current `self._client` and create a new one rather than relying on httpx to do this? This behaviour could change in the future

tests/llm/clients/test_openai_compatible.py
```diff
@@ -641,6 +641,277 @@ async def always_connect_error(*args, **kwargs):
         assert call_count == 2
         assert isinstance(exc_info.value.__cause__, httpx.ConnectError)
 
+    @pytest.mark.asyncio
+    async def test_runtime_error_retries_then_succeeds(self):
```
tgasser-nv, line 645: Could you add a test with a real AsyncClient and mocked HTTP Transport which can detect when it's called from a different loop? Something like:

```python
  class _LoopBoundTransport(httpx.AsyncBaseTransport):
      def __init__(self):
          self._lock = None
          self.calls = 0

      async def handle_async_request(self, request: httpx.Request) -> httpx.Response:
          if self._lock is None:
              self._lock = asyncio.Lock()  # Bound to whatever loop is running now.
          async with self._lock:
              self.calls += 1
              return httpx.Response(200, json=ok_response(), request=request)


  def test_recovers_with_real_httpx_client_across_asyncio_run_boundary():
      transport = _LoopBoundTransport()
      http_client = httpx.AsyncClient(transport=transport)
      client = make_client(http_client=http_client)

      # First loop: binds the transport's lock, then asyncio.run closes the loop.
      result1 = asyncio.run(
          client.chat_completion("gpt-4o", [{"role": "user", "content": "hi"}])
      )
      assert result1.body["choices"][0]["message"]["content"] == "Hello"

      # Second loop: same cached client, same transport, lock still bound to
      # the now-closed loop. Must succeed.
      result2 = asyncio.run(
          client.chat_completion("gpt-4o", [{"role": "user", "content": "hi"}])
      )
      assert result2.body["choices"][0]["message"]["content"] == "Hello"

      assert transport.calls >= 2
```

tgasser-nv (approved):
Looks good, left notes on the current httpx implementation changing in future, and a test which includes a real `httpx.AsyncClient`

tests/llm/clients/test_openai_compatible.py
Pouyanpi: Tried this. The proposed transport with a single asyncio.Lock doesn't reproduce the bug it's meant to catch.

I'll add integrations tests

nemoguardrails/llm/clients/base.py
Pouyanpi: I think the httpx behavior we rely on isn't really httpx's but python's asyncio. let's not do this speculative future proofing right now. The integration tests I'm going to add should cover such regression. Given the new test if a future httpx version (or Python release) stops rebuilding stale primitives on retry, the assertion flips immediately.

PR #1862, opened 2026-05-07
fix(iorails): strip trailing /v1 from base_url to avoid doubled path

### LINKED ISSUE #1861: bug: IORails ModelEngine double /v1 in URL when base_url includes /v1

3.12

### Operating system/version

26.4

_No response_

### Describe the bug

`ModelEngine._resolve_base_url()` returns the user-provided `base_url` as-is, and `_prepare_request()` appends `/v1/chat/completions` to it. When the user sets `base_url` to an OpenAI-compatible endpoint that already includes `/v1` (the standard convention for vLLM, LiteLLM, etc.), the constructed URL becomes `/v1/v1/chat/completions`, resulting in a 404.

This only affects IORails — LLMRails uses the LangChain OpenAI client which appends only `/chat/completions` to the base URL, so the same config works fine under LLMRails.

`_CHAT_COMPLETIONS_ENDPOINT = "/v1/chat/completions"`, while the default `_ENGINE_BASE_URLS`  intentionally omit `/v1`. The mismatch breaks user-provided URLs that follow the LLMRails / OpenAI SDK convention of including `/v1`.

### Steps To Reproduce

1.  Create a config with `base_url` ending in `/v1` and flows that trigger IORails
2. Use the Guardrails Python API:

```python
import asyncio
from nemoguardrails.guardrails.guardrails import Guardrails

async def main():
config = RailsConfig.from_path("path/to/config")
g = Guardrails(config=config)
print(f"Engine: {type(g.rails_engine).__name__}") # IORails
await g.startup()
result = await g.generate_async(
messages=[{"role": "user", "content": "Hello"}]
)
print(result)

asyncio.run(main())
```

3. Observe the log output showing the doubled path:

```
HTTP POST https://my-openai-compatible-server.com/v1/v1/chat/completions model='my-model'
HTTP 404 from model 'my-model': {"detail":"Not Found"}
```

### Expected Behavior

The URL should be `https://my-openai-compatible-server.com/v1/chat/completions` — a single `/v1` prefix — regardless of whether the user includes `/v1` in `base_url` or not.

### Actual Behavior

IORails constructs URL with double `/v1`: `https://my-openai-compatible-server.com/v1/v1/chat/completions`, which returns 404 from the upstream server.

The same config works correctly under LLMRails because the LangChain OpenAI client only appends `/chat/completions`.

SAY-5:
## Description

`ModelEngine._prepare_request()` appends `/v1/chat/completions` to `self.base_url`, so a user-supplied `base_url` that already ends in `/v1` (the OpenAI / LLMRails / vLLM convention) produced `/v1/v1/chat/completions` and a 404 from the upstream server. The same config worked under LLMRails because the LangChain OpenAI client only appends `/chat/completions`.

This strips an optional trailing `/v1` from `base_url` before composing the URL, so both `https://host/v1` and `https://host` produce the correct `https://host/v1/chat/completions`.

## Related Issue(s)

- Fixes #1861

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where API endpoint paths could duplicate the `/v1` segment when users included it in their base URL configuration, ensuring the final request URL consistently targets `/v1/chat/completions`.

* **Tests**
  * Added unit and end-to-end tests verifying base URL normalization across various inputs (trailing slashes, `/v1` present/absent) to prevent double `/v1` segments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Guardrails/pull/1862)

### DIFF
--- a/nemoguardrails/guardrails/model_engine.py
+++ b/nemoguardrails/guardrails/model_engine.py
@@ -233,7 +233,13 @@ class ModelEngine(BaseEngine):
     def _prepare_request(self, messages: LLMMessages, **kwargs: Any) -> _RequestParams:
         """Build the client, URL, headers, and body common to every request."""
         client = cast(RetryClient, self._client)
-        url = self.base_url.rstrip("/") + _CHAT_COMPLETIONS_ENDPOINT
+        # Strip an optional trailing "/v1" from base_url so users can follow the
+        # OpenAI / LLMRails convention of including "/v1" in base_url without
+        # producing a doubled "/v1/v1/chat/completions" path.
+        base = self.base_url.rstrip("/")
+        if base.endswith("/v1"):
+            base = base[:-3]
+        url = base + _CHAT_COMPLETIONS_ENDPOINT
 
         headers: dict[str, str] = {"Content-Type": "application/json"}
         if self.api_key:
--- a/nemoguardrails/guardrails/model_engine.py
+++ b/nemoguardrails/guardrails/model_engine.py
@@ -182,19 +182,28 @@ class ModelEngine(BaseEngine):
         )
 
     def _resolve_base_url(self) -> str:
-        """Resolve the base URL from model parameters or engine type."""
+        """Resolve the base URL from model parameters or engine type.
+
+        Strips an optional trailing "/v1" so users can follow the OpenAI / LLMRails
+        convention of including "/v1" in base_url without producing a doubled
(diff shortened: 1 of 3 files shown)


tgasser-nv (approved):
Thanks a lot for your contribution @SAY-5 !

I added a couple of commits to move the base url parsing into the init-time `_resolve_base_url()` rather than inference-time `_prepare_request()` and add some extra test coverage.

PR #1857, opened 2026-05-06
docs(custom-initialization): add customLLM and customFramework guides

Pouyanpi:
## Description

add customLLM and customFramework guides

## Documentation
  - Added guide for implementing custom LLM frameworks to integrate alternative LLM stacks.
  - Added guide for implementing custom LLM models for non-OpenAI-compatible backends.
  - Updated configuration documentation with new customization guides and navigation.
  - Added Python API reference section for extensibility types and protocols.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Guardrails/pull/1857)

PR #1863, opened 2026-05-07
feat(llm/frameworks): validate framework on registration

Pouyanpi:
## Summary

`register_framework()` now validates two invariants before storing the instance:
1. The object structurally matches the `LLMFramework` Protocol (`@runtime_checkable` `isinstance` check covers `create_model`, `register_provider`, `get_provider_names`, `reset`).
2. Its `reset` attribute is an `async` coroutine function (the registry awaits it directly).

Both raise `TypeError` with a clear message at registration time, before the framework is ever exercised.

## Why

The docs in `docs/configure-rails/custom-initialization/custom-llm-framework.md` (PR #1857) describe these guarantees in a "Failure Modes" section. Without this change those claims are false: `register_framework("bad", BadFramework())` silently accepts a sync-`reset` framework today and the user only sees a confusing failure deep inside the runtime.

## Test plan

- [x] `pytest tests/llm/frameworks/test_registry.py` — 19 passed (existing tests + 3 new failure-mode tests for sync reset, missing protocol methods, partial protocol)
- [x] Anti-example reproduction confirmed to raise the documented `TypeError`:
  - `BadFramework` (sync `reset`) -> `TypeError: Framework 'bad'.reset must be an async coroutine function.`
  - `NotAFramework` (no methods) -> `TypeError: Framework 'nope' does not implement LLMFramework. Required methods: create_model, register_provider, get_provider_names, reset.`

## Release Notes

* **Bug Fixes**
  * Enhanced framework registration validation to immediately detect and report configuration issues, including duplicate registrations and frameworks that don't meet protocol requirements. Invalid configurations now fail fast with clear error messages instead of causing cryptic failures during runtime.

### DIFF
--- a/nemoguardrails/llm/frameworks/registry.py
+++ b/nemoguardrails/llm/frameworks/registry.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import asyncio
+import inspect
 import logging
 import os
 from typing import Dict
@@ -44,11 +45,15 @@ def register_framework(name: str, framework: LLMFramework) -> None:
     if name in _frameworks:
         raise ValueError(f"Framework '{name}' is already registered.")
     if not isinstance(framework, LLMFramework):
-        raise TypeError(
-            f"Framework '{name}' does not implement LLMFramework. "
-            "Required methods: create_model, register_provider, get_provider_names, reset."
+        required = sorted(
+            getattr(
+                LLMFramework,
+                "__protocol_attrs__",
+                {"create_model", "register_provider", "get_provider_names", "reset"},
+            )
         )
-    if not asyncio.iscoroutinefunction(getattr(framework, "reset", None)):
+        raise TypeError(f"Framework '{name}' does not implement LLMFramework. Required methods: {', '.join(required)}.")
+    if not inspect.iscoroutinefunction(getattr(framework, "reset", None)):
         raise TypeError(f"Framework '{name}'.reset must be an async coroutine function.")
     _frameworks[name] = framework
 
--- a/nemoguardrails/llm/frameworks/registry.py
+++ b/nemoguardrails/llm/frameworks/registry.py
@@ -53,7 +53,7 @@ def register_framework(name: str, framework: LLMFramework) -> None:
             )
         )
         raise TypeError(f"Framework '{name}' does not implement LLMFramework. Required methods: {', '.join(required)}.")
-    if not inspect.iscoroutinefunction(getattr(framework, "reset", None)):
+    if not inspect.iscoroutinefunction(framework.reset):
         raise TypeError(f"Framework '{name}'.reset must be an async coroutine function.")
     _frameworks[name] = framework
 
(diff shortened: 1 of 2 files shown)


nemoguardrails/llm/frameworks/registry.py
```diff
@@ -27,8 +27,29 @@
...
+        TypeError: `framework` does not implement the `LLMFramework`
+            Protocol, or its `reset` is not an async coroutine function.
+    """
     if name in _frameworks:
         raise ValueError(f"Framework '{name}' is already registered.")
+    if not isinstance(framework, LLMFramework):
+        raise TypeError(
+            f"Framework '{name}' does not implement LLMFramework. "
+            "Required methods: create_model, register_provider, get_provider_names, reset."
+        )
```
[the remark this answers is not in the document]
Pouyanpi: won't fix, requires Python 3.13+ and project min is below that.

nemoguardrails/llm/frameworks/registry.py
```diff
@@ -27,8 +27,29 @@
...
+    if not asyncio.iscoroutinefunction(getattr(framework, "reset", None)):
```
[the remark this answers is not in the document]
Pouyanpi: addressed in last commit 👍🏻

nemoguardrails/llm/frameworks/registry.py
```diff
@@ -27,8 +27,29 @@
...
+        raise TypeError(f"Framework '{name}'.reset must be an async coroutine function.")
```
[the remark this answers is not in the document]
Pouyanpi: yes, addressed it in the last commit 👍🏻

nemoguardrails/llm/frameworks/registry.py
tgasser-nv: Should this use the new `register_framework()` function rather than setting the _frameworks value directly?

tgasser-nv: Should this use the new `register_framework()` function rather than setting the _frameworks value directly?

nemoguardrails/llm/frameworks/registry.py
```diff
@@ -27,8 +28,33 @@
...
+    if not isinstance(framework, LLMFramework):
+        required = sorted(
+            getattr(
+                LLMFramework,
+                "__protocol_attrs__",
+                {"create_model", "register_provider", "get_provider_names", "reset"},
+            )
+        )
+        raise TypeError(f"Framework '{name}' does not implement LLMFramework. Required methods: {', '.join(required)}.")
+    if not inspect.iscoroutinefunction(getattr(framework, "reset", None)):
```
tgasser-nv: nit: getattr isn't needed here, we know from line 47 there's a reset method on framework. Can just use framework.reset rather than getattr(framework, "reset")

tgasser-nv (approved):
LGTM! I don't think we have to worry about the change to the LLMModel protocol since we haven't released it in a new version yet.

nemoguardrails/llm/frameworks/registry.py
Pouyanpi: Better, done 👍🏻

Pouyanpi: Better, done 👍🏻

PR #1878, opened 2026-05-12
 feat(telemetry): add anonymous usage reporting 

Pouyanpi:
## Stacked PR Note

This work is split into smaller stacked PRs:

- This PR : core telemetry implementation and direct unit tests.
- #1879 : smoke driver, Kibana export verifier, fixtures, and smoke runbook. Stacked on 1878.
- #1891 : user-facing telemetry docs and README/legal notes. Stacked on 1878.

## The Complete Stack Summary

Adds anonymous NeMo Guardrails usage telemetry for process-level adoption signals, plus local/unit/smoke coverage for the emitted wire contract.

- Emit a best-effort `guardrails_usage_event` when `LLMRails` / `IORails` / `Guardrails` is instantiated.
- Emit periodic heartbeat events from the same process session.
- Capture deployment shape only: deployment type, rails engine, Colang version, rail categories/counts, LLM provider identifiers, tracing/streaming/KB flags, built-in feature identifiers, custom-flow count, platform, Python version, and package version.
- Do not collect prompts, completions, request metrics, API keys, URLs, model names, file paths, usernames, or user-provided rail/config contents.
- Send telemetry asynchronously on daemon threads.
- Support opt-out via `NEMO_GUARDRAILS_NO_USAGE_STATS=1`, `DO_NOT_TRACK=1`, and `~/.config/nemoguardrails/do_not_track`.
- Auto-disable telemetry in CI and pytest so tests do not pollute usage metrics.

## Smoke And Verification

Adds an operator-driven smoke flow for staging telemetry:

- `scripts/telemetry_smoke.py` exercises library, rich config, IORails, API server, multi-config API server, 3-worker Uvicorn API server, CLI chat, heartbeat, and opt-out scenarios.
- `scripts/kibana_verify_export.py` verifies copied Kibana Discover JSON exports against the smoke manifest.
- Smoke verification now records exact startup `sessionId` values in the manifest and prints a Kibana filter using those exact IDs.

Successful UAT smoke verification:

```text
[library_llmrails] PASS  (kibana=1, expected>=1)
[library_rich_config] PASS  (kibana=1, expected>=1)
[library_iorails] PASS  (kibana=1, expected>=1)
[server_single_config] PASS  (kibana=1, expected>=1)
[server_multi_config] PASS  (kibana=3, expected>=3)
[server_multi_worker] PASS  (kibana=3, expected>=3)
[cli_chat] PASS  (kibana=1, expected>=1)
[heartbeat] PASS  (kibana=2, expected>=2)
[opt_out_explicit] PASS  (kibana=0, expected==0)
[opt_out_ci] PASS  (kibana=0, expected==0)
[opt_out_pytest] PASS  (kibana=0, expected==0)

=== summary: 11 PASS, 0 FAIL ===
```

## Temporary CI No-Emit Check

This branch includes a temporary workflow verification hook:

- `_test.yml` sets a unique `NEMO_SESSION_PREFIX` for pytest jobs.
- `_test.yml` optionally points telemetry at `secrets.NEMO_GUARDRAILS_SMOKE_STAGING_URL`.
- CI prints the exact prefix so Kibana can be queried afterward.

Expected result after CI:

```text
client.sessionId : ci-pytest-<run_id>-<attempt>-*
```

should return zero documents. This verifies that pytest/CI suppression works even when a staging endpoint is configured.

Before final merge, decide whether to keep this as ongoing regression coverage or drop/squash the temporary workflow plumbing.

## Test Plan

Local static checks:

```bash
poetry run python -m py_compile \
  nemoguardrails/telemetry.py \
  scripts/telemetry_smoke.py \
  scripts/kibana_verify_export.py \
  tests/telemetry/test_usage_reporting.py
```

Telemetry unit tests:

```bash
poetry run pytest tests/telemetry -q
```

Observed:

```text
81 passed
```

Kibana verifier tests:

```bash
poetry run pytest tests/telemetry/test_usage_reporting.py::TestKibanaVerifyExport -q
```

Observed:

```text
3 passed
```

Local smoke sanity with intentionally unreachable loopback endpoint:

```bash
poetry run python scripts/telemetry_smoke.py \
  --staging-url http://127.0.0.1:9 \
  --run-dir /tmp/smoke-local-check
```

Expected: local smoke passes because audit/schema behavior is validated locally; network delivery is intentionally out of scope for this loopback check.

UAT smoke:

```bash
unset CI GITHUB_ACTIONS PYTEST_CURRENT_TEST
export NEMO_GUARDRAILS_SMOKE_STAGING_URL=<internal-staging-events-url>/v1.1/events/json
poetry run python scripts/telemetry_smoke.py
```

Then search Kibana with the printed exact-ID filter, for example:

```text
client.sessionId : ("<session_id_1>" or "<session_id_2>" or "<session_id_3>")
```

Copy visible rows as JSON and verify:

```bash
poetry run python scripts/kibana_verify_export.py \
  --manifest <run_dir>/manifest.json \
  --export kibana.json
```

Acceptance criteria:

- Positive scenarios have at least their expected Kibana document counts.
- Opt-out scenarios have zero Kibana documents.
- `server_multi_worker` has 3 documents from 3 distinct Uvicorn worker sessions.
- Heartbeat has startup plus heartbeat events.

## Notes For Reviewers

- Telemetry is best effort. Network failures, audit failures, and thread-start failures do not block user code.
- Session IDs are process-local UUIDs; they are not persisted across process restarts.
- The smoke scripts intentionally keep positive scenario processes alive briefly after local audit lines appear so daemon send threads have time to reach staging.
- CI and pytest suppression is covered by unit tests and smoke verification.

## Summary by CodeRabbit

* **New Features**
  * Anonymous usage telemetry now collects non-user-content data (deployment type, configurations, feature usage) to improve product insights. Full opt-out support available via environment variables or CI/test environment detection.

* **Chores**
  * Added development dependencies and comprehensive test coverage for telemetry functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Guardrails/pull/1878)


nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,999 @@
...
+    except Exception:
+        data.nemoguardrails_version = "unknown"
+
+    data.python_version = sys.version.split()[0]
+    data.platform = platform.platform()
+    data.os_name = platform.system()
+
+    if config is not None:
+        data.colang_version = getattr(config, "colang_version", ".1.0")
+
```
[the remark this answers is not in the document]
Pouyanpi: Yes already addressed in the last commit

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,999 @@
...
+        try:
+            heartbeat = GuardrailsUsageEvent(
+                timestamp=time.time(),
+                event=EventTypeEnum.HEARTBEAT,
+            )
+            heartbeat.session_id = session_id
+            _write_audit_file(heartbeat.model_dump(by_alias=True))
+            _send_report(heartbeat, server_url, client_version, session_id)
+        except Exception:
+            log.debug("Heartbeat iteration failed; loop continues", exc_info=True)
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai addressed it in last commit, do you confirm?

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,999 @@
...
+      test suite that happens to import nemoguardrails.
+
+    The intent is that adoption metrics reflect real deployments only,
+    not synthetic test/CI traffic.
+
+    Returns:
+        True if reporting should proceed, False if any opt-out is active.
+    """
+    if os.environ.get("NEMO_GUARDRAILS_NO_USAGE_STATS", "0") == "1":
+        return False
```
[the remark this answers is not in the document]
Pouyanpi: Yes, fixed in last commit please confirm @greptileai

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,999 @@
...
+    if os.environ.get("DO_NOT_TRACK", "0") == "1":
+        return False
```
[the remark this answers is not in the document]
Pouyanpi: addressed in last commit please confirm @greptileai

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,999 @@
...
+        data.nemoguardrails_version = _get_version("nemoguardrails")
+    except Exception:
+        data.nemoguardrails_version = "unknown"
+
+    data.python_version = sys.version.split()[0]
+    data.platform = platform.platform()
+    data.os_name = platform.system()
+
+    if config is not None:
+        data.colang_version = getattr(config, "colang_version", ".1.0")
```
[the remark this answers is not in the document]
Pouyanpi: addressed in last commit

.gitignore
```diff
@@ -61,7 +61,7 @@ docs/**/config
 docs/reference/python-api/generated/
 
 # Ignoring this for now
-/scripts
+# /scripts
```
tgasser-nv: nit: Can you remove the line and comment above rather than commenting out?

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,1005 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tgasser-nv, line 1: Can you rename this file to usage_report.py or something similar? Telemetry is a time-series of aggregated data from a service. This is more like a point-in-time trace of what configs people are using. It's also confusing with the actual OpenTelemetry telemetry we report (traces, logging and metrics) and the name clashes with [nemoguardrails/guardrails/ltelemetry.py](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/nemoguardrails/guardrails/telemetry.py) which does deal with telemetry

nemoguardrails/guardrails/guardrails.py
```diff
@@ -67,6 +67,11 @@ def __init__(
         use_iorails_engine = use_iorails and llm is None and self._has_only_iorails_flows()
         self._rails_engine = IORails(config) if use_iorails_engine else LLMRails(config, llm, verbose)
 
+        if use_iorails_engine:
```
tgasser-nv: Why do this in Guardrails rather than inside IORails itself (in the same way as LLMRails)? If we instantiate IORails directly for any reason, the usage-reporting will break

nemoguardrails/cli/__init__.py
```diff
@@ -107,6 +107,13 @@ def chat(
             simplify=verbose_simplify,
         )
 
+    # Claim the deployment context before LLMRails is constructed in
+    # run_chat. The subsequent report_usage call inside LLMRails.__init__
+    # will read this override and emit with deploymentType="cli".
+    from nemoguardrails.telemetry import set_deployment_type
+
+    set_deployment_type("cli")
```
tgasser-nv: Can you use enums for these string categorical values? ("cli" and "api " on line 176)?

schemas/anonymous_events.snapshot.json
```diff
@@ -0,0 +1,196 @@
...
+                    },
+                    "platform": {
+                        "type": "string",
+                        "description": "OS and architecture string (e.g. 'Linux-5.15.0-x86_64-with-glibc2.35')."
+                    },
+                    "osName": {
+                        "type": "string",
+                        "description": "Operating system name ('Darwin', 'Linux', 'Windows')."
+                    },
+                    "colangVersion": {
```
tgasser-nv, line 101: What value does this take if we're using IORails and don't have a Colang engine to report? None or a special text value (`""` or `"unknown"`)?

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,1005 @@
...
+    _lock = threading.Lock()
+    _session_uuid = None
+    _heartbeat_started = False
+
+
+if hasattr(os, "register_at_fork"):
+    os.register_at_fork(after_in_child=_reset_for_fork)
+
+
+class NemoSourceEnum(str, Enum):
```
tgasser-nv, line 154: Can you save out the data models and enums into a separate file (which is auto-generated from the schema JSON)? `usage_report_types.py` maybe? Then we know there's one file which can be changed as a result of the schema JSON changes

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,1005 @@
...
+_CONFIG_BUILTIN_FEATURE_ALIASES = {
+    "fact_checking": "factchecking",
+    "patronus": "patronusai",
+    "regex_detection": "regex",
+}
+
+_COLANG_V2_LIBRARY_DIR = Path(__file__).resolve().parent / "colang" / "v2_x" / "library"
+
+
+def _normalize_flow_name(flow_name: str) -> str:
```
tgasser-nv: Why duplicate the existing `_normalize_flow_id()` method in the [colang 1 runtime](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/nemoguardrails/colang/v1_0/runtime/flows.py#L736) and call it something different? Recommend refactoring into a colang-language-specific helper file

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,1005 @@
...
+
+    _event_name: ClassVar[str] = "guardrails_usage_event"
+
+    nemo_source: NemoSourceEnum = Field(
+        default=NemoSourceEnum.GUARDRAILS,
+        alias="nemoSource",
+        description="The NeMo product that created the event. Always 'guardrails' for this event type.",
+    )
+    session_id: str = Field(
+        default_factory=lambda: f"{os.environ.get('NEMO_SESSION_PREFIX', '')}{uuid.uuid4()}",
```
tgasser-nv: Users can put any kind of proprietary values into `NEMO_SESSION_PREFIX` which defeats the "Not traceable to any user or machine" guarantee on line 249. Need to be explicit in docs that this will be sent along with the usage data, and not to expose any confidential info

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,1005 @@
...
+    )
+    os_name: str = Field(
+        default="unknown",
+        alias="osName",
+        description='Operating system name (e.g. "Darwin", "Linux", "Windows").',
+    )
+    colang_version: str = Field(
+        default="unknown",
+        alias="colangVersion",
+        description='Colang version in use. Values: "1.0", "2.x", or "unknown" if no config.',
```
tgasser-nv, line 274: What value does this take for IORails? We still read in Colang 1 format config.yml files (so "unknown" wouldn't be valid). Also using unknown for IORails would be misleading since we do know no Colang engine is being used. Maybe add `unused` or similar?

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,1005 @@
...
+
+    Args:
+        session_id: The process-stable session ID, mirrored into the
+            heartbeat event's ``session_id`` field and the envelope.
+        client_version: Value to set as ``clientVer`` in the envelope.
+    """
+    while True:
+        time.sleep(_HEARTBEAT_INTERVAL_S)
+        try:
+            heartbeat = GuardrailsUsageEvent(
```
tgasser-nv: This GuardrailsUsageEvent will have default values for all fields not provided in the constructor here. So while _collect_usage_data() gathers all the correct fields and populates them, this heartbeat object will have all the fields set to defaults like "unknown" which contradicts that.

Recommend defining a GuardrailsHeartbeatEvent with only these fields to separate usage and heartbeat information. You'll need an extra `session_id` field to join GuardrailsUsageEvent and the subsequent heartbeats from that process in downstream analytics pipelines. That was the original proposal

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,1005 @@
...
+
+    if not _is_usage_stats_enabled():
+        return
+
+    with _lock:
+        effective_deployment_type = (
+            _deployment_type_override.value if _deployment_type_override is not None else deployment_type
+        )
+
+    try:
```
tgasser-nv: Should lines after this try .. except also be wrapped to prevent exceptions here taking down the guardrails service?

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,1005 @@
...
+        return Path(_DO_NOT_TRACK_FILE)
+    return _get_config_dir() / "do_not_track"
+
+
+def _get_usage_stats_server_url() -> str:
+    """Return the telemetry server URL, honoring runtime env overrides."""
+    return os.environ.get("NEMO_GUARDRAILS_USAGE_STATS_SERVER", _USAGE_STATS_SERVER)
+
+
+def _get_heartbeat_interval_s() -> float:
```
tgasser-nv, line 96: If we add the same deterministic delay value to all heartbeats, then we'll generate a traffic spike as the whole cluster reports at the same time. 
Recommend specifying the interval as a `HEARTBEAT_INTERVAL_MINUTES`, and then adding a random uniform seconds value in the range [0, 59]. This spreads out the heartbeat reporting over time, while ensuring each process always reports every `HEARTBEAT_INTERVAL_MINUTES`

tgasser-nv (changes requested):
High level comments

1. Can we rename this to something other than telemetry? It's confusing as we already have OpenTelemetry tracing/logs/metrics with docs. It's not telemetry a user would configure like Prometheus of a Kibana dashboard. Something like `usage_reporting`? 
2. We need to make sure the reporting code stays in-sync with`schemas/anonymous_events.snapshot.json`. At the moment we're using strings which are prone to typos with no schema enforcement. We need:
  a) A script which reads in the schema JSON and writes out a file with Python enums.
  b) A pre-commit hook which re-generates the Python enums and uses `git diff --exit-code` to force checkin of the generated file if it changed. This keeps the JSON and Python enums in-sync.
  c) Versioning control - every time we modify the schema we need to bump a version number somewhere, so when we process these files we can enforce a schema-on-read
  d) Remove all the strings and replace with enums.
3. We also have an entry-point as a RunnableRail / Langchain middleware. Should we capture this also?
4. Missing line-coverage in telemetry.py

nemoguardrails/telemetry.py
Pouyanpi: the data model is **the** source of truth, the json schema is auto generated from the data model. We also own the data model on the upstream repo. The slim snapshot of GuardrailsUsageEvent is extracted from the shared nemo-telemetry auto generated schema. To ensure that changes to the downstream data model doesn't break the upstream contract.

we have  that validate emitted pydantic datamodel payloads against that snapshot using jsonschema so the snapshot is a contract test fixture. Current unit tests and smoke tests validate emitted payloads against that snapshot via `jsonschema`. Please see TestUpstreamSchemaConformance

you can run 
```
poetry run pytest tests/telemetry/test_usage_reporting.py::TestUpstreamSchemaConformance
```

Pouyanpi:
Thanks @tgasser-nv for the review. 

> 1. Can we rename this to something other than telemetry? It's confusing as we already have OpenTelemetry tracing/logs/metrics with docs. It's not telemetry a user would configure like Prometheus of a Kibana dashboard. Something like `usage_reporting`?

I agree there is a real naming collision, and `telemetry.py` is not the most precise name if we look only at this repo. I think the better fix is to rename the unreleased IORails module rather than the anonymous usage reporting module.

I chose `nemoguardrails/telemetry.py` deliberately for cross-product consistency. the shared legal/product language calls this feature “**telemetry**”, and the other NeMo projects adopting the same anonymous usage reporting pattern also expose it under telemetry-oriented modules. For example:

- [Data Designer: telemetry model/module naming](https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/packages/data-designer-engine/src/data_designer/engine/models/telemetry.py)
- [Evaluator: `nemo_evaluator/telemetry.py`](https://github.com/NVIDIA-NeMo/Evaluator/blob/7a44667f0b4e361b410b53bbace9698b9d2d1a08/packages/nemo-evaluator/src/nemo_evaluator/telemetry.py)
- [NeMo Agent Toolkit: telemetry configuration command](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/main/packages/nvidia_nat_core/src/nat/cli/commands/configure/telemetry.py)
- [Safe Synthesizer: `nemo_safe_synthesizer/telemetry.py`](https://github.com/NVIDIA-NeMo/Safe-Synthesizer/blob/92d4375e13d610c0697b20a21194beaa20c859b5/src/nemo_safe_synthesizer/telemetry.py)

That said, the confusing part is `nemoguardrails.guardrails.telemetry`, which is actually inline OTEL instrumentation for IORails: spans, metrics, request ids, and OTEL api isolation. as that module has not shipped yet, I propose renaming it to `nemoguardrails.guardrails.otel_instrumentation` before release. That keeps the usage-reporting module aligned with the NeMo suite while removing the local ambiguity with tracing/metrics instrumentation.

> 2. We need to make sure the reporting code stays in-sync with`schemas/anonymous_events.snapshot.json`. At the moment we're using strings which are prone to typos with no schema enforcement. We need:


agreed that the reporting code needs to stay in sync with `schemas/anonymous_events.snapshot.json`. The reason this PR uses a checked-in snapshot plus `jsonschema` validation is that the canonical schema/data model lives in the shared `nemo-telemetry` repo; this repository should not become a second source of truth for that model. The snapshot is here as a contract fixture so changes in `nemoguardrails/telemetry.py` are caught by tests.
for more context see https://github.com/NVIDIA-NeMo/Guardrails/pull/1878#discussion_r3246534063 

[a side note: there is a very demanding process to update the schema and we will barely touch it]

> 3. We also have an entry-point as a RunnableRail / Langchain middleware. Should we capture this also? 

No, it is not a missing event: both construct `LLMRails`, so reporting fires. If needed we should update the schema for future releases.

nemoguardrails/guardrails/guardrails.py
Pouyanpi: Fair point, I somewhat assumed no one should directly import IORails that's why we have the facade. But changed it 👍🏻

nemoguardrails/telemetry.py
Pouyanpi: see my comment here https://github.com/NVIDIA-NeMo/Guardrails/pull/1878#issuecomment-4457842064

nemoguardrails/telemetry.py
Pouyanpi: No, see my comment https://github.com/NVIDIA-NeMo/Guardrails/pull/1878#issuecomment-4457842064

nemoguardrails/guardrails/guardrails.py
Pouyanpi: I just realized that it exposes a small lifecycle smell: the sync api creates a temporary internal IORails instance, so we need `_report_usage=False` to avoid counting that bridge as user-created usage. 

I think this is ok for this PR because it keeps the fix scoped, but longer term it might be cleaner to separate the public construction path from internal engine creation in iorails.

nemoguardrails/telemetry.py
Pouyanpi: I think we should keep these as orthogonal fields rather than introducing `"unused"` or use `colangVersion` as a runtime marker.

`colangVersion` should represent the resolved RailsConfig or colang config language version (`RailsConfig.colang_version`, defaulting to `"1.0"`), while `railsEngine` should identify the runtime (`"LLMRails"` or `"IORails"`). So `colangVersion="1.0"` + `railsEngine="IORails"` is a valid combination: it means an IORails runtime was created from a config whose resolved Colang version is `1.0`. If Colang version alone was sufficient we didn't have to introuce `railsEngine` at all.

`"unknown"` should stay reserved for events that genuinely have no config context, such as server lifespan startup.

[ I also moved IORails reporting into `IORails.__init__`, so direct `IORails(config)` construction reports `railsEngine="IORails"`, and the `Guardrails` wrapper path does not double-report. ]

nemoguardrails/telemetry.py
Pouyanpi: Done 👍🏻

nemoguardrails/telemetry.py
Pouyanpi: Not sure about the SMS behavior but added it, Good point 👍🏻

nemoguardrails/telemetry.py
Pouyanpi: Good catch. I updated this so heartbeats no longer create a fresh event with default `"unknown"` fields.
The loop now copies the first startup event metadata and changes only `event`, `timestamp`, and the shared `sessionId`.  that keeps heartbeats self-contained for querying without requiring a join back to the startup event. The tradeoff is repeating static metadata, but at the heartbeat interval that seems acceptable.

nemoguardrails/telemetry.py
Pouyanpi: dropped it entirely and updated smoke

schemas/anonymous_events.snapshot.json
Pouyanpi: pls see https://github.com/NVIDIA-NeMo/Guardrails/pull/1878#discussion_r3247258617

nemoguardrails/telemetry.py
Pouyanpi: I imported it from colang runtime, the refactoring can be done in a follow-up as it is already used in IORails

nemoguardrails/telemetry.py
```diff
@@ -0,0 +1,995 @@
...
+    )
+    session_id: str = Field(
+        default_factory=lambda: str(uuid.uuid4()),
+        alias="sessionId",
+        description=(
+            "Random UUID4 generated in memory at process start; acts as the session ID. "
+            "Held only for the process lifetime, never persisted to disk. Not traceable "
+            "to any user or machine."
+        ),
+    )
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai Yes the PR description needs to be updated. This is correct now, we've dropped the prefix

nemoguardrails/cli/__init__.py
tgasser-nv: @Pouyanpi Can you add the enum? Or will this be in a follow-on PR?

nemoguardrails/guardrails/guardrails.py
tgasser-nv: The sync `generate()` method is only around for backwards-compatibility. I can't imagine anyone using it in a production system where we care about their usage

nemoguardrails/telemetry.py
tgasser-nv: Ok so the `colang_version` refers to the config file, not the runtime. That makes sense.

tgasser-nv:
> Thanks @tgasser-nv for the review.
> 
> > 1. Can we rename this to something other than telemetry? It's confusing as we already have OpenTelemetry tracing/logs/metrics with docs. It's not telemetry a user would configure like Prometheus of a Kibana dashboard. Something like `usage_reporting`?
> 
> I agree there is a real naming collision, and `telemetry.py` is not the most precise name if we look only at this repo. I think the better fix is to rename the unreleased IORails module rather than the anonymous usage reporting module.
> 
> I chose `nemoguardrails/telemetry.py` deliberately for cross-product consistency. the shared legal/product language calls this feature “**telemetry**”, and the other NeMo projects adopting the same anonymous usage reporting pattern also expose it under telemetry-oriented modules. For example:
> 
> * [Data Designer: telemetry model/module naming](https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/packages/data-designer-engine/src/data_designer/engine/models/telemetry.py)
> * [Evaluator: `nemo_evaluator/telemetry.py`](https://github.com/NVIDIA-NeMo/Evaluator/blob/7a44667f0b4e361b410b53bbace9698b9d2d1a08/packages/nemo-evaluator/src/nemo_evaluator/telemetry.py)
> * [NeMo Agent Toolkit: telemetry configuration command](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/main/packages/nvidia_nat_core/src/nat/cli/commands/configure/telemetry.py)
> * [Safe Synthesizer: `nemo_safe_synthesizer/telemetry.py`](https://github.com/NVIDIA-NeMo/Safe-Synthesizer/blob/92d4375e13d610c0697b20a21194beaa20c859b5/src/nemo_safe_synthesizer/telemetry.py)
> 

Being consistent with our legal terms is more important, let's keep the (misleading and bad) `telemetry.py` name.

> That said, the confusing part is `nemoguardrails.guardrails.telemetry`, which is actually inline OTEL instrumentation for IORails: spans, metrics, request ids, and OTEL api isolation. as that module has not shipped yet, I propose renaming it to `nemoguardrails.guardrails.otel_instrumentation` before release. That keeps the usage-reporting module aligned with the NeMo suite while removing the local ambiguity with tracing/metrics instrumentation.

This is a stretch. Having Open Telemetry code in a file called `telemetry.py` isn't confusing. The module hasn't shipped, but it has been merged (unlike this PR). Given the alignment with other "telemetry" users and legal terms I'll rename the existing nemoguardrails/guardrails/telemetry.py file to open_telemetry.py.

> 
> > 2. We need to make sure the reporting code stays in-sync with`schemas/anonymous_events.snapshot.json`. At the moment we're using strings which are prone to typos with no schema enforcement. We need:
> 
> agreed that the reporting code needs to stay in sync with `schemas/anonymous_events.snapshot.json`. The reason this PR uses a checked-in snapshot plus `jsonschema` validation is that the canonical schema/data model lives in the shared `nemo-telemetry` repo; this repository should not become a second source of truth for that model. The snapshot is here as a contract fixture so changes in `nemoguardrails/telemetry.py` are caught by tests. for more context see [#1878 (comment)](https://github.com/NVIDIA-NeMo/Guardrails/pull/1878#discussion_r3246534063)
> 
> [a side note: there is a very demanding process to update the schema and we will barely touch it]

"this repository should not become a second source of truth for that model": It already is, but that's necessary since we have client and server code in separate repos.

> 
> > 3. We also have an entry-point as a RunnableRail / Langchain middleware. Should we capture this also?
> 
> No, it is not a missing event: both construct `LLMRails`, so reporting fires. If needed we should update the schema for future releases.

Sounds good.

nemoguardrails/telemetry.py
tgasser-nv: I don't see any changes in this PR. Is there a follow-on ?

tgasser-nv (approved):
Approved, I'll prepare a PR to rename nemoguardrails/guardails/telemetry.py to avoid namespace clashes.

A couple of the comments were resolved but not addressed. Please check them and let me know if there's a follow-on PR with this

nemoguardrails/cli/__init__.py
Pouyanpi: you're right, sorry! Just added it 👍🏻

nemoguardrails/telemetry.py
Pouyanpi: it is addressed by adding random jitter to heartbeat sleeps.

code does this:

```python
jitter = random.uniform(0, min(60.0, _HEARTBEAT_INTERVAL_S * 0.1))
time.sleep(_HEARTBEAT_INTERVAL_S + jitter)
```

 also there is a test named `test_heartbeat_loop_sleep_adds_jitter_without_shortening_interval`.

PR #1900, opened 2026-05-19
feat(cli): add --workers flag to server command

### LINKED ISSUE #1746: feature: Add --workers flag to nemoguardrails server for multi-worker uvicorn support
### Did you check the docs?

The server seemingly runs [uvicorn with a single worker](https://github.com/NVIDIA-NeMo/Guardrails/blob/94fc57efb4a0084c441b86a122542f8e7bf6c57e/nemoguardrails/cli/__init__.py#L194)

Add a --workers CLI option (defaulting to 1 for backward compatibility):

```
@app.command()
def server(..., workers: int = 1):
    ...
    uvicorn.run(server_app, port=port, log_level="info", host="0.0.0.0", workers=workers)
```

Presumably, we'd need to ensure that make api.py spawn-safe

* Gunicorn wrapper: can run `gunicorn -w 4 -k uvicorn.workers.UvicornWorker nemoguardrails.server.api:app`  but this bypasses the CLI's config loading IIUC

### Additional context

_No response_

sheitabrk:
Closes #1746.

Adds a `--workers` flag to `nemoguardrails server` so it can run uvicorn with multiple worker processes.

The catch with uvicorn's multi-worker mode is that each worker is spawned in its own process and re-imports the FastAPI app from scratch, so the existing setattr-based config done in the CLI process (`api.app.rails_config_path = ...`, etc.) is invisible to the workers.

To make it work end to end without breaking the single-worker flow:

- `--workers` defaults to `1`, in which case the existing setattr-based path is used unchanged.
- When `--workers > 1`, the CLI propagates config via env vars (`NEMO_GUARDRAILS_CONFIG_PATH`, `NEMO_GUARDRAILS_DEFAULT_CONFIG_ID`, `NEMO_GUARDRAILS_AUTO_RELOAD`) before importing `api`, and hands uvicorn the import string `"nemoguardrails.server.api:app"` instead of the app object.
- `api.py` now reads those same env vars at module-init time, following the existing `NEMO_GUARDRAILS_DISABLE_CHAT_UI` pattern.
- `--prefix` is rejected with `--workers > 1` because the `FastAPI().mount(...)` wrapping happens in the CLI process and does not carry across to workers; failing closed is clearer than silently dropping the prefix.

Tests added in `tests/cli/test_cli_main.py`:
- Default (`workers=1`) keeps passing the app object, no `workers` kwarg.
- `--workers=4` calls uvicorn with the import string and `workers=4`.
- `--workers=2` with `--config`, `--default-config-id`, `--auto-reload` sets the corresponding env vars.
- `--workers=2 --prefix=/api/v1` exits non-zero and never calls uvicorn.

`ruff check` and `ruff format --check` pass on all touched files. Pytest could not run locally (this environment has Python 3.9 and 3.14 installed and the project requires `>=3.10,<3.14`), so the new tests will be confirmed by CI.

## Summary by CodeRabbit

* **New Features**
  * Added `--workers` option to the server CLI command to support multi-worker deployments (default: 1, minimum: 1).
  * Server configuration now reads from environment variables, enabling settings propagation across worker processes.
  * Configuration validation: `--prefix` option is incompatible with multi-worker mode.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Guardrails/pull/1900?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/nemoguardrails/cli/__init__.py
+++ b/nemoguardrails/cli/__init__.py
@@ -149,6 +149,15 @@ def server(
         default="",
         help="A prefix that should be added to all server paths. Should start with '/'.",
     ),
+    workers: int = typer.Option(
+        default=1,
+        min=1,
+        help=(
+            "Number of uvicorn worker processes. When greater than 1, the "
+            "server is launched via an import string and CLI configuration "
+            "is propagated to workers through environment variables."
+        ),
+    ),
 ):
     """Start a NeMo Guardrails server."""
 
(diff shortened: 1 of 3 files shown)


sheitabrk:
Closing as duplicate of #1851, which I missed when I ran my pre-submit overlap check. 

#1851 already covers the core feature (--workers flag, env-var propagation for config, rejecting --prefix with multiple workers) in the same spirit as this PR.

PR #1923, opened 2026-05-23
fix(server): honor api_key env vars when overriding main model

### LINKED ISSUE #1895: bug: current version haven't handle llm provider api_key env variable

python 3.10.19

### Operating system/version

WSL Ubuntu 22.04

0.21.0

### Describe the bug

As mentioned in the [documentation](https://docs.nvidia.com/nemo/guardrails/0.21.0/run-rails/using-fastapi-server/list-models.html#provider-configuration), we can configure the main model using an environment variable, which works as expected. However, when using a provider that requires an API key, such as OpenAI, the OPENAI_API_KEY seems to be read from config.yml instead of the environment variable in `api.py`.

### Steps To Reproduce

1. config a main model with config.yml
2. config OPENAI_API_KEY, MAIN_MODEL_ENGINE="openai" and MAIN_MODEL_BASE_URL
3. send a request

### Expected Behavior

successful request from any openai-compatible model provider

### Actual Behavior

invalid api key

adityasingh2400:
Closes #1895.

When a request to `/v1/chat/completions` supplies a `model` field, the server rebuilds the main model entry by constructing a fresh `Model` from the request value plus `MAIN_MODEL_ENGINE` and `MAIN_MODEL_BASE_URL`, then hands it to `_update_models_in_config` to swap into the loaded `RailsConfig`. That swap replaced the existing entry wholesale, so any `api_key_env_var` declared in `config.yml` was silently dropped. The downstream initializer then had no env-var hint and the credential resolution path that the CLI and non-server code uses was never taken, surfacing as an invalid-API-key error from the upstream provider.

This change carries `api_key_env_var` over from the existing main model entry when the override does not set one explicitly, so a `config.yml` that points at a custom environment variable continues to resolve credentials correctly when the model name is swapped at request time. Explicit `api_key_env_var` on the override still takes precedence. Implicit env-var fallback for engines like `openai` continues to work via the default framework, which was already correct on the no-override path.

Added regression coverage on `_update_models_in_config` for both the preserve-from-config and override-wins paths, plus an integration-level test that drives `_get_rails` end to end with `OPENAI_API_KEY` set and a request-side model override, asserting the resolved HTTP client carries the env-derived key.

## Summary by CodeRabbit

* **Bug Fixes**
  * Model overrides now preserve existing credential configuration fields when not explicitly set in the override request.
  * Environment-based credential resolution continues to work correctly when applying model overrides.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Guardrails/pull/1923?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/nemoguardrails/server/api.py
+++ b/nemoguardrails/server/api.py
@@ -328,9 +328,7 @@ def _update_models_in_config(config: RailsConfig, main_model: Model) -> RailsCon
         # set, so a request-side model swap inherits credential resolution
         # (api_key_env_var) and other model-level settings from config.yml.
         api_key_env_var = (
-            main_model.api_key_env_var
-            if main_model.api_key_env_var is not None
-            else existing.api_key_env_var
+            main_model.api_key_env_var if main_model.api_key_env_var is not None else existing.api_key_env_var
         )
         models[main_model_index] = main_model
         models[main_model_index].parameters = parameters
--- a/nemoguardrails/server/api.py
+++ b/nemoguardrails/server/api.py
@@ -309,6 +309,9 @@ def _update_models_in_config(config: RailsConfig, main_model: Model) -> RailsCon
     """Update the main model in the RailsConfig.
 
     If a model with type="main" exists, it replaces it. Otherwise, adds it.
+    Fields not explicitly set on the override model (such as ``api_key_env_var``)
+    are carried over from the existing entry so request-driven model overrides
+    do not silently drop credential resolution configured in ``config.yml``.
     """
     models = config.models.copy()
     main_model_index = None
@@ -319,9 +322,15 @@ def _update_models_in_config(config: RailsConfig, main_model: Model) -> RailsCon
             break
 
     if main_model_index is not None:
-        parameters = {**models[main_model_index].parameters, **main_model.parameters}
+        existing = models[main_model_index]
+        parameters = {**existing.parameters, **main_model.parameters}
+        # Preserve fields from the existing entry that the override does not
+        # set, so a request-side model swap inherits credential resolution
+        # (api_key_env_var) and other model-level settings from config.yml.
+        api_key_env_var = main_model.api_key_env_var or existing.api_key_env_var
         models[main_model_index] = main_model
         models[main_model_index].parameters = parameters
+        models[main_model_index].api_key_env_var = api_key_env_var
     else:
         models.append(main_model)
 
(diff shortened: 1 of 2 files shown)


adityasingh2400:
Switched the override check from `or` to `is not None` so an explicitly set empty string wins over the existing entry. Leaving the broader carry-over of `mode`/`cache` and the integration test's private attribute access for a follow-up since neither was part of the original credential-drop fix and I want to keep this PR focused.

adityasingh2400:
Pushed a formatting-only commit. Lint was failing because ruff-format wanted the api_key_env_var ternary collapsed onto a single line, which fits within the 120 char limit. No logic change.

m-misiura:
@adityasingh2400 -- I also encountered this error recently, so thanks for the PR

IIUC, your fix patches `api_key_env_var` back after replacement, but `mode` and `cache` are still silently dropped. Moreover, every future field added to `Model` will need to be manually carried over here — or hit the same class of bug again.

Perhaps a more robust approach would be to start from the existing model and only override the fields the caller actually changes, using Pydantic's `model_copy(update=...)` along the lines of

```python
if main_model_index is not None:
    existing = models[main_model_index]
    models[main_model_index] = existing.model_copy(update={
        "model": main_model.model,
        "engine": main_model.engine,
        "parameters": {**existing.parameters, **main_model.parameters},
    })
else:
    models.append(main_model)
```

WDYT @adityasingh2400?

cc @Pouyanpi @tgasser-nv

adityasingh2400:
Thanks, and good catch confirming the broader problem. You are right that patching api_key_env_var back was a band aid. mode and cache were still dropped, and every new Model field would re-hit the same class of bug.

I pushed f2eeaae taking your approach. It now starts from the existing entry and applies only the fields the override actually set:

```python
override = main_model.model_dump(exclude_unset=True)
if "parameters" in override:
    override["parameters"] = {**existing.parameters, **main_model.parameters}
models[main_model_index] = existing.model_copy(update=override)
```

I used model_dump(exclude_unset=True) rather than naming model, engine, and parameters explicitly, so that any field the caller did set (including an explicit api_key_env_var, mode, or cache) still wins, while anything left unset is carried over from config.yml. Parameters keep the override-wins merge, so a request extends the config.yml parameters instead of replacing them.

I also added a regression test that sets mode=text and a cache config on the existing entry, overrides only the model name, and asserts mode, cache, api_key_env_var, and the existing parameters all survive. The earlier api_key_env_var tests still pass, and ruff is clean.

Happy to switch to an explicit field list instead if you would prefer that, just let me know.

Issue #1931, opened 2026-05-26
bug: `detect_regex_pattern()` crashes with TypeError during output streaming


m-misiura:
### Did you check docs and existing issues?

- [x] I have read all the NeMo-Guardrails docs
- [x] I have updated the package to the latest version before submitting this issue
- [x] (optional) I have used the develop branch
- [x] I have searched the existing issues of NeMo-Guardrails

### Python version (python --version)

Python 3.12

### Operating system/version

MacOS 26.5

### NeMo-Guardrails version (if you must use a specific version and not the latest

0.22

### Describe the bug

The built-in `detect_regex_pattern` action crashes with a `TypeError` when used as an output rail with output streaming enabled. The action dispatcher injects runtime parameters (`context`, `llm_task_manager`, `model_name`, `llms`, `llm`) during streaming output rail execution, but `detect_regex_pattern` does not accept `**kwargs` to absorb them.

Peer built-in actions like `detect_sensitive_data`, `self_check_input`, and `self_check_output` all accept `**kwargs` and work correctly in this scenario.

### Steps To Reproduce

- start server locally; with the following config.yaml located in e.g `./configs/regex-streaming-bug`

```yaml
models:
  - type: main
    engine: openai
    parameters:
      base_url: "INSERT_ME"
      model_name: "INSERT_ME"
      api_key: "INSERT_ME"

rails:
  output:
    flows:
      - regex check output
    streaming:
      enabled: true
      chunk_size: 200
      context_size: 50
      stream_first: true

  config:
    regex_detection:
      output:
        patterns:
          - "\\bconfidential\\b"
          - "\\bsecret\\b"

streaming: true
```

- make a request, referencing the config

```bash
curl -N -X POST http://localhost:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen3.6-35B-A3B",
    "messages": [{"role": "user", "content": "Tell me something interesting"}],
    "stream": true,
    "guardrails": {"config_id": "regex-streaming-bug"}
  }'

```

### Expected Behavior

The `regex check output` flow executes successfully during streaming, checking each output chunk against the configured regex patterns.

### Actual Behavior

We get the following error:

```
ERROR:nemoguardrails.actions.action_dispatcher:detect_regex_pattern() got an unexpected keyword argument 'context'
Traceback (most recent call last):
  File ".../nemoguardrails/actions/action_dispatcher.py", line 213, in execute_action
[...]

PR #1932, opened 2026-05-26
fix(library): `detect_regex_pattern()` crashes with TypeError during output streaming

m-misiura:
## Introduction

This PR attemps to provide a minimal fix for [GH Issue](https://github.com/NVIDIA-NeMo/Guardrails/issues/1931)

## Root cause

When streaming is enabled, output regex rails yields

```bash
ERROR:nemoguardrails.actions.action_dispatcher:detect_regex_pattern() got an unexpected keyword argument 'context'
Traceback (most recent call last):
  File ".../nemoguardrails/actions/action_dispatcher.py", line 213, in execute_action
    result = fn(**params)
             ^^^^^^^^^^^^
TypeError: detect_regex_pattern() got an unexpected keyword argument 'context'
```

The output streaming rail path in `llmrails.py` (`_prepare_params()`, ~line 1753) injects runtime params into the action call:

```python
{
    "context": context,
    "llm_task_manager": self.runtime.llm_task_manager,
    "config": self.config,
    "model_name": model_name,
    "llms": ...,
    "llm": ...,
    **action_params,
}
```

The action dispatcherpasses all params via `fn(**params)` without filtering.

The `detect_regex_pattern` signature (`regex/actions.py:32`) is:

```python
async def detect_regex_pattern(source, text, config) -> RegexDetectionResult:
```

It does not accept `**kwargs`, so the extra injected params cause a `TypeError`.

Compare with `detect_sensitive_data` (`sensitive_data_detection/actions.py`):

```python
async def detect_sensitive_data(source, text, config, **kwargs):
```

## Suggested Fix

Add `**kwargs` to `detect_regex_pattern` to match the established pattern:

```python
async def detect_regex_pattern(
    source: str,
    text: str,
    config: RailsConfig,
    **kwargs,
) -> RegexDetectionResult:
```

## Potential alternative

Later on, we might want to modify `execute_action()` in `action_dispatcher.py` to introspect the action's signature and drop params it doesn't accept, rather than forwarding all of them

## Summary by CodeRabbit

* **Refactor**
  * Updated regex detection action signature to accept additional parameters.

* **Tests**
  * Added regression test validating regex detection functionality with additional dispatcher-injected parameters.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Guardrails/pull/1932?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/nemoguardrails/library/regex/actions.py
+++ b/nemoguardrails/library/regex/actions.py
@@ -33,6 +33,7 @@ async def detect_regex_pattern(
     source: str,
     text: str,
     config: RailsConfig,
+    **kwargs,
 ) -> RegexDetectionResult:
     """Checks whether the provided text matches any forbidden regex pattern.
 
--- a/tests/test_regex_detection.py
+++ b/tests/test_regex_detection.py
@@ -19,6 +19,7 @@ from pydantic import ValidationError
 from nemoguardrails import RailsConfig
 from nemoguardrails.actions import action
 from nemoguardrails.actions.actions import ActionResult
+from nemoguardrails.library.regex.actions import detect_regex_pattern
 from tests.utils import TestChat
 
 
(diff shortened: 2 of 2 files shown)


Pouyanpi (approved):
LGTM 🚀 thanks for the fix.

m-misiura:
> LGTM 🚀 thanks for the fix.

No problem! Thanks for dealing with this so quickly

PR #1941, opened 2026-05-28
feat(library): add context bloat detection rail

MuneezaAzmat:
## Description 
Context-window manipulation is a class of attacks where adversarial content, injected via tool outputs, retrieved documents, or user input, uses padding, repetition, or excessive length to push critical instructions out of the model's attention window or exhaust its token budget. Existing NeMo Guardrails cover prompt injection and content safety, but there is no built-in defense against context bloat. This PR adds a lightweight, pure-Python guardrail that catches these attacks before they reach the LLM, with no external dependencies.

## Summary
- Adds a new `context_bloat_detection` guardrail that detects context-manipulation attacks (padded, oversized, or repetitive content in tool outputs, RAG chunks, or user input)
- Checks (cheapest first): size cap, Shannon entropy, longest char run, n-gram repetition
- Supports `reject`, `truncate`, and `warn` actions via config
- `reject` — stops the flow with a user-facing message
- `truncate` — truncates to max_chars at size cap, rejects on remaining checks, writes truncated text back to source variable
- `warn` — logs detection, does not block or modify
- Short texts (< `min_chars`, default 50) skip entropy/run/repetition checks to avoid false positives on messages like "Hi" or "Hello"
- Registers `ContextBloatDetectionConfig` Pydantic model in `RailsConfigData` with field validation (Literal action type, numeric bounds)
- Includes Colang v1 and v2 flow definitions
- Includes 36 unit tests

## Checklist


nemoguardrails/rails/llm/config.py
```diff
@@ -1189,6 +1218,11 @@ class RailsConfigData(BaseModel):
         description="Configuration for content safety rails.",
     )
 
+    context_bloat_detection: Optional[ContextBloatDetectionConfig] = Field(
+        default_factory=ContextBloatDetectionConfig,
+        description="Configuration for context bloat / context manipulation detection.",
+    )
```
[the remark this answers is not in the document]
MuneezaAzmat: Every other rail in RailsConfigData uses `Optional[...] = Field(default_factory=...)`. This is the established pattern.
  Changing just ours would be inconsistent.

nemoguardrails/library/context_bloat_detection/config.yml
```diff
@@ -0,0 +1,42 @@
+# Example config.yml for context_bloat_detection wiring.
+# Drop this into your config.yml under the existing models/rails keys.
+
+rails:
+  retrieval:
+    flows:
+      - context bloat detection on retrieval
+
+  execution:
```
Pouyanpi: Could we please drop the tool-related rails from this PR for now? That surface is not publicly released yet, so including tool rail config/examples here may confuse users. We can add it back once tool rails are public. 

Also from actions.py and colang flows and tests.

nemoguardrails/library/context_bloat_detection/actions.py
```diff
@@ -0,0 +1,211 @@
...
+
+    # Checks 2-4 are skipped for short texts to avoid false positives
+    if char_count >= cfg.min_chars:
+        # ---- 2. Entropy ----
+        entropy = _shannon_entropy(text)
+        metrics["entropy"] = round(entropy, 3)
+        if entropy < cfg.min_entropy:
+            detections.append("low_entropy")
+            if cfg.action in ("reject", "truncate"):
+                log.info(f"context bloat detected: low_entropy | entropy={entropy:.3f}")
```
Pouyanpi: In `truncate` mode, non-size hits still return `action="truncate"`. the flow only blocks `reject`, so bad content under `max_chars` can still go through unchanged. 

should these cases return `reject` instead, as the config comment says?

nemoguardrails/library/context_bloat_detection/flows.co
Pouyanpi, line 1: for the Colang 2, can we add `global` before rewriting `$user_message` / `$relevant_chunks`? 

vars are local by default in Colang 2, so without this the assignment shadows locally and downstream still sees the original text.

Pouyanpi (changes requested):
Thank you @MuneezaAzmat  for your contribution! Please see the review comments below:
(summary)
1. drop tool-related rails because not public.
2. truncate mode must reject non-size detections.
3. add global before rewriting $relevant_chunks / $user_message in Colang 2.

nemoguardrails/library/context_bloat_detection/flows.co
MuneezaAzmat: fixed

nemoguardrails/library/context_bloat_detection/config.yml
MuneezaAzmat: done

MuneezaAzmat:
Thanks for the feedback @Pouyanpi, I've addressed all three comments (summary) 
1. Tool rails removed from config.yml, both Colang flow files, and docstring references
2. Non-size detections in truncate mode now return action="reject" 
3. Global keywords declared in Colang v2 flows so truncated text is visible downstream

Pouyanpi:
@MuneezaAzmat thanks for making the changes. I've opened a review PR that does some minor cleanups. Would you please merge that? after that do a rebase on latest develop and accept both changes. I'll merge afterwards.

https://github.com/MuneezaAzmat/NeMo-Guardrails/pull/1

Pouyanpi (approved):
Thanks @MuneezaAzmat 🚀

Issue #1615, opened 2026-02-03
feature: Tool calling openAI compatible


lasschehj:
### Did you check the docs?

- [x] I have read all the NeMo-Guardrails docs

### Is your feature request related to a problem? Please describe.

The proxy supports OpenAI-compatible tool calls, which allows it to work with the OpenAI SDK. However, for guardrails, it does not accept tool calls in the same format. As a result, I’m forced to choose between using guardrails or tool calls—a trade-off I’d rather avoid.

### Describe the solution you'd like

Guardrails chat completions should accept OpenAI-compatible tool calls.

### Describe alternatives you've considered

This forces a choice between tool calls and guardrails, neither of which is an option for me.

### Additional context

Thank you!

Pouyanpi:
Thanks @lasschehj, this looks promising! Would you mind sharing an example of what you’d expect the final API to look like?

We’re currently working on enabling this, and your input could help

norika1207-lab:
This feature would be especially useful if guardrails can see tool calls as first-class structured events, not just text around the model response.

For agent systems, the risky part is often not the final assistant text. It is the transition from intent to action: tool name, arguments, authority, and whether the tool result should be trusted on the next turn.

If OpenAI-compatible tool calls are added, I would want guardrail hooks at a few boundaries:

- before tool execution: validate tool name, argument schema, policy scope
- after tool execution: validate result class and redaction rules
- before next model turn: decide whether the tool result may enter memory/context

That avoids forcing users to choose between tool use and governance. The guardrail layer becomes part of the action pipeline, not only a wrapper around natural-language output.

PR #1942, opened 2026-05-28
feat(server): Add tool calling params

christinaexyou:
## Description

Enables tool calls via `v1/chat/completions` endpoint. Currently, NeMo Guardrails is incompatible with applications using tool calls because there was missing API coverage. This PR addresses this bug by adding the following:
* Tool calling params in `OpenAIChatCompletionRequest`
* Normalizes internal/Langchain tool calls to OpenAI format, including streaming chunks

## Related Issue(s)

#1615 

## Testing 
* Updates `test_schema_utils.py` and `test_api.py` with unit tests

### Unit tests
```
 poetry run pytest -q
................................................ssss........... [  1%]
............................................................... [  2%]
.................s............................................. [  4%]
............................................................... [  5%]
............................................................... [  6%]
............................................................... [  8%]
............................................................... [  9%]
............................................................... [ 11%]
............................................................... [ 12%]
............................................................... [ 13%]
............................................................... [ 15%]
............................................................... [ 16%]
............................................................... [ 18%]
............................................................... [ 19%]
............................................................... [ 20%]
............................................................... [ 22%]
............................................................... [ 23%]
...............................s......ss...................ssss [ 24%]
sss............................................................ [ 26%]
............................................................... [ 27%]
....s.......s.................................................. [ 29%]
............................................................... [ 30%]
............................................................... [ 31%]
............................................................... [ 33%]
.......................................s....................... [ 34%]
............................................................... [ 36%]
............................................................... [ 37%]
............................................................... [ 38%]
............................................................... [ 40%]
........................................................sssssss [ 41%]
s......ssssss..ss.s.............ssssssss....................... [ 42%]
............................................................... [ 44%]
............................................................... [ 45%]
......................................ss....................... [ 47%]
........s...............s.......sssss.......................... [ 48%]
.................................s............................. [ 49%]
............................................................... [ 51%]
............................................................... [ 52%]
......ss........ss...ss........................................ [ 54%]
....s.....................................................s.... [ 55%]
........s...................................................... [ 56%]
............................................................... [ 58%]
............................................................... [ 59%]
............................................................... [ 60%]
............................................................... [ 62%]
...............sssss......ssssssssssssssssss..........sssss.... [ 63%]
............................................................... [ 65%]
.................s...........ss................................ [ 66%]
...................sssssssss.ssssssssss........................ [ 67%]
...............s............................................... [ 69%]
.....s....s.................................................... [ 70%]
....ssssssss.................sss...ss...ss.....sssssssssssss... [ 72%]
............................................................... [ 73%]
............................................................... [ 74%]
.......................s....................................... [ 76%]
............................................................... [ 77%]
........s....................ssssssss.........ss............... [ 78%]
............................................................... [ 80%]
........................................................sssssss [ 81%]
............................................................... [ 83%]
.s............................................................. [ 84%]
............................................................... [ 85%]
............................................................... [ 87%]
............................................................... [ 88%]
............................................................... [ 90%]
............................................................... [ 91%]
............................................................... [ 92%]
............................................................... [ 94%]
.s............................................................. [ 95%]
............................................................... [ 97%]
............................................................... [ 98%]
............................................................... [ 99%]
..........                                                      [100%]
4382 passed, 164 skipped in 127.60s (0:02:07)
```

### Pre-commit 
```
poetry run pre-commit run --all-files
ruff format..............................................................Passed
```
## Checklist

### DIFF
--- a/benchmark/mock_llm_server/models.py
+++ b/benchmark/mock_llm_server/models.py
@@ -39,6 +39,9 @@ class ChatCompletionRequest(BaseModel):
     presence_penalty: Optional[float] = Field(0.0, description="Presence penalty", ge=-2.0, le=2.0)
     frequency_penalty: Optional[float] = Field(0.0, description="Frequency penalty", ge=-2.0, le=2.0)
     logit_bias: Optional[dict[str, float]] = Field(None, description="Modify likelihood of specified tokens")
+    tools: Optional[list[dict]] = Field(None, description="Tools parameter.")
+    tool_choice: Optional[str | dict] = Field(None, description="Tool choice parameter.")
+    parallel_tool_calls: Optional[bool] = Field(None, description="Whether to allow parallel tool calls.")
     user: Optional[str] = Field(None, description="Unique identifier representing your end-user")
 
 
--- a/nemoguardrails/server/api.py
+++ b/nemoguardrails/server/api.py
@@ -533,6 +533,17 @@ async def chat_completion(body: GuardrailsChatCompletionRequest, request: Reques
             detail="thread_id message-history replay is not supported for Colang 2.0.",
         )
 
+    if (body.tools or body.tool_choice is not None or body.parallel_tool_calls is not None) and (
+        llm_rails.config.passthrough is not True or body.stream
+    ):
+        raise HTTPException(
+            status_code=422,
+            detail=(
+                "The 'tools', 'tool_choice', and 'parallel_tool_calls' parameters are only "
+                "supported for non-streaming requests when the guardrails configuration has 'passthrough: true'."
+            ),
+        )
+
     try:
         messages = body.messages or []
         if body.guardrails.context:
--- a/nemoguardrails/server/api.py
+++ b/nemoguardrails/server/api.py
@@ -28,8 +28,6 @@ from typing import Any, AsyncIterator, Callable, List, Optional, Union
 import httpx
 from fastapi import FastAPI, HTTPException, Request
 from fastapi.middleware.cors import CORSMiddleware
-from openai.types.chat.chat_completion import Choice
-from openai.types.chat.chat_completion_message import ChatCompletionMessage
 from pydantic import BaseModel, ValidationError
 from starlette.responses import RedirectResponse, StreamingResponse
 
@@ -43,11 +41,15 @@ from nemoguardrails.server.schemas.openai import (
     OpenAIModelsList,
 )
 from nemoguardrails.server.schemas.utils import (
+    bot_message_to_chat_completion,
     create_error_chat_completion,
     extract_bot_message_from_response,
     fetch_models,
     format_streaming_chunk_as_sse,
     generation_response_to_chat_completion,
+    normalize_tool_calls_openai,
+    resolve_tool_calls,
+    warn_if_thread_history_invalid_for_tool_use,
 )
 
 try:
@@ -554,6 +556,7 @@ async def chat_completion(body: GuardrailsChatCompletionRequest, request: Reques
             # the string `thread-` to all thread keys.
             datastore_key = "thread-" + body.guardrails.thread_id
             thread_messages = json.loads(await datastore.get(datastore_key) or "[]")
+            warn_if_thread_history_invalid_for_tool_use(thread_messages)
 
             # And prepend them.
             messages = thread_messages + messages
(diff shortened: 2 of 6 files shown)


Pouyanpi:
Nice work @christinaexyou, the OpenAI shape looks mostly right here 👍 But I think there are a few things we should sort out before merging, otherwise it can break normal OpenAI clients:

- openai import vs pin: `utils.py` imports `chat_completion_message_function_tool_call`, but it only exists in openai ~1.99+. our pin is `>=1.0.0`, so it fails on older versions (I checked, breaks on 1.61.1). we should either bump the floor or use the stable `ChatCompletionMessageToolCall` import.
- streaming: we don't send `finish_reason: "tool_calls"` on the final chunk, and streamed turns are not saved to the thread at all. I will open a PR to introduce `GenerationChunkResponse` or something similar that we can use here.
- thread history: when we persist the assistant message, we drop `GenerationResponse.tool_calls` (only `response.response[0]` is stored). So on the next turn the history has a `tool` message without the matching `tool_calls` before it, which is invalid for OpenAI. I'm not saying we should support it but we should watch out for it.

The response builder already handles the tool_calls merge, so I think we can reuse the same helper for the persistence path if we plan to add support. Happy to help if useful 🙂

christinaexyou:
thanks for the feedback @Pouyanpi ! i made the following updates:

* replaced `chat_completion_message_function_tool_call` with  `ChatCompletionMessageToolCall` for backward OpenAI compatability
* removed streaming tool calls - I can wait for your PR to be merged and then open up a separate PR to add this back in
* i added support for threaded multi-turn tool use by re-using `resolve_tool_calls` in `api.py` to merge `bot_message.tool_calls` with `GenerationResponse.tool_calls` - thanks for the hint ! i also added helper functions in `utils.py` to validate OpenAI tool-calling message ordering (`warn_if_thread_history_invalid_for_tool_use`) and a fallback if the tool call id is missing (`_generate_fallback_tool_call_id`)

RobGeada (approved):
Tested a running server built from this PR:

```python
response = requests.post(
    "http://localhost:9999/v1/chat/completions",
    json={
      "model": "qwen36-fp8",
      "messages": [
        {"role": "user", "content": "What is the weather in London?"}
      ],
      "tools": [
        {
          "type": "function",
          "function": {
            "name": "get_weather",
            "description": "Get the current weather for a location",
            "parameters": {
              "type": "object",
              "properties": {
                "location": {
                  "type": "string",
                  "description": "The city name"
                }
              },
              "required": ["location"]
            }
          }
        }
      ],
      "tool_choice": "auto"
    }
)
result = response.json()
print(json.dumps(result, indent=2))
```

Returns:
```json
{
  "id": "chatcmpl-8403868f-7e54-4649-bdf3-2e144ad528e5",
  "choices": [
    {
      "finish_reason": "tool_calls",
      "index": 0,
      "message": {
        "role": "assistant",
        "tool_calls": [
          {
            "id": "chatcmpl-tool-8f1ad866af35673e",
            "function": {
              "arguments": "{\"location\": \"London\"}",
              "name": "get_weather"
            },
            "type": "function"
          }
        ]
      }
    }
  ],
  "created": 1781516029,
  "model": "qwen36-fp8",
  "object": "chat.completion",
  "guardrails": {
    "config_id": "my-config"
  }
}
```

✅ tool calls are sent and returned as expected!

Pouyanpi (changes requested):
Thanks @christinaexyou , let's see how we can resolve following issue.

IMO this needs a guard for non-passthrough configs

`tools`, `tool_choice`, and `parallel_tool_calls` are being added to `GenerationOptions.llm_params`, which applies to calls to the main LLM inside the rails orchestration .

reproduced the same forced tool-call request works with `passthrough: true`, but with `passthrough: false` it returns:

```json
{
  "choices": [{
    "finish_reason": "stop",
    "message": {
      "role": "assistant",
      "content": ""
    }
  }]
}
```
no tool_calls are returned, even though the request forced:
```json
"tool_choice": {
  "type": "function",
  "function": {"name": "get_weather"}
}
```
with passthrough: true, the response correctly has finish_reason: "tool_calls" and message.tool_calls.

so I don’t think we should expose OpenAI tool params for non-passthrough configs yet. 

minimal fix: reject requests with tools,tool_choice, parallel_tool_calls unless the selected config is passthrough. Is this ideal?

just note that, passthrough is not fully ok either. with stream: true, even bare passthrough emits empty content chunks and no delta.tool_calls as mentioned before. 
Output rails streaming variants return an internal server error etc.


the working scope is:
normal chat completion passthrough:True (stream: False)

Pouyanpi (approved):
Thank you @christinaexyou for making the changes, LGTM! 🚀 

Look forward to the follow-up issues 👍🏻

PR #1947, opened 2026-05-31
fix(colang): guard ' or' line continuation at end of file

vidigoat:
## Summary

In `get_numbered_lines` (`nemoguardrails/colang/v1_0/lang/utils.py`), the line-continuation loop condition is:

```python
while i < len(raw_lines) - 1 and text[-1] == "\\" or text.endswith(" or"):
```

Because `and` binds tighter than `or`, this evaluates as:

```python
(i < len(raw_lines) - 1 and text[-1] == "\\") or text.endswith(" or")
```

So the bounds check `i < len(raw_lines) - 1` only guards the backslash continuation, **not** the `" or"` continuation. A line ending in `" or"` as the **last line of the file** enters the loop and runs `raw_lines[i]` with `i` past the end, raising `IndexError` instead of parsing cleanly.

Reproducer:
```python
from nemoguardrails.colang.v1_0.lang.parser import parse_colang_file
parse_colang_file("test.co", "define flow test\n  user express greeting or")
# IndexError: list index out of range
```

The backslash sibling is correctly guarded (a `\` at EOF does not crash), which confirms the asymmetry is a bug, not intended.

## Fix

Parenthesize the two continuation conditions so the bounds check guards both:

```python
while i < len(raw_lines) - 1 and (text[-1] == "\\" or text.endswith(" or")):
```

## Tests

Added two regression tests in `tests/test_parser_utils.py`: a `" or"` line at EOF no longer raises (kept as-is), and a valid `" or"` continuation still merges onto the next line.

Verified locally:
```
pytest tests/test_parser_utils.py tests/colang/parser/test_basic.py   # 6 passed
ruff check                                                            # clean
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a parser issue with line continuation handling in CoLang that could cause errors when continuations appear at the end of files.

* **Tests**
  * Added test cases to validate proper handling of line continuations in edge cases.

### DIFF
--- a/nemoguardrails/colang/v1_0/lang/utils.py
+++ b/nemoguardrails/colang/v1_0/lang/utils.py
@@ -157,7 +157,7 @@ def get_numbered_lines(content: str):
         # but without the indentation.
         # Also, if there's an active "operator" like "or", we also continue to the next line
         text = raw_line
-        while i < len(raw_lines) - 1 and text[-1] == "\\" or text.endswith(" or"):
+        while i < len(raw_lines) - 1 and (text[-1] == "\\" or text.endswith(" or")):
             i += 1
             if text[-1] == "\\":
                 text = text[0:-1]
--- a/tests/test_parser_utils.py
+++ b/tests/test_parser_utils.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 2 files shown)


Pouyanpi (approved):
LGTM! Thank you @vidigoat for the fix 👍🏻

PR #1950, opened 2026-06-01
fix(actions): ip_filter validator crashes on dict responses

vidigoat:
## Problem

`@validate_response(validators=["ip_filter"])` crashes with `ValueError: too many values to unpack (expected 2)` whenever the decorated action returns a `dict`.

`nemoguardrails/actions/validation/base.py`:

```python
elif isinstance(response_value, dict):
    for key, value in response_value:          # iterates KEYS, not items
        response_value[key] = filter_ip(value)
```

Iterating a dict yields its keys (strings). Unpacking each key into `(key, value)` raises `ValueError` for any key longer than two characters. This validator is applied to the LangChain safetool actions in `nemoguardrails/integrations/langchain/actions/safetools.py`, so any of them returning a dict response hits the crash on otherwise-valid input. The `dict` branch had no test coverage, so the bug shipped undetected.

## Fix

```python
elif isinstance(response_value, dict):
    for key, value in response_value.items():
        response_value[key] = filter_ip(value)
```

## Verification

```
before:  pytest tests/test_actions_validation.py::test_ip_filter_on_dict_response
         ValueError: too many values to unpack (expected 2)   (base.py:102)
after :  3 passed   (full tests/test_actions_validation.py, incl. new regression test)
```

Added `test_ip_filter_on_dict_response` covering a dict response containing an IP. `ruff check` and `ruff format --check` pass on both changed files (matches the repo's pre-commit config, ruff v0.14.6).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed IP address filtering to correctly identify and remove sensitive IP addresses from dictionary-structured responses, ensuring consistent protection across all response types.

### DIFF
--- a/nemoguardrails/actions/validation/base.py
+++ b/nemoguardrails/actions/validation/base.py
@@ -99,7 +99,7 @@ def validate_response(validators: List[str] = [], **validation_args):
                 if isinstance(response_value, str):
                     response_value = filter_ip(response_value)
                 elif isinstance(response_value, dict):
-                    for key, value in response_value:
+                    for key, value in response_value.items():
                         response_value[key] = filter_ip(value)
 
             if "is_default_resp" in validators:
--- a/nemoguardrails/actions/validation/base.py
+++ b/nemoguardrails/actions/validation/base.py
@@ -100,7 +100,8 @@ def validate_response(validators: List[str] = [], **validation_args):
                     response_value = filter_ip(response_value)
                 elif isinstance(response_value, dict):
                     for key, value in response_value.items():
-                        response_value[key] = filter_ip(value)
+                        if isinstance(value, str):
+                            response_value[key] = filter_ip(value)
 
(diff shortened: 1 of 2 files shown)


vidigoat:
Thanks @greptile-apps and @coderabbitai — good catch. `filter_ip` runs `re.sub`, which would raise `TypeError` on non-string dict values (int/bool/nested dict). Latest commit guards with `isinstance(value, str)` so only string values are filtered, and adds a mixed-value-dict regression test.

vidigoat:
Thanks @Pouyanpi! Per the merge-guidance bot, I've signed both commits and force-pushed — they now show as verified on GitHub, and the signing requirement is satisfied. The force-push dismissed your earlier approval, but the diff is unchanged (commit-signing only, no content changes). Could you re-approve when you get a moment? 🙏

PR #1954, opened 2026-06-01
test(parallel_rails): replace wall-clock assertion with structural overlap check — closes #1953

### LINKED ISSUE #1953: test(parallel_rails): timing assertion is flaky on slow CI runners

Reproduced on the CI matrix's Python 3.10 / Ubuntu runner.

### Operating system/version

GitHub Actions `ubuntu-latest` (the failing job specifically; 3.11 / 3.12 / 3.13 all pass on the same runner pool).

### NeMo-Guardrails version (if you must use a specific version and not the latest)

`develop` (HEAD as of 2026-06-01).

### Describe the bug

`tests/test_parallel_rails.py::test_parallel_rails_success` asserts that the input-rails and output-rails wall-clock duration is `< 1.5s`:

https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/tests/test_parallel_rails.py#L74-L78

```python
# Time should be close to 2 seconds due to parallel processing:
# check blocked input terms: 1s
# check blocked output terms: 1s
assert result.log.stats.input_rails_duration < 1.5 and result.log.stats.output_rails_duration < 1.5, (
    "Rails processing took too long, parallelization seems to be not working."
)
```

The threshold sits between the parallel-execution lower bound (~1.0s, two `asyncio.sleep(1.0)` in flight) and the serial-execution upper bound (~2.0s). On a slow / loaded CI runner, event-loop scheduling overhead alone can push the parallel duration above 1.5s, failing the test even though parallelism is working correctly.

This is a **false-positive flake**, not a real regression — when the failure happens the rails really did run in parallel, the wall clock just didn't co-operate.

### Steps To Reproduce

It reproduced on a recent unrelated PR (#1943, commit `5c423d5`) — `pr-tests-matrix (Ubuntu, 3.10)` step:

```
FAILED tests/test_parallel_rails.py::test_parallel_rails_success
- AssertionError: Rails processing took too long, parallelization seems to be not working.
```

[CI log](https://github.com/NVIDIA-NeMo/Guardrails/actions/runs/26667249504). The other Python versions in the same matrix passed on the same commit, which is consistent with a runner-speed flake rather than a code regression.

### Expected Behavior

The test should pass whenever the parallel-rails machinery actually runs the two `check blocked input terms` rails concurrently, regardless of how fast the CI runner happens to be.

### Actual Behavior

The test passes deterministically on a developer workstation and on faster CI runners, but flakes intermittently on slower ones (observed on Ubuntu / Python 3.10 in #1943's matrix run).

### Proposed fix direction

Replace the wall-clock threshold with a **structural overlap check** on `ActivatedRail.started_at` / `finished_at`. Two rails are running in parallel iff the second rail starts before the first one finishes:

```python
# Pull the two parallel "check blocked input terms" rails (indices 1 + 2 per the assertion above)
parallel_input_rails = [
    rail for rail in result.log.activated_rails
    if "check blocked input terms" in rail.name
]
assert len(parallel_input_rails) == 2
r1, r2 = sorted(parallel_input_rails, key=lambda r: r.started_at)
assert r2.started_at < r1.finished_at, (
    f"input rails did not run in parallel: "
    f"rail 1 [{r1.started_at:.3f}, {r1.finished_at:.3f}], "
    f"rail 2 started at {r2.started_at:.3f}"
)
# Same shape for the two output rails.
```

This proves parallelism without depending on absolute wall-clock duration. As long as the second rail starts before the first ends, the assertion holds — even if the entire bucket takes longer than 1.5s due to CI load.

`ActivatedRail` already exposes `started_at` / `finished_at` / `duration` (`nemoguardrails/rails/llm/options.py`), so the data is there.

Happy to send the PR if this direction is agreeable. The same wall-clock pattern is only in this one assertion in `test_parallel_rails.py`; `test_parallel_rails_exceptions.py` uses behaviour-based checks and isn't affected.

ppcvote:
## Summary

Closes #1953.

`tests/test_parallel_rails.py::test_parallel_rails_success` previously asserted that the parallel-rails bucket completed in `< 1.5s` wall-clock:

```python
# Time should be close to 2 seconds due to parallel processing:
# check blocked input terms: 1s
# check blocked output terms: 1s
assert result.log.stats.input_rails_duration < 1.5 and result.log.stats.output_rails_duration < 1.5, (
    \"Rails processing took too long, parallelization seems to be not working.\"
)
```

The 1.5s threshold was chosen to split the parallel lower bound (~1.0s, two `asyncio.sleep(1.0)` in flight) from the serial upper bound (~2.0s). On a slow CI runner, event-loop scheduling overhead can push parallel execution above 1.5s, producing a false-positive failure even when parallelism is working correctly. Observed on Ubuntu / Python 3.10 in [#1943's matrix run](https://github.com/NVIDIA-NeMo/Guardrails/actions/runs/26667249504).

## Fix

Replace the wall-clock check with a structural overlap check on `ActivatedRail.started_at` / `finished_at`. Two paired rails are running in parallel iff the second rail starts before the first one finishes — true regardless of how slow the runner is.

```python
for rail_type in (\"input\", \"output\"):
    paired_rails = [
        rail
        for rail in result.log.activated_rails
        if f\"check blocked {rail_type} terms\" in rail.name
    ]
    assert len(paired_rails) == 2
    first, second = sorted(paired_rails, key=lambda r: r.started_at or 0)
    assert second.started_at < first.finished_at, ...
```

`ActivatedRail` already exposes `started_at` / `finished_at` / `duration` (`nemoguardrails/rails/llm/options.py`), so no schema changes are needed.

## What's preserved unchanged

- All the prior index-based name assertions on `activated_rails[0..6]`.
- The response-content assertion at the top.
- Every other test in the file — they're behaviour-based (blocked-term detection, stop flag, etc.) and aren't affected by this flake.

## Verification

- `python -m py_compile tests/test_parallel_rails.py` passes locally.
- I haven't run the full pytest matrix from this fork (no LLM credentials configured here), but the logic is straightforward: if parallel runs, sibling rails overlap → assertion holds; if serial regression sneaks in, sibling rails will not overlap → assertion fails (with a clearer error message that includes the actual start/end timestamps).

## DCO

Signed-off-by line is on the commit.

## Notes

- `tests/test_parallel_rails_exceptions.py` uses behaviour-based checks (no wall-clock asserts), so it isn't affected.
- The same `started_at` / `finished_at` structural-overlap pattern could be useful as a helper if more parallel-rail tests get added; happy to factor it into `tests/utils` in a follow-up if you'd like.

## Summary by CodeRabbit

* **Tests**
  * Improved parallel execution test reliability by switching from timing-based assertions to structural validation, reducing CI flakiness and ensuring consistent test results across environments.

### DIFF
--- a/tests/test_parallel_rails.py
+++ b/tests/test_parallel_rails.py
@@ -74,22 +74,15 @@ async def test_parallel_rails_success():
     # parallel bucket above 1.5s even when parallelism worked correctly.
     # See issue #1953.
     for rail_type in ("input", "output"):
-        paired_rails = [
-            rail
-            for rail in result.log.activated_rails
-            if f"check blocked {rail_type} terms" in rail.name
-        ]
+        paired_rails = [rail for rail in result.log.activated_rails if f"check blocked {rail_type} terms" in rail.name]
         assert len(paired_rails) == 2, (
-            f"expected 2 paired `check blocked {rail_type} terms` rails, "
-            f"got {len(paired_rails)}"
+            f"expected 2 paired `check blocked {rail_type} terms` rails, got {len(paired_rails)}"
         )
         first, second = sorted(paired_rails, key=lambda r: r.started_at or 0)
         assert first.started_at is not None and first.finished_at is not None, (
             f"{rail_type} rail timing fields not populated: {first}"
         )
-        assert second.started_at is not None, (
-            f"{rail_type} rail timing fields not populated: {second}"
-        )
+        assert second.started_at is not None, f"{rail_type} rail timing fields not populated: {second}"
         assert second.started_at < first.finished_at, (
             f"{rail_type} rails did not run in parallel: "
             f"first  [{first.started_at:.3f}, {first.finished_at:.3f}], "
--- a/tests/test_parallel_rails.py
+++ b/tests/test_parallel_rails.py
@@ -65,12 +65,36 @@ async def test_parallel_rails_success():
     assert result.log.activated_rails[5].name == "check blocked output terms $duration=1.0"
     assert result.log.activated_rails[6].name == "check blocked output terms $duration=1.0"
 
-    # Time should be close to 2 seconds due to parallel processing:
-    # check blocked input terms: 1s
-    # check blocked output terms: 1s
-    assert result.log.stats.input_rails_duration < 1.5 and result.log.stats.output_rails_duration < 1.5, (
-        "Rails processing took too long, parallelization seems to be not working."
-    )
+    # Verify parallelism structurally instead of via a wall-clock threshold.
+    # Each paired `check blocked input/output terms $duration=1.0` rail sleeps
+    # for ~1s; if the two siblings run in parallel, the second one starts
+    # before the first one finishes. This is true regardless of how slow the
+    # CI runner is — the previous `duration < 1.5s` assertion was flaky on
+    # slow runners because event-loop scheduling overhead could push the
+    # parallel bucket above 1.5s even when parallelism worked correctly.
+    # See issue #1953.
+    for rail_type in ("input", "output"):
+        paired_rails = [
+            rail
+            for rail in result.log.activated_rails
+            if f"check blocked {rail_type} terms" in rail.name
+        ]
+        assert len(paired_rails) == 2, (
+            f"expected 2 paired `check blocked {rail_type} terms` rails, "
+            f"got {len(paired_rails)}"
+        )
+        first, second = sorted(paired_rails, key=lambda r: r.started_at or 0)
+        assert first.started_at is not None and first.finished_at is not None, (
+            f"{rail_type} rail timing fields not populated: {first}"
+        )
+        assert second.started_at is not None, (
+            f"{rail_type} rail timing fields not populated: {second}"
+        )
+        assert second.started_at < first.finished_at, (
+            f"{rail_type} rails did not run in parallel: "
+            f"first  [{first.started_at:.3f}, {first.finished_at:.3f}], "
+            f"second started at {second.started_at:.3f}"
+        )
 
 
 @pytest.mark.asyncio


Pouyanpi:
Thanks for the PR @ppcvote. Could you please GPG-sign the commit per CONTRIBUTING.md? I’ll continue the code review after that.

ppcvote:
Thanks @Pouyanpi — both commits are now GPG-signed (verified by GitHub). Ready for your code review.

ppcvote:
@Pouyanpi — gentle ping. Both commits are GPG-signed as of June 2; happy to address any review feedback when you have a chance. No urgency — just keeping the PR alive in case it slipped off the list.

ppcvote:
Hi @Pouyanpi — gentle follow-up on this one. It's a test-only change (single file) that fixes the flaky wall-clock assertion in #1953; CI matrix (3.10–3.13) and codecov are green, no conflicts, and commits are GPG-signed as you requested. Happy to add a triage label or any info that would help move this through your triage queue. Thanks!

tests/test_parallel_rails.py
```diff
@@ -65,12 +65,29 @@ async def test_parallel_rails_success():
...
+        first, second = sorted(paired_rails, key=lambda r: r.started_at or 0)
+        assert first.started_at is not None and first.finished_at is not None, (
+            f"{rail_type} rail timing fields not populated: {first}"
+        )
+        assert second.started_at is not None, f"{rail_type} rail timing fields not populated: {second}"
+        assert second.started_at < first.finished_at, (
+            f"{rail_type} rails did not run in parallel: "
+            f"first  [{first.started_at:.3f}, {first.finished_at:.3f}], "
+            f"second started at {second.started_at:.3f}"
+        )
```
Pouyanpi, line 90: this assertion is vacuous: it passes even when rails run sequentially, so it does not actually verify parallelism (the regression it claims to guard against)

tests/test_parallel_rails.py
```diff
@@ -65,12 +65,29 @@ async def test_parallel_rails_success():
...
-        "Rails processing took too long, parallelization seems to be not working."
-    )
+    # Verify parallelism structurally instead of via a wall-clock threshold.
+    # Each paired `check blocked input/output terms $duration=1.0` rail sleeps
+    # for ~1s; if the two siblings run in parallel, the second one starts
+    # before the first one finishes. This is true regardless of how slow the
+    # CI runner is — the previous `duration < 1.5s` assertion was flaky on
+    # slow runners because event-loop scheduling overhead could push the
+    # parallel bucket above 1.5s even when parallelism worked correctly.
+    # See issue #1953.
```
Pouyanpi, line 75: once fixed we don't need this comment, please remove.

Pouyanpi (changes requested):
@ppcvote this doesn't seem to test the expected behavior. the new structural assertion is vacuous. Please see my comments below.

Pouyanpi:
closed in favor of #2070

PR #1955, opened 2026-06-01
fix(embeddings): recover missing FastEmbed ONNX cache

Pouyanpi:
## Description

Retry FastEmbed initialization with an explicit local cache when ONNXRuntime reports a missing model.onnx from a stale cache.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced robustness of embedding model initialization with automatic recovery from missing model cache scenarios.

### DIFF
--- a/nemoguardrails/embeddings/providers/fastembed.py
+++ b/nemoguardrails/embeddings/providers/fastembed.py
@@ -19,6 +19,13 @@ from typing import List
 from .base import EmbeddingModel
 
 
+def _is_missing_onnx_model_error(ex: Exception) -> bool:
+    message = str(ex)
+    return "model.onnx" in message and (
+        "Could not find model.onnx in" in message or "NO_SUCHFILE" in message or "File doesn't exist" in message
+    )
+
+
 def get_executor():
     from . import embeddings_executor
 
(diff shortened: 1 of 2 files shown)


nemoguardrails/embeddings/providers/fastembed.py
```diff
@@ -51,14 +58,16 @@ def __init__(self, embedding_model: str, **kwargs):
...
-        except ValueError as ex:
+        except Exception as ex:
             # Sometimes the cached model in the temporary folder gets removed,
             # but the folder still exists, which causes an error. In this case,
             # we fall back to an explicit cache directory.
-            if "Could not find model.onnx in" in str(ex):
-                self.model = Embedding(embedding_model, cache_dir=".cache", **kwargs)
+            if _is_missing_onnx_model_error(ex):
+                fallback_kwargs = dict(kwargs)
+                fallback_kwargs["cache_dir"] = ".cache"
```
christinaexyou, line 67: can we make this an absolute path eg:
```
fallback_kwargs["cache_dir"] = str(Path.home() / ".cache" / "fastembed")
```

my concern is that is we use a relative path, it will fail in docker containers since cwd is usually / or read only

christinaexyou: or alternatively, log a warning when overriding the value

PR #1967, opened 2026-06-01
Make mask_sensitive_data honor configured score_threshold

Kymi808:
## Summary

`mask_sensitive_data` (`nemoguardrails/library/sensitive_data_detection/actions.py`) called `_get_analyzer()` with no arguments, so the analyzer was built (and `@lru_cache`-d) at the function's default `0.4` regardless of what the user configured in `sensitive_data_detection.<source>.score_threshold` (which itself defaults to `0.2` in `SensitiveDataDetectionOptions`).

`_get_analyzer` is the only place `default_score_threshold` is set on the underlying Presidio `AnalyzerEngine`, so values between the configured threshold and `0.4` were still masked even though `detect_sensitive_data` (which already passes the configured threshold through, see line 121) reported them as non-sensitive. Worse, because `_get_analyzer` is `@lru_cache`-d, the masking analyzer stays at the wrong threshold for the lifetime of the process — even after the user later adjusts the configuration.

```diff
-    analyzer = _get_analyzer()
+    default_score_threshold = getattr(options, "score_threshold")
+    ...
+    analyzer = _get_analyzer(score_threshold=default_score_threshold)
```

## Fix

Mirror `detect_sensitive_data`: read `default_score_threshold` from the options and pass it through to `_get_analyzer`. The behavior of `detect_sensitive_data` is unchanged.

## Test plan

Added a mock-based unit test in `tests/test_sensitive_data_detection_unit.py` that monkey-patches `_get_analyzer` (plus the optional Presidio anonymizer surface), invokes `mask_sensitive_data` with `score_threshold: 0.85`, and asserts `_get_analyzer` was called with `score_threshold=0.85`. Test lives in a separate file so it isn't caught by the existing module-level skip when the optional Presidio + spaCy stack isn't installed.

- **WITH this change:** `pytest tests/test_sensitive_data_detection_unit.py` → 1 passed.
- **On `main`:** the test FAILS — `_get_analyzer` is called with the default `0.4` instead of the configured `0.85`.
- `ruff check` and `ruff format --check` clean on both files (using the repo's pinned `ruff==0.14.6`).

## Notes

- All existing `score_threshold` tests in `test_sensitive_data_detection.py` use the `detect sensitive data` flow, so none pin the masking path's current (buggy) threshold behavior.
- Commit is DCO sign-off-signed per `CONTRIBUTING.md`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive data masking now properly respects user-configured score threshold settings instead of defaulting to internal values, ensuring behavior aligns with user expectations and configuration.

* **Tests**
  * Added unit tests to verify that configured score thresholds are correctly propagated and applied during sensitive data masking operations.


tests/test_sensitive_data_detection_unit.py
```diff
@@ -0,0 +1,92 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
christinaexyou: can you consolidate all your tests in `tests/tests/test_sensitive_data_detection.py` ? 

i think having a test file solely for `mask_sensitive_data` adds bloat to the codebase and can lead to potential confusion. i suggest after updating `tests/tests/test_sensitive_data_detection.py`, you test your unit tests locally with SDD_SETUP_PRESENT dependencies installed

nemoguardrails/library/sensitive_data_detection/actions.py
```diff
@@ -149,12 +149,17 @@ async def mask_sensitive_data(source: str, text: str, config: RailsConfig):
...
     assert source in ["input", "output", "retrieval"]
     options: SensitiveDataDetectionOptions = getattr(sdd_config, source)
+    default_score_threshold = getattr(options, "score_threshold")
 
     # If we don't have any entities specified, we stop
     if len(options.entities) == 0:
         return text
 
-    analyzer = _get_analyzer()
+    # Honor the configured score_threshold, mirroring detect_sensitive_data;
```
christinaexyou: i don't think that these comments are necessary esp. since they point out an issue that won't exist anymore once your code is merged. it's clear from the code that you are passing in a `score_threshold` to the function

Kymi808:
Thanks for the review @christinaexyou — both points addressed in `c2b68bb`:

- `test_mask_sensitive_data_honors_configured_score_threshold` is now in `tests/test_sensitive_data_detection.py` with the same `skipif(not SDD_SETUP_PRESENT) + unit + asyncio` decorator stack the other tests use, and `tests/test_sensitive_data_detection_unit.py` is deleted.
- Dropped the four-line explanatory comment above the fixed `_get_analyzer` call.

One small extra: while verifying locally as you suggested, every test in `test_sensitive_data_detection.py` was getting silently skipped with `Unexpected error during setup: name 'spacy' is not defined` — `setup_module` calls `spacy.util.is_package(...)` but `spacy` is never imported and the bare `except Exception` was swallowing the `NameError`. Added a one-line `if SDD_SETUP_PRESENT: import spacy` shim above `setup_module` so the file actually runs when the SDD extras are installed. Happy to split that out into a separate PR if you'd prefer.

Verified with `pip install presidio-analyzer presidio-anonymizer spacy && python -m spacy download en_core_web_lg`: the regression test **PASSES** with the fix and **FAILS** (`AssertionError: assert 0.4 == 0.85`) on a manually reverted `actions.py`.

Kymi808:
Quick note on the codecov 0% patch report: it reflects the existing CI tier pattern, not a regression introduced by this PR.

All 8 tests in `test_sensitive_data_detection.py` (the 7 pre-existing ones plus the consolidated regression test) are skipped in `pr-tests-matrix` because `presidio_analyzer` / `presidio_anonymizer` / `spacy` aren't installed in that tier, so `SDD_SETUP_PRESENT` is `False` and `setup_module` calls `pytest.skip`. From the run-26788987717 log:

```
test_masking_input_output                                     SKIPPED
test_detection_input_output                                   SKIPPED
test_masking_retrieval                                        SKIPPED
test_score_threshold                                          SKIPPED
test_invalid_score_threshold                                  SKIPPED
test_invalid_score_threshold_chat_message                     SKIPPED
test_high_score_threshold_disables_rails                      SKIPPED
test_mask_sensitive_data_honors_configured_score_threshold    SKIPPED  ← this PR
```

The `codecov/patch` status check is **passing** — the bot comment is informational. Per your suggestion I verified locally with the SDD extras installed (`presidio-analyzer`, `presidio-anonymizer`, `spacy`, `en_core_web_lg`): the regression test **passes** with the fix and **fails** (`AssertionError: assert 0.4 == 0.85`) on a manually reverted `actions.py`.

Happy to split out the `setup_module` cleanup or refactor it into per-test `skipif` markers (so the tests actually run in the default CI tier and codecov can measure them) as a separate PR if that would be useful — kept it out of this one to stay in scope.

tests/test_sensitive_data_detection.py
christinaexyou, line 44: can you add "import spacy" here and remove lines 36-38 ? otherwise, we check whether if SDD_SETUP_PRESENT twice

tests/test_sensitive_data_detection.py
```diff
@@ -420,3 +423,68 @@ def test_high_score_threshold_disables_rails():
     # This will trigger the input rail
     chat >> "Hi! I am Mr. John!"
     chat << "Hi! My name is John as well."
+
+
+@pytest.mark.skipif(not SDD_SETUP_PRESENT, reason="Sensitive Data Detection setup is not present.")
```
christinaexyou, line 427: can you refactor your test to match the formatting of the existing ones ? i would use `test_high_score_threshold_disables_rails()` as a reference and create an example config where the score_threshold is 1.0. we know that a score_threshold of 1.0 would NOT mask anything so if the user message with a PERSON entity is unaltered then we know that mask_sensitive_data honors a configured score threshold

tests/test_sensitive_data_detection.py
```diff
@@ -420,3 +423,68 @@ def test_high_score_threshold_disables_rails():
...
+@pytest.mark.unit
+@pytest.mark.asyncio
+async def test_mask_sensitive_data_honors_configured_score_threshold(monkeypatch):
+    """Regression: ``mask_sensitive_data`` must honor ``options.score_threshold``.
```
christinaexyou: also, no need for this comment per my previous feedback

Kymi808:
Thanks for the review @christinaexyou — addressed in the latest commits:

- Rewrote `test_mask_sensitive_data_honors_configured_score_threshold` in the config-based mask-flow style, mirroring `test_high_score_threshold_disables_rails` (`score_threshold: 1.0`, `mask sensitive data on input` flow, with a `check_user_message` action asserting the user message reaches the inner pipeline unaltered — proving the configured threshold is honored end-to-end).
- Moved `import spacy` into `setup_module` after the SDD skip check and removed the duplicate top-level conditional import.
- Dropped the explanatory docstring on the regression test per your earlier feedback.

Also rebased the branch onto current `develop`.

PR #1970, opened 2026-06-02
test: make xdist the default Makefile test path

Pouyanpi:
## Summary

Make the default Makefile test path run the suite in parallel with pytest-xdist, while keeping an explicit serial pytest escape hatch.

## What Changed

- `make test` now runs non-live pytest with xdist.
- `make test-parallel` remains available as an alias for `make test`.
- `make test-serial` runs plain pytest without xdist or live-env filtering.
- Live provider keys and live-mode flags are cleared for the non-live parallel path.
- Default FastEmbed usage in tests is replaced with deterministic embeddings unless a test explicitly opts into real embeddings.
- Shared test state is isolated so the suite behaves correctly under xdist.
- Real language-detection checks are gated so they do not run in normal unit-test mode.

## Why

The recommended local and agent test path should be fast, deterministic, and safe to run without live provider credentials. This makes `make test` the project-level test command, while preserving `make test-serial` for debugging raw pytest behavior.

## Impact

Local full-suite timing:

- Serial pytest path: 138.24s
- Parallel Makefile path: 29s

This makes the recommended local and agent test command about 4.8x faster on the measured machine.

The same change also reduces CI wall time by switching the reusable workflow from raw `poetry run pytest` to Makefile targets:

- non-coverage jobs run `make test`
- the coverage job runs `make test-coverage`

The current CI matrix runs PR tests across 4 Ubuntu/Python jobs, full tests across 8 Windows/macOS Python jobs, and latest-deps tests across 12 OS/Python jobs.

CI timing comparison:

- Baseline raw pytest run: https://github.com/Pouyanpi/NeMo-Guardrails/actions/runs/26810039479
- Current Makefile xdist run: https://github.com/Pouyanpi/NeMo-Guardrails/actions/runs/26812554221

| Job | Baseline job | Current job | Baseline test step | Current test step |
| --- | ---: | ---: | ---: | ---: |
| Ubuntu / Python 3.10 non-coverage | 4m55s | 2m18s | 4m01s | 1m49s |
| Ubuntu / Python 3.11 coverage | 7m43s | 3m20s | 6m55s | 2m44s |
| Ubuntu / Python 3.12 non-coverage | 3m33s | 1m57s | 2m39s | 1m31s |
| Ubuntu / Python 3.13 non-coverage | 4m09s | 2m01s | 3m09s | 1m31s |

Note: the measured coverage run completed pytest successfully and failed later during Codecov upload.

## Validation

```bash
make help
```

```bash
env LIVE_TEST=1 OPENAI_API_KEY=sk-test make test TEST=tests/test_embeddings_openai.py WORKERS=1 ARGS='-q --tb=short'
```

```bash
make test TEST=tests/test_utils.py WORKERS=1 ARGS='-q'
```

## Release Notes

* **Tests**
  * Added support for parallel test execution with configurable workers
  * Introduced pytest markers (serial, slow, live, real_embeddings) for organizing and filtering tests
  * Enhanced test determinism with deterministic embedding provider
  * Improved test isolation with automated state cleanup between tests

* **Chores**
  * Refactored test targets and GitHub Actions workflows for consistency


tests/conftest.py
```diff
@@ -47,3 +109,26 @@ def langchain_framework():
...
+
+
+def _uses_default_fastembed(provider_config, default_engine, default_model):
+    if provider_config.name != "default":
+        return False
+
+    parameters = provider_config.parameters
+    engine = parameters.get("embedding_engine", default_engine)
+    model = parameters.get("embedding_model", default_model)
+    return engine in _FASTEMBED_ENGINES and model in _FASTEMBED_MODELS
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai addressed in https://github.com/NVIDIA-NeMo/Guardrails/pull/1970/commits/34c018ad31017d64bb6f603aa265472c744841a8, do you confirm?

tests/testing/embeddings.py
```diff
@@ -0,0 +1,115 @@
...
+
+        for index, item in enumerate(self.items):
+            score = _similarity(text, item.text, text_embedding, self.embeddings[index])
+            if threshold != float("inf") and not _exact_match(text, item.text):
+                score = min(score, threshold - 1.0)
+            if threshold == float("inf") or score >= threshold:
+                scored_items.append((score, index, item))
+
+        scored_items.sort(key=lambda result: (-result[0], result[1]))
+        return [item for _, _, item in scored_items[:max_results]]
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai addressed in https://github.com/NVIDIA-NeMo/Guardrails/pull/1970/commits/34c018ad31017d64bb6f603aa265472c744841a8, do you confirm?

tests/v2_x/test_state_serialization.py
```diff
@@ -148,5 +140,24 @@ async def test_serialization():
     assert output_events[0]["script"] == "Hello again!"
 
 
+@pytest.mark.skip(reason="Flaky wall-clock performance assertion.")
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai addressed in https://github.com/NVIDIA-NeMo/Guardrails/pull/1970/commits/34c018ad31017d64bb6f603aa265472c744841a8, do you confirm?

tests/testing/embeddings.py
```diff
@@ -0,0 +1,115 @@
...
+    normalized_item_text = _normalized_text(item_text)
+
+    if normalized_text and (
+        normalized_text == normalized_item_text
+        or normalized_text in normalized_item_text
+        or normalized_item_text in normalized_text
+    ):
+        return 1.0
+
+    return sum(a * b for a, b in zip(text_embedding, item_embedding))
```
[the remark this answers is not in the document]
Pouyanpi: let's resolve if you confirm it is addressed in the latest commit

tests/testing/embeddings.py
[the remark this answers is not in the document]
Pouyanpi: Agreed. The inf default is intentional because it matches BasicEmbeddingsIndex, but it was not obvious. I added a class docstring documenting that default behavior.
see https://github.com/NVIDIA-NeMo/Guardrails/pull/1970/commits/91237ff17c0daf8d214a257b6e91682e0595b2e6

.github/workflows/_test.yml
```diff
@@ -103,11 +103,11 @@ jobs:
 
       - name: Run pytest with coverage
         if: inputs.with-coverage
-        run: poetry run pytest --cov=nemoguardrails tests/ --cov-report=xml:coverage.xml -v
+        run: make test-coverage
 
       - name: Run pytest without coverage
         if: inputs.with-coverage == false
-        run: poetry run pytest -v
+        run: make test
```
tgasser-nv, line 103: The `poetry run pytest -v` command this replaces uses the [`pytest.ini`](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/pytest.ini#L13-L16) testpaths: tests, docs/colang-2/examples, benchmark/tests ).

The new Makefile target specifies `tests/` if nothing is provided (line 8). So the Makefile overrides pytest's default behaviour of pulling the test paths from the pytest.ini file.

The net effect is the docs/colang-2/examples and benchmark/tests dirs are dropped. The docs/colang-2/examples directory doesn't seem to have any tests, but the benchmark/tests one does. This drops 239 tests from CI coverage.

Recommend not providing the TESTS arg in the Makefile to pytest to remain compatible with the previous behaviour

tests/v2_x/test_state_serialization.py
```diff
@@ -148,5 +140,25 @@ async def test_serialization():
     assert output_events[0]["script"] == "Hello again!"
 
 
+@pytest.mark.asyncio
+async def test_serialization_performance():
```
tgasser-nv, line 145: We shouldn't have performance tests as part of our unit-tests. This threshold magic-number (2.0) has been adjusted upwards many times, and can cause flaky CI results. If we have to make the threshold so high that it always passes, it neither verifies functionality nor performance. Recommend removing this test

tgasser-nv (commented):
Needs some updates to add missing tests and remove performance-test running as a unit-test. 

Also could you run on Windows and make sure there are no issues with the Windows Makefile operation vs MacOS or Linux? The Windows checks aren't part of the PR tests

tests/v2_x/test_state_serialization.py
Pouyanpi: Yes, I'll drop it once more 👍🏻 

I need to let @greptileai know about this. It's a bad code change suggestion @greptileai, originally this flaky test was skipped and your review kind of cause this issues. see https://github.com/NVIDIA-NeMo/Guardrails/pull/1970/commits/8128bd807d63a04f510490f769f081f56fefb869

.github/workflows/_test.yml
Pouyanpi: I won't include docs/colang-2 .

I added a separate `make test-benchmark` target and a path-filtered benchmark workflow instead of folding benchmark/tests into make test (see https://github.com/NVIDIA-NeMo/Guardrails/pull/1970/commits/a3cdef2555147619d94f40832c8ea7a13321089e).

`benchmark/tests` only cover benchmark tooling and do not import core nemoguardrails code.
keeping them separate preserves make test as the core library suite. 
while still ensuring benchmark tooling is tested whenever benchmark code, dependencies, the Makefile, or the workflow changes. we also don't need the full matrix for benchmarking. What do you think Tim?

the new target passed localy: 239 passed.

Pouyanpi:
@tgasser-nv full tests were run: 

https://github.com/NVIDIA-NeMo/Guardrails/actions/runs/26814510332
https://github.com/NVIDIA-NeMo/Guardrails/actions/runs/26814315586

tests/v2_x/test_state_serialization.py
Pouyanpi: @coderabbit also note this down about flaky tests, specially performance tests being part of unit-tests.

.github/workflows/_test.yml
tgasser-nv: I'd rather have tests for everything in the repo under the same Makefile for simplicity.

tests/v2_x/test_state_serialization.py
tgasser-nv: Looks like we still have this test in as of the latest commit, can you remove it @Pouyanpi ?

tgasser-nv (approved):
Thanks for running the Windows full-tests as well. Approving with two blocking items before merging:

* Adding benchmark/tests back into the main Makefile.
* Removing the `test_serialization_performance` test

tests/v2_x/test_state_serialization.py
Pouyanpi, line 143: @tgasser-nv this is already skipped.

Pouyanpi:
@tgasser-nv 

I agree that the Makefile should be the central CI entrypoint, and that is a goal of this PR.

I do not think `benchmark/tests` should be folded into the core `make test` target or the full unittest matrix. The previous bare `poetry run pytest -v` command picked up `benchmark/tests` through `pytest.ini` `testpaths`, but that looks like accidental coupling rather than an intentional CI contract. These tests validate benchmark
tooling, not core library behavior. 


these tests validate benchmark tooling, not core library behavior. they also do not appear to be reviewed with the same rigor or stability expectations as the core, and the benchmark directory is not part of the package we ship. treating them as part of the core test gate would blur that quality boundary and add matrix cost without much signal.

the compromise II’m aiming for is:

- keep `make test` focused on the core `tests/` suite,
- keep `make test-benchmark` in the same Makefile,
- run benchmark tests through a dedicated workflow when benchmark-related files, dependencies, pytest config, the makefile, or the workflow itself change.

devs can still run bare `poetry run pytest` if they want pytest’s default `testpaths` behavior locally. For CI, the Makefile gives us explicit, named targets for the suites we intentionally run. What do you think? 


the performance test is actually skipped.

tgasser-nv:
> @tgasser-nv
> 
> I agree that the Makefile should be the central CI entrypoint, and that is a goal of this PR.
> 
> I do not think `benchmark/tests` should be folded into the core `make test` target or the full unittest matrix. The previous bare `poetry run pytest -v` command picked up `benchmark/tests` through `pytest.ini` `testpaths`, but **that looks like accidental coupling** rather than an intentional CI contract. These tests validate benchmark tooling, not core library behavior.

No, it was intentional.

> 
> these tests validate benchmark tooling, not core library behavior. **they also do not appear to be reviewed with the same rigor or stability expectations as the core**

Provide evidence for this statement. I've never asked for relaxed or less rigorous reviews for benchmark PRs. Tests are tests.

> ... the benchmark directory is not part of the package we ship.

We do ship the benchmarking tools via git clone, not via PyPI. They still have to work just as our PyPI package. We should have the same high standards for both.

> treating them as part of the core test gate would blur that quality boundary and add matrix cost without much signal.
>
> the compromise II’m aiming for is:
> 
> * keep `make test` focused on the core `tests/` suite,
> * keep `make test-benchmark` in the same Makefile,
> * run benchmark tests through a dedicated workflow when benchmark-related files, dependencies, pytest config, the makefile, or the workflow itself change.
> 
> devs can still run bare `poetry run pytest` if they want pytest’s default `testpaths` behavior locally. For CI, the Makefile gives us explicit, named targets for the suites we intentionally run. What do you think?

Your compromise is predicated on 1. benchmark tests being added to CI unintentionally, 2. Our not shipping benchmarking tools via PyPI packages justifying a lower code-quality. Neither are true.

There's no need to compromise, you're trying to justify a regression in baheviour.

tgasser-nv (changes requested):
See comment https://github.com/NVIDIA-NeMo/Guardrails/pull/1970#issuecomment-4632835962

Pouyanpi:
@tgasser-nv do you see any cases that we need to run benchmark related tests except

> when benchmark-related files, dependencies, pytest config, the makefile, or the workflow itself change.

?


I don't understand why we should run those tests as part of the core suite given the full matrix. Do we really need to run the benchmarking test on 4 python versions * 3 different OSs when the change is made to a file under `nemoguardrails/`? They don't depend on a single line of code in `nemoguardrails`.

Do we also need codecov coverage reports for those?

Pouyanpi:
@tgasser-nv 

`make test` and `make test-coverage`` no longer force TEST ?= tests/, so pytest now uses pytest.ini testpaths again. that means benchmark/tests is back in the main CI test path. I also removed the separate benchmark workflow so those tests do not run twice.

committed as:

c7ed441d9 ci: preserve pytest testpaths in make test

validation:

```
make test TEST=benchmark/tests WORKERS=2 ARGS='-q --maxfail=1'
239 passed

make test WORKERS=2 ARGS='--collect-only -q'
4661 tests collected

make test ARGS='-q --maxfail=1'
4481 passed, 180 skipped
```


If you think this is the right call please approve and merge the PR (or I'll merge when I'm back). ow I can revert this commit. TY!

tgasser-nv (approved):
Thanks for making the updates, approving

PR #2147, opened 2026-07-08
fix(injection_detection): clear error when inline yara_rules miss a requested name

WatchTree-19:
### problem
`_load_rules` in `nemoguardrails/library/injection_detection/actions.py` crashes with an opaque `KeyError` when injection detection is configured with inline `yara_rules` and the `injections` list names a rule that isn't in those inline rules (e.g. mixing a built-in name like `sqli` with an inline custom rule, or a typo):
```python
rules_source = {name: rule for name, rule in yara_rules.items() if name in rule_names}   # intersection
rules = yara.compile(sources={rule_name: rules_source[rule_name] for rule_name in rule_names})  # indexes ALL names -> KeyError
```
neither `_validate_injection_config` nor `_extract_injection_config` catches this (name validation is skipped when inline rules are used), so the guardrail dies with `KeyError: 'sqli'` instead of a clear config error.

### repro
```python
from pathlib import Path
from nemoguardrails.library.injection_detection.actions import _load_rules
_load_rules(Path("."), ("sqli", "custom_rule"),
            {"custom_rule": 'rule custom_rule { strings: $a = "foo" condition: $a }'})
# KeyError: 'sqli'
```

### fix
validate up front and raise a clear, actionable `ValueError` listing the missing names (fail-closed, mirroring the `ValueError` the `yara_path` branch already raises for unknown rules), then compile directly since every name is now guaranteed present.

### test
added `test_inline_yara_rules_missing_injection_name_raises_value_error` (mirrors `test_load_inline_yara_rules`). fails on `develop` (`KeyError`), passes with the fix; full `test_injection_detection.py` 32 pass, ruff clean.

found by reading the code. (a maintainer could alternatively hoist the check into `_extract_injection_config` next to the existing `yara_path` name validation for earlier failure - happy to move it.)

### DIFF
--- a/nemoguardrails/library/injection_detection/actions.py
+++ b/nemoguardrails/library/injection_detection/actions.py
@@ -168,10 +168,25 @@ def _load_rules(
         log.warning("Injection config was provided but no modules were specified. Returning None.")
         return None
 
+    if yara_rules:
+        # When inline `yara_rules` are provided, only those rules are available
+        # (built-in rules are not loaded in this mode). Every requested rule name
+        # in `injections` must therefore be present in `yara_rules`; otherwise we
+        # fail with a clear, actionable error instead of an opaque KeyError.
+        missing_rules = [name for name in rule_names if name not in yara_rules]
+        if missing_rules:
+            msg = (
+                "Provided set of `injections` in injection config contains rule names "
+                "not present in the inline `yara_rules`: %r. Available inline rules are %r. "
+                "Note that built-in rules are not loaded when `yara_rules` is provided."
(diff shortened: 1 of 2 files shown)


WatchTree-19:
thanks for the review! one note on the signature check: the commit is DCO signed-off (`-s`) but not GPG/SSH signed - i don't currently have commit signing set up on this account. would a squash-merge work on your end (github signs the squash commit, which should satisfy the require-signed-commits check), or do you need a signed commit on the branch? happy to sort it whichever way is easiest for you.

PR #2145, opened 2026-07-08
fix(colang): raise ValueError for unbalanced closing bracket in split_args

vidigoat:
Fixes #2144

### Description
`split_args` reads `closing_char[stack[-1]]` for every closing character, but when a `)`, `]`, or `}` appears with an empty bracket stack, `stack[-1]` raises an uncaught `IndexError`. (Quote characters are unaffected — they are treated as openers on an empty stack.) This adds an explicit empty-stack guard that raises the same `ValueError` the function already uses to report malformed input, so callers get a consistent, catchable error instead of an `IndexError`.

### Steps to reproduce (before this fix)
```python
from nemoguardrails.colang.v1_0.lang.utils import split_args
split_args("x=1, y)")   # IndexError: list index out of range
```
Also affected: `")"`, `"a)b"`, `"foo]"`, `"}"`.

### Test plan
- Added a parametrized regression test in `tests/test_parser_utils.py` covering `")"`, `"a)b"`, `"foo]"`, `"x=1, y)"`, `"}"`; each now asserts `ValueError`.
- Verified the new test fails before the fix (`IndexError`) and passes after; existing `split_args` tests and the mismatched-bracket `ValueError` path are unchanged.

### AI disclosure
This change was drafted with the assistance of an AI coding assistant; I reproduced the bug and verified the fix and tests before submitting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved argument parsing to handle unexpected closing brackets more gracefully.
  * Replaced an internal crash case with a clear error message when bracket or quote characters are unbalanced.
  * Added coverage for several malformed input examples to prevent regressions.

Issue #2144, opened 2026-07-08
split_args raises IndexError on a stray closing bracket instead of ValueError


vidigoat:
### Did you check docs and existing issues?

- [x] I have read all the NeMo-Guardrails docs
- [x] I have updated the package to the latest version before submitting this issue
- [ ] (optional) I have used the develop branch
- [x] I have searched the existing issues of NeMo-Guardrails

### Python version (python --version)

Python 3.11

### Operating system/version

MacOS 26

### Describe the bug

nemoguardrails.colang.v1_0.lang.utils.split_args` keeps a bracket `stack` and,
on any closing character in )]}"' , immediately reads `closing_char[stack[-1]]`.
When a closing bracket ) ] or } appears with an empty stack (no matching opener),
`stack[-1]` raises `IndexError: list index out of range`. The function otherwise
reports malformed input via `ValueError` (the following lines), so this should
raise `ValueError`, not crash.

### Steps To Reproduce

```python
from nemoguardrails.colang.v1_0.lang.utils import split_args
split_args("x=1, y)")   # IndexError: list index out of range
Also affected: ")", "a)b", "foo]", "}"

### Expected Behavior

a ValueError describing the invalid syntax (consistent with the
mismatched-bracket branch already in the function).
Actual: uncaught IndexError.

### Actual Behavior

`split_args("x=1, y)")` raises an uncaught `IndexError: list index out of range`
(from reading `stack[-1]` on an empty stack) instead of the `ValueError` the
function already uses to report invalid syntax. Same for ")", "a)b", "foo]", "}".

Pouyanpi:
Thanks @vidigoat, I would approve the PR but in future first open an issue. Cnce the issue is triaged and assigned to you, you can open a PR. See https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/CONTRIBUTING.md

vidigoat:
Thanks @Pouyanpi — understood, and sorry for jumping ahead. Going forward I'll open an issue first and wait for it to be triaged and assigned before opening a PR. Appreciate you assigning #2144 and taking the time to look.

PR #2164, opened 2026-07-11
Handle brackets inside string arguments in split_args

winklemad:
## Description

`split_args` (the Colang 1.0 argument parser in `nemoguardrails/colang/v1_0/lang/utils.py`) treats every bracket and quote as structural. While inside a quoted string argument, a closing bracket `)`, `]`, `}` — or the other quote character — is still handled as syntax, so it raises `ValueError` and crashes flow parsing.

This hits ordinary string values: an emoticon, a citation marker, or any clause that ends with a bracket.

```python
>>> from nemoguardrails.colang.v1_0.lang.utils import split_args
>>> split_args('msg="Hello :)"')
ValueError: Invalid syntax for string: msg="Hello :)"; expecting " and got )
```

It also crashes through the public parse path — any Colang 1.0 flow with such an inline string parameter fails to parse:

```python
>>> from nemoguardrails.colang.v1_0.lang.coyml_parser import parse_flow_elements
>>> parse_flow_elements([{"user": 'ask(reason="closed ]")'}])
ValueError: Invalid syntax for string: reason="closed ]"; expecting " and got ]
```

The function's own docstring says it "correctly handles strings and lists/dicts", and the sibling splitters in the same module (`word_split`, `char_split`, `params_tokenize`) already track string state — only `split_args` doesn't.

## Fix

Add a real in-string short-circuit: while the top of the stack is a quote, only the matching closing quote is special; brackets, commas, and the other quote type are literal text. This preserves list/dict argument parsing and still raises `ValueError` for genuinely unbalanced *unquoted* brackets (the behavior added in #2145).

## Test Plan

Added `test_split_args_bracket_inside_string` to `tests/test_parser_utils.py`, covering closing brackets `)]}`, an emoticon, a citation marker, a comma, and the other quote character inside a quoted string, plus a bracket-in-string argument alongside a normal one. The new cases fail on `develop` (`ValueError`) and pass with the fix; the existing `test_1` and the unbalanced-unquoted-bracket cases still pass.

```
$ pytest tests/test_parser_utils.py tests/colang
70 passed

$ ruff check nemoguardrails/colang/v1_0/lang/utils.py tests/test_parser_utils.py
All checks passed!
$ ruff format --check nemoguardrails/colang/v1_0/lang/utils.py tests/test_parser_utils.py
2 files already formatted
```

No linked issue — found by reading the code.

### DIFF
--- a/nemoguardrails/colang/v1_0/lang/utils.py
+++ b/nemoguardrails/colang/v1_0/lang/utils.py
@@ -50,13 +50,21 @@ def split_args(args_str: str) -> List[str]:
     closing_char = {"[": "]", "(": ")", "{": "}", "'": "'", '"': '"'}
 
     for char in args_str:
+        # While inside a string literal, only the matching closing quote is
+        # special; brackets, commas and the other quote type are literal text.
+        if stack and stack[-1] in "\"'":
+            if char == stack[-1]:
+                stack.pop()
+            current.append(char)
+            continue
+
         if char in "([{":
             stack.append(char)
             current.append(char)
-        elif char in "\"'" and (len(stack) == 0 or stack[-1] != char):
+        elif char in "\"'":
             stack.append(char)
             current.append(char)
-        elif char in ")]}\"'":
+        elif char in ")]}":
             if not stack:
                 raise ValueError(f"Invalid syntax for string: {args_str}; unexpected closing '{char}'")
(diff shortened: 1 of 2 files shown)


Pouyanpi:
duplicate of #2145 note that #2144 is already closed

PR #2209, opened 2026-07-23
feat(http): add canonical outbound HTTP foundation

Pouyanpi:
## Description

Define the transport-neutral outbound HTTP foundation used by built-in library
rails.

The neutral response owns decoded bytes rather than exposing a live transport
response. Callers can therefore inspect a response after the underlying client
context closes. The HTTPX adapter provides pooled connections while keeping
transport-specific types out of rail actions.

Retry behavior is bounded and conservative by default. Safe methods are
retryable, POST retries require an explicit rail-owned policy, and vendor
override headers are ignored unless a policy opts in. `Retry-After` remains
bounded by policy, total request timeouts cover the complete operation, and
redirect following is disabled unless explicitly enabled.

### Design

There is one real transport and a set of thin, single-purpose decorators, not a
family of client implementations. `HTTPClient` is a runtime-checkable protocol,
so `HttpxHTTPClient`, a recording fake, and any future `aiohttp` adapter satisfy
the same contract without sharing a base class. Rail actions and tests depend on
the protocol and never import a concrete transport.

Behavior is added by composition. The default stack is
`InstrumentedHTTPClient(RetryingHTTPClient(HttpxHTTPClient(...)))`, where each
layer satisfies `HTTPClient` and wraps the next:

- `HttpxHTTPClient` is the only layer that touches the network. It owns the
  connection pool, the total-operation timeout, and TLS.
- `RetryingHTTPClient` re-issues a request under a bounded policy and knows
  nothing about HTTPX or telemetry.
- `InstrumentedHTTPClient` opens one span per logical call and knows nothing
  about HTTPX or retries. It is a transparent pass-through when no tracer is
  supplied, so importing the module never requires the OpenTelemetry SDK.

This design keeps each concern independently testable and configurable: retry
logic is unit-tested against a fake transport with no HTTPX and no telemetry,
and a layer is only present when it is needed (retries only when a policy is
passed, instrumentation only when tracing is enabled). The composition order is
load-bearing. Instrumenting outside retry yields one span per logical operation
that records the resend count. The reverse order would emit one sibling span per
attempt and make a single operation look like several unrelated calls.

### Impact

- Library integrations gain one transport-neutral HTTP contract.
- Response parsing and status errors no longer depend on a live HTTPX response.
- POST requests are not retried unless the integration explicitly declares
  that resending is safe.
- Redirect, timeout, retry-header, and backoff behavior is deterministic and
  covered by focused tests.
- No library rail is migrated in this PR; call-site migrations remain isolated
  in later stack layers.

### Stack

| Layer | Branch | Base | PR |
| --- | --- | --- | --- |
| HTTP foundation | `pouyanpi/rail-library-stack-9-http-contract` | `develop` | #2209 |
| HTTP instrumentation | `pouyanpi/rail-library-stack-10-http-instrumentation` | Stack 9 | #2219 |
| Request lifecycle | `pouyanpi/rail-library-stack-11-http-observability` | Stack 9 | #2210 |
| Request migrations | `pouyanpi/rail-library-stack-12-http-request-migrations` | Stack 11 | #2211 |
| Action and specialized migrations | `pouyanpi/rail-library-stack-13-http-action-migrations` | Stack 12 | #2212 |

## Summary by CodeRabbit

* **New Features**
  * Added an asynchronous HTTP client with request and response handling.
  * Added response text/JSON parsing and clear status, timeout, connection, and decoding errors.
  * Added configurable retries with exponential backoff, retry limits, and `Retry-After` support.
  * Added optional redirect handling, timeout enforcement, and safe resource cleanup.
* **Tests**
  * Added comprehensive coverage for HTTP requests, errors, retries, timeouts, redirects, and response parsing.

### DIFF
--- a/nemoguardrails/http/errors.py
+++ b/nemoguardrails/http/errors.py
@@ -26,7 +26,8 @@ def _safe_url(url: str) -> str:
     parts = urlsplit(url)
     hostname = parts.hostname
     if hostname is None:
-        return urlunsplit((parts.scheme, parts.netloc, parts.path, "", ""))
+        netloc = parts.netloc.rsplit("@", 1)[-1]
+        return urlunsplit((parts.scheme, netloc, parts.path, "", ""))
     host = f"[{hostname}]" if ":" in hostname else hostname
     try:
         port = parts.port
--- a/nemoguardrails/http/errors.py
+++ b/nemoguardrails/http/errors.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Transport-neutral exceptions raised by the outbound HTTP subsystem."""
+
 from typing import TYPE_CHECKING
 from urllib.parse import urlsplit, urlunsplit
 
@@ -35,20 +37,33 @@ def _safe_url(url: str) -> str:
 
 
 class HTTPClientError(Exception):
+    """Base class for outbound HTTP failures.
+
+    ``retry_count`` records how many retries completed before the error was
+    returned to the caller.
+    """
+
     def __init__(self, *args: object):
         super().__init__(*args)
         self.retry_count = 0
 
 
 class HTTPConnectionError(HTTPClientError):
-    pass
+    """Raised when the transport cannot establish or maintain a connection."""
 
 
 class HTTPTimeoutError(HTTPClientError):
-    pass
+    """Raised when an outbound request exceeds its total timeout."""
 
 
 class HTTPStatusError(HTTPClientError):
+    """Raised for an unsuccessful HTTP status.
+
+    Attributes:
+        response: The materialized response that produced the error.
+        request: Request metadata when supplied by the caller.
+    """
+
     def __init__(self, response: "HTTPResponse", request: "HTTPRequest | None" = None):
         message = f"HTTP request failed with status {response.status_code}"
         if request is not None:
@@ -59,6 +74,8 @@ class HTTPStatusError(HTTPClientError):
 
 
 class HTTPResponseDecodeError(HTTPClientError):
+    """Raised when a response body cannot be decoded as JSON."""
+
     def __init__(self, response: "HTTPResponse"):
         super().__init__(f"HTTP response body is not valid JSON (status {response.status_code})")
         self.response = response
--- a/nemoguardrails/http/errors.py
+++ b/nemoguardrails/http/errors.py
@@ -16,25 +16,11 @@
 """Transport-neutral exceptions raised by the outbound HTTP subsystem."""
 
 from typing import TYPE_CHECKING
-from urllib.parse import urlsplit, urlunsplit
+
+from nemoguardrails.http._url import sanitize_url
 
 if TYPE_CHECKING:
     from nemoguardrails.http.types import HTTPRequest, HTTPResponse
-
-
-def _safe_url(url: str) -> str:
-    parts = urlsplit(url)
-    hostname = parts.hostname
-    if hostname is None:
-        netloc = parts.netloc.rsplit("@", 1)[-1]
-        return urlunsplit((parts.scheme, netloc, parts.path, "", ""))
-    host = f"[{hostname}]" if ":" in hostname else hostname
-    try:
-        port = parts.port
-    except ValueError:
-        port = None
-    netloc = f"{host}:{port}" if port is not None else host
-    return urlunsplit((parts.scheme, netloc, parts.path, "", ""))
 
 
 class HTTPClientError(Exception):
@@ -68,7 +54,7 @@ class HTTPStatusError(HTTPClientError):
     def __init__(self, response: "HTTPResponse", request: "HTTPRequest | None" = None):
         message = f"HTTP request failed with status {response.status_code}"
         if request is not None:
-            message = f"{message}: {request.method.upper()} {_safe_url(request.url)}"
+            message = f"{message}: {request.method.upper()} {sanitize_url(request.url)}"
         super().__init__(message)
         self.response = response
         self.request = request
--- a/nemoguardrails/http/errors.py
+++ b/nemoguardrails/http/errors.py
@@ -1,0 +1,62 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TYPE_CHECKING
+from urllib.parse import urlsplit, urlunsplit
+
+if TYPE_CHECKING:
+    from nemoguardrails.http.types import HTTPRequest, HTTPResponse
+
+
+def _safe_url(url: str) -> str:
+    parts = urlsplit(url)
+    hostname = parts.hostname
+    if hostname is None:
+        return urlunsplit((parts.scheme, parts.netloc, parts.path, "", ""))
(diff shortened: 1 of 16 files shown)


nemoguardrails/http/transport.py
```diff
@@ -0,0 +1,141 @@
...
+        if timeout is not None and timeout <= 0:
+            raise ValueError("HTTP timeout must be greater than zero")
+        self._owns_client = client is None
+        self._timeout = timeout if self._owns_client else None
+        self._client = client or httpx.AsyncClient(
+            timeout=None,
+            limits=limits or httpx.Limits(max_connections=100, max_keepalive_connections=20),
+            follow_redirects=follow_redirects,
+        )
+        self._closed = False
```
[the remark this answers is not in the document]
Pouyanpi: resolved in latest commits, do you confirm?

nemoguardrails/http/transport.py
```diff
@@ -0,0 +1,141 @@
...
+            response = await asyncio.wait_for(
+                self._client.request(method, url, **kwargs),
+                timeout=deadline,
+            )
+        except asyncio.TimeoutError as error:
+            raise HTTPTimeoutError("HTTP request timed out") from error
+        except httpx.TimeoutException as error:
+            raise HTTPTimeoutError("HTTP request timed out") from error
+        except httpx.TransportError as error:
+            raise HTTPConnectionError("HTTP transport failed") from error
```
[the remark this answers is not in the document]
Pouyanpi: resolved in latest commits, do you confirm?

nemoguardrails/http/errors.py
```diff
@@ -0,0 +1,81 @@
...
+    hostname = parts.hostname
+    if hostname is None:
+        return urlunsplit((parts.scheme, parts.netloc, parts.path, "", ""))
+    host = f"[{hostname}]" if ":" in hostname else hostname
+    try:
+        port = parts.port
+    except ValueError:
+        port = None
+    netloc = f"{host}:{port}" if port is not None else host
+    return urlunsplit((parts.scheme, netloc, parts.path, "", ""))
```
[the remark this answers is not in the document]
Pouyanpi: resolved in latest commits, do you confirm?

nemoguardrails/http/transport.py
```diff
@@ -0,0 +1,141 @@
...
+        return HTTPResponse(
+            status_code=response.status_code,
+            headers=dict(response.headers),
+            content=response.content,
+            extensions={"http_version": response.http_version},
+        )
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai

dict(response.headers) already produces comma-joined duplicate values. you incorrectly assumes normal header iteration yields every occurrence; that requires response.headers.multi_items().

  the suggested replacement produces the same result:

  {"set-cookie": "a=1, b=2"}


skip it as INVALID

nemoguardrails/http/transport.py
```diff
@@ -0,0 +1,141 @@
...
+        }
+        deadline = timeout if timeout is not None else self._timeout
+        if deadline is not None and deadline <= 0:
+            raise ValueError("HTTP timeout must be greater than zero")
+        try:
+            response = await asyncio.wait_for(
+                self._client.request(method, url, **kwargs),
+                timeout=deadline,
+            )
+        except asyncio.TimeoutError as error:
```
tgasser-nv, line 126: When raising these exceptions from `error`, it could be an `https.RequestError` or `https.TimeoutException`. These both have a `request.url` which contains the full credentials and query URL. This might leak any credentials. Recommend sanitizing the URL in case it contains any keys or creds

nemoguardrails/http/transport.py
```diff
@@ -0,0 +1,141 @@
...
+    """Adapt an HTTPX asynchronous client to the neutral HTTP contract.
+
+    A client created by this adapter is owned and closed by the adapter. An
+    injected client remains owned by its caller. The configured timeout covers
+    the complete request rather than each HTTPX timeout phase independently.
+    """
+
+    def __init__(
+        self,
+        client: httpx.AsyncClient | None = None,
```
tgasser-nv, line 49: If a user provides a `client` (httpx.AsyncClient) themselves, the `limits` and `follow_redirects` aren't used. The `timeout` value is still validated. Recommend raising a `ValueError` if client is provided with a non-default `timeout`, `limits` or `follow_redirects`

nemoguardrails/http/retry.py
```diff
@@ -0,0 +1,227 @@
...
+            try:
+                parsed = parsedate_to_datetime(value)
+            except (TypeError, ValueError):
+                return None
+            if parsed.tzinfo is None:
+                parsed = parsed.replace(tzinfo=timezone.utc)
+            delay = (parsed - now).total_seconds()
+        if self.clamp_retry_after:
+            return min(max(delay, 0.0), self.max_retry_after)
+        if 0 < delay <= self.max_retry_after:
```
tgasser-nv: nit: If we're not clamping than a value of 0 (retry immediately) is valid under [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3) . So this should be `0<= delay <= ...`

nemoguardrails/http/retry.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+
+        self._client = client
+        self._policy = policy or RetryPolicy()
+        self._sleep = sleep
+        self._random_value = random_value
+        self._now = now or (lambda: datetime.now(timezone.utc))
+        self._closed = False
+
+    async def request(
```
tgasser-nv, line 150: nit: The `retry_count` can be surfaced through either `response.extensions["retry_count"]` or `error.retry_count`. If an `HttpxHTTPClient` client without a retry decorator is used, the `response.extensions["retry_count"]` doesn't contain a retry_count.

Is there a way to simplify this, maybe with a getter like `retry_count` which checks the object type and gives a uniform way to access retry_count

```
def retry_count(obj: HTTPResponse | HTTPClientError) -> int:
    if isinstance(obj, HTTPResponse):
        return int(obj.extensions.get("retry_count", 0))
    return getattr(obj, "retry_count", 0)
```

nemoguardrails/testing/http.py
```diff
@@ -0,0 +1,78 @@
...
+"""Deterministic HTTP test doubles for NeMo Guardrails tests."""
+
+from collections import deque
+from collections.abc import Iterable
+from typing import Any, Mapping
+
+from nemoguardrails.http.types import HTTPRequest, HTTPResponse
+
+
+class RecordingHTTPClient:
```
tgasser-nv, line 25: Should this be included under `nemoguardrails` or moved out under `tests`?

tgasser-nv (approved):
Looks good! Can you check the code coverage and add tests to improve this? Looks like missing lines in the owned-client request-path (where client is None) and a couple of other places.

nemoguardrails/testing/http.py
Pouyanpi: Yes, you're right! I moved it under `nemoguardrails/testing` and updated the doc

Pouyanpi: https://github.com/NVIDIA-NeMo/Guardrails/pull/2209/commits/cbc9f131ab37b4b6326a8ccda53b02cbf52409b4

nemoguardrails/http/retry.py
Pouyanpi: Done 👍🏻 https://github.com/NVIDIA-NeMo/Guardrails/pull/2209/commits/52c1a45e6eb6a1b3959492590a511b9c92be4c56

nemoguardrails/http/transport.py
Pouyanpi: change is made in https://github.com/NVIDIA-NeMo/Guardrails/pull/2209/commits/653654261241f18d4baf24b58fdb1e12f0ad57e0

nemoguardrails/http/transport.py
Pouyanpi: Good point 👍🏻 URLs are sanitized now https://github.com/NVIDIA-NeMo/Guardrails/pull/2209/commits/69b529aaf39c934364d952eddc60f1fe9130068b

nemoguardrails/http/retry.py
Pouyanpi: Reasonable suggestion, but shall we keep the API minimal until a consumer needs uniform retry-count access across responses and errors?

Pouyanpi:
Thanks @tgasser-nv for the review. I've added coverage and addressed your review comments.

PR #2240, opened 2026-07-31
fix(server): correct thread and model injection errors

Pouyanpi:
## Description

  Fixes two server issues:

  - Return HTTP 400 when `thread_id` is used without a configured datastore.
  - Preserve the main model’s `api_key_env_var` during request model injection.

  Includes regression tests for both fixes.

## AI Assistance

- [ ] No AI tools were used.
- [ ] AI tools were used; a human reviewed and can explain every change (tool: ___).

## Checklist

- [ ] This PR links to a triaged issue assigned to me.
- [ ] My PR title follows the project commit convention.
- [ ] I've noted any verification beyond CI and any checks I couldn't run.
- [ ] I did not update generated changelog files manually.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved configured API-key environment variables, base URLs, and default headers when updating model settings.
  * Returned a structured HTTP 400 error when thread-based requests are made without a configured datastore.
  * Ensured model configuration updates do not modify the original configuration.

### DIFF
--- a/nemoguardrails/server/api.py
+++ b/nemoguardrails/server/api.py
@@ -641,12 +641,9 @@ async def chat_completion(body: GuardrailsChatCompletionRequest, request: Reques
 
     if body.guardrails.thread_id:
         if datastore is None:
-            raise RuntimeError("No DataStore has been configured.")
-        # We make sure the `thread_id` meets the minimum complexity requirement.
-        if len(body.guardrails.thread_id) < 16:
             raise HTTPException(
-                status_code=422,
-                detail="The `thread_id` must have a minimum length of 16 characters.",
+                status_code=400,
+                detail="Conversation threads are not enabled on this server.",
             )
 
         # Fetch the existing thread messages. For easier management, we prepend
--- a/nemoguardrails/server/api.py
+++ b/nemoguardrails/server/api.py
@@ -381,9 +381,14 @@ def _update_models_in_config(config: RailsConfig, main_model: Model) -> RailsCon
             break
 
     if main_model_index is not None:
-        parameters = {**models[main_model_index].parameters, **main_model.parameters}
-        models[main_model_index] = main_model
-        models[main_model_index].parameters = parameters
+        configured_model = models[main_model_index]
+        parameters = {**configured_model.parameters, **main_model.parameters}
+        models[main_model_index] = main_model.model_copy(
+            update={
+                "api_key_env_var": main_model.api_key_env_var or configured_model.api_key_env_var,
+                "parameters": parameters,
+            }
+        )
     else:
         models.append(main_model)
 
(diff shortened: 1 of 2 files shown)


nemoguardrails/server/api.py
```diff
@@ -381,9 +381,14 @@ def _update_models_in_config(config: RailsConfig, main_model: Model) -> RailsCon
             break
 
     if main_model_index is not None:
-        parameters = {**models[main_model_index].parameters, **main_model.parameters}
-        models[main_model_index] = main_model
-        models[main_model_index].parameters = parameters
+        configured_model = models[main_model_index]
+        parameters = {**configured_model.parameters, **main_model.parameters}
```
tgasser-nv, line 385: Should we update the engine from the configured_model, rather than use the `MAIN_MODEL_ENGINE` default of `openai` here?

nemoguardrails/server/api.py
```diff
@@ -636,12 +641,9 @@ async def chat_completion(body: GuardrailsChatCompletionRequest, request: Reques
 
     if body.guardrails.thread_id:
         if datastore is None:
-            raise RuntimeError("No DataStore has been configured.")
-        # We make sure the `thread_id` meets the minimum complexity requirement.
-        if len(body.guardrails.thread_id) < 16:
             raise HTTPException(
-                status_code=422,
-                detail="The `thread_id` must have a minimum length of 16 characters.",
+                status_code=400,
```
tgasser-nv, line 645: Should this error code match the Colang 2 exception raised on line 613? They are 422 and 400 respectively

tests/server/test_api.py
```diff
@@ -166,6 +168,68 @@ def test_model_field_independent_of_config_id():
...
+                        "default_headers": {"X-Tenant": "acme"},
+                    },
+                }
+            ]
+        }
+    )
+
+    injected = api._inject_model(config, "requested-model")
+    main_model = injected.models[0]
+    headers = ModelEngine(main_model)._prepare_request([{"role": "user", "content": "hi"}]).headers
```
tgasser-nv, line 198: Could you check this for LLMRails also, to make sure the headers propagate correctly in both engines?

tgasser-nv (commented):
Looks good, please take a look at the comments before merging

nemoguardrails/server/api.py
Pouyanpi: keeping `MAIN_MODEL_ENGINE` is intentional. the server documents it as authoritatiev when the request specifies `model`, with `openai` as the default. current fix only preserves `api_key_env_var`; using `configured_model.engine` would change the existing routing behavior.

nemoguardrails/server/api.py
Pouyanpi: 400 is intentional for a request/server-configuration mismatch, consistent with `StreamingNotSupportedError`.
the Colang 2 check is semantic state validation, so it remains 422. schema validation also continues returning 422 for invalid `thread_id` values.

tests/server/test_api.py
Pouyanpi: Yes, added the regression test and it passes 👍🏻

PR #1992, opened 2026-06-04
feat(library): add Agent Threat Rules (ATR) detection rail

eeee2345:
Closes #1991

Adds a `library/atr/` input rail backed by Agent Threat Rules (ATR), an open MIT-licensed detection standard for AI-agent attacks, via the `pyatr` package.

ATR is already shipped in Cisco AI Defense and in Microsoft's agent-governance-toolkit. The rules are bundled in the `pyatr` PyPI package and run locally, with no API key or network call.

Changes:
- `nemoguardrails/library/atr/` (`actions.py`, `flows.co`, `flows.v1.co`, `__init__.py`): an `@action` (`atr_detection`) that evaluates the user message with `pyatr` and flags matches at or above a configurable severity (default `critical`/`high`). Mirrors the `injection_detection` rail.
- `pyatr` is lazy-imported with a `pip install pyatr` hint, the same optional-dependency pattern as `yara` for `injection_detection`, so no hard dependency is added.
- `tests/test_atr_rail.py` + `tests/test_configs/atr/config.yml`.
- Docs: an Agent Threat Rules section in `docs/configure-rails/guardrail-catalog/agentic-security.md`.
- CHANGELOG entry under Unreleased.

Usage:

    rails:
      input:
        flows:
          - atr detection

Tested locally against nemoguardrails 0.22.0 + pyatr 0.2.6: the rail flags prompt-injection input, passes benign input, and the `atr_detection` action registers while the `atr detection` flow loads. `black` and the new tests pass.

## Release Notes

* **New Features**
  * Added Agent Threat Rules (ATR) detection input rail for identifying and blocking agent attacks (prompt injection, jailbreaks, tool poisoning, MCP attacks) locally without API calls or additional dependencies.

* **Documentation**
  * Added ATR configuration documentation and setup instructions.

* **Tests**
  * Added comprehensive test coverage for ATR detection functionality.

Issue #1991, opened 2026-06-04
Add an Agent Threat Rules (ATR) detection library rail


eeee2345:
The library already ships detection rails from a range of providers (ActiveFence, Cisco AI Defense, CrowdStrike AIDR, Trend Micro, Prompt Security, and others). I would like to contribute one for Agent Threat Rules (ATR).

ATR is an open, MIT-licensed detection standard for AI-agent attacks: prompt injection, jailbreak, tool poisoning, MCP attacks, and skill compromise. It is already shipped in Cisco AI Defense and in Microsoft's agent-governance-toolkit. The rule set is distributed as the pyatr PyPI package and runs locally, with no API key or network call.

Proposal: a library/atr/ rail mirroring injection_detection. An @action evaluates the user message against the bundled ATR rules via pyatr and flags matches at or above a configurable severity (default critical/high), with flows.co + flows.v1.co and a test. pyatr would be lazy-imported with a "pip install pyatr" hint, the same optional-dependency pattern as yara for injection_detection, so no hard dependency is added to the package.

The implementation is ready (rail + test + docs + CHANGELOG) and I will open a PR referencing this issue.

eeee2345:
@Pouyanpi — a small process request, and a finding that explains why the earlier PR never moved.

**Request:** could this issue be triaged and assigned? The PR template requires "This PR links to a triaged issue assigned to me", and this issue has no labels and no assignee, so the PR that referenced it had no path to review regardless of its contents. Happy for the answer to be "not a fit" — that closes it cleanly.

**Finding:** #1992 predates the rail-manifest migration and could not have merged in its shape. A library package with actions and no `rail.py` fails `test_library_action_packages_declare_manifests`, and its action returned a `TypedDict` rather than `RailOutcome`, which fails `test_portable_surface_actions_declare_rail_outcome_returns`.

While rebasing onto the new structure I also found a defect worth flagging on its own: the Colang 1 flow was declared `define flow` rather than `define subflow`. As a top-level flow the runtime activated it independently of configuration — it ran for configs that never enabled it and overrode the configured rail's refusal message. Locally that surfaced as `injection_detection_reject_blocks_injection` failing with an ATR refusal string in place of the injection one. It is fixed, with a regression guard.

I have closed #1992 and opened a replacement against the current `develop`.

PR #2251, opened 2026-08-04
feat(library): add Agent Threat Rules (ATR) detection rail

eeee2345:
## Description

Adds an input rail that matches the user message against [Agent Threat Rules](https://github.com/Agent-Threat-Rule/agent-threat-rules), an open MIT-licensed detection catalog for AI-agent attacks (prompt injection, jailbreak, tool poisoning, MCP attacks, skill compromise). Rules ship inside the optional `pyatr` package and are evaluated in-process, so the rail needs no API key, no service endpoint and no network access, and the same input yields the same verdict on every run. It is modelled on `injection_detection`, the other local optional-dependency rail.

This supersedes #1992, which predates the rail-manifest migration and could not have merged in its shape — a library package with actions and no `rail.py` fails `test_library_action_packages_declare_manifests`. Rebasing was not enough; the module needed rewriting to the manifest pattern.

**Areas worth careful review:**

- `rail.py` declares `RailPrivacy()` with every disclosure flag false and no remote services. Please sanity-check that this is the intended way to express a rail that contacts nothing, rather than simply leaving the field at its default.
- `actions.py` returns `RailOutcome` and carries matched rule IDs plus max severity in metadata **on allow as well as block**, so sub-threshold matches remain visible in traces. Say the word if you would rather allow paths carried empty metadata.
- `schemas/rails_config.snapshot.json` is regenerated. The diff is purely additive (+25 / -0) and contains only the `atr` config key and `ATRDetection` definition.

**One defect found while porting, flagged separately** because it is the kind of thing that would have been unpleasant in a release: the Colang 1 flow in #1992 was declared `define flow` rather than `define subflow`. As a top-level flow the runtime activated it independently of configuration — it ran for configs that never enabled the rail and overrode the configured rail's refusal message, rendering an ATR message with an empty rule list. Locally that showed up as `injection_detection_reject_blocks_injection` failing with an ATR refusal string in place of the injection one. Both flows now emit the shared `bot refuse to respond` intent instead of an inline templated `bot say`, and `test_colang_1_flow_is_declared_as_a_subflow` guards the declaration.

`block_severities: []` is honored as monitor-only; only an absent config section falls back to `["critical", "high"]`.

## Related Issue(s)

- #1991 — I have asked there for triage and assignment, since the issue currently has no labels or assignee. Happy for the answer to be "not a fit"; that closes it cleanly and I will withdraw this PR.

## Verification

Run locally against `develop` @ `origin/develop`, Python 3.12, `pyatr` 0.2.7:

- `tests/test_atr_rail.py`, `tests/manifests/`, `tests/rails/llm/`, `tests/test_runtime_flow_gate_equivalence.py` — **399 passed**.
- Baselined every failure against clean `develop` before treating it as mine; all four initial failures were caused by this change and are fixed, not suppressed.
- `ruff format --check` and `ruff check` clean on the touched files.
- `uv lock --check` passes.

Not run: the recorded/live-credential suites and the docs build.

Residual risk: `uv.lock` is regenerated with uv 0.10.1. The marker normalization and `exclude-newer` churn in that file comes from the uv version rather than from this change — `uv-latest.yml` already runs `uv lock --check` on latest uv, but if you pin a different version please regenerate.

## AI Assistance

- [ ] No AI tools were used.
- [x] AI tools were used; a human reviewed and can explain every change (tool: Claude Code).

## Checklist

- [ ] This PR links to a triaged issue assigned to me. — #1991 is open but not yet triaged or assigned; requested in the issue.
- [x] I've noted any verification beyond CI and any checks I couldn't run.
- [ ] I addressed all CodeRabbit, Greptile, and other review comments, or replied with why no change is needed. — none yet on this PR.
- [x] @mentions of the person or team responsible for reviewing proposed changes. — @Pouyanpi

### DIFF
--- a/docs/configure-rails/guardrail-catalog/agentic-security.mdx
+++ b/docs/configure-rails/guardrail-catalog/agentic-security.mdx
@@ -154,6 +154,20 @@ Texts shorter than `min_chars` (default `50`) are checked against the size cap o
 
 Add the flow for each direction you want to guard, and set the thresholds under `rails.config.context_bloat_detection`.
 All configuration fields are optional.
+## Agent Threat Rules (ATR)
+
+The NeMo Guardrails library can evaluate input against [Agent Threat Rules (ATR)](https://github.com/Agent-Threat-Rule/agent-threat-rules), an open, MIT-licensed detection standard for AI-agent attacks such as prompt injection, jailbreak, tool poisoning, MCP attacks, and skill compromise.
+
+The rules are bundled inside the [`pyatr`](https://pypi.org/project/pyatr/) package and run locally -- no API key or network call.
+As an input rail, the rail evaluates the user message and flags content matching a rule at or above a configured severity.
+It is intended as a fast, deterministic first gate as part of a defense-in-depth strategy.
+A flagged input emits the standard `bot refuse to respond` intent, so the refusal wording is configured in one place for every rail; set `enable_rails_exceptions` to raise `AtrDetectionRailException` instead.
+
+### Configuring Agent Threat Rules
+
+Install the optional dependency with `pip install "nemoguardrails[atr]"` (or `pip install pyatr`).
+
+To activate the rail, include the `atr detection` input flow:
 
 ```yaml
 rails:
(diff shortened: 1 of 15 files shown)


eeee2345:
@tgasser-nv — this has been sitting in `needs triage` for about two weeks, so a quick pointer to help it find a reviewer.

It adds one local input rail under `nemoguardrails/library/atr/`, modelled on `injection_detection`. Optional `pyatr` dependency, evaluated in-process, no API key, no endpoint, no network, deterministic verdicts. The blast radius is that new folder plus one optional-dependency line in `pyproject.toml`; nothing in the existing rails changes. CI is green and it is currently mergeable.

If a smaller diff would be easier to review, I can pull the docs out into a separate PR, or rebase if it has drifted. Whatever lowers the review cost.

Issue #2285, opened 2026-08-12
bug(manifests): Jailbreak rail manifest conflates in-process vs remote, model vs heuristics with conflicting dependencies


tgasser-nv:
## Did you check docs and existing issues?

- [x] I have read all the [NeMo-Guardrails documentation](https://docs.nvidia.com/nemo/guardrails/latest/index.html)
- [x] I have updated the package to the latest version before submitting this issue
- [x] (optional) I have used the [develop branch](https://github.com/NVIDIA-NeMo/Guardrails/tree/develop)
- [x] I have searched the existing issues of NeMo-Guardrails

## Python version

```
3.13.2
```

## Operating system/version

```
macOS (Darwin 25.5.0)
```

## NeMo-Guardrails version

```
0.24.0.dev0 (develop)
```

---

## Describe the bug

`nemoguardrails/library/jailbreak_detection/rail.py` declares:

```python
optional_dependencies=("scikit-learn", "torch"),
```

Three things are wrong with that list, and a fourth problem means no corrected list can be right.

**1. `scikit-learn` is declared but never imported.**
`JailbreakClassifier` loads the random forest through ONNX Runtime, not scikit-learn:

```python
# nemoguardrails/library/jailbreak_detection/model_based/models.py:56
from onnxruntime import InferenceSession
...
self.classifier = InferenceSession(random_forest_path, providers=["CPUExecutionProvider"])
```

The `sklearn-onnx` URL on line 59 is a comment describing the format the model was *exported* from.
`scikit-learn` appears nowhere in the library except this declaration, that comment, and one log message.
`onnxruntime`, which is genuinely required, is a core dependency (`pyproject.toml:27-28`) so it happens to be present regardless.

**2. `transformers` is required but was not declared.**
Both in-process paths import it, one at module scope:

- `heuristics/checks.py:19` — `from transformers import GPT2LMHeadModel, GPT2TokenizerFast`
- `model_based/models.py:27` — `from transformers import AutoModel, AutoTokenizer`

The package's own `requirements.txt` lists `transformers>=5.3.0`.

**3. The user-facing remediation message is wrong.**
`nemoguardrails/library/jailbreak_detection/actions.py:158`, logged when the in-process import fails:

> Failed to import required dependencies for local model. Install scikit-learn and torch, or use NIM-based approach

Someone following this installs a package that is not used, does not install `transformers`, and hits the same failure again.

**4. The structural problem: requirements here are conditional, and the field is not.**

[...]

PR #2293, opened 2026-08-17
refactor(rails): make shared action dependencies explicit

### LINKED ISSUE #2279: Rail Manifest required updates
## Introduction

Keeping a scratchpad of follow-ups on the new RailManifest system used for actions by both LLMRails and IORails. Will discuss these with @Pouyanpi once he's back from vacation. Here's the summary:

1. `context` is a dict, from which which actions.py can try and extract various keys. There's no information in the manifests on which keys are expected in the `context` dict and are mandatory. This applies to `relevant_chunks` for retrieval rails and `bot_message` and `_last_bot_prompt` for `self_check_hallucination()`. The arguments can't be constructed for the rail without these keys. Recommend explicitly passing strings into each action rather than unpacking context inside. Also not passing `model` through `context` but instead through `RailRequirements.models`.
2. No way to distinguish between an action requiring an API call or LLM inference.
3. Jailbreak rail manifest conflates local and NIM-based operations. Running locally needs additional heavyweight dependencies (torch, transformers, scikit-learn). Using the NIM-based one only requires an OpenAI-compatible client. Recommend splitting into two separate rails to cleanly decouple the two.

### Details for 1 (context keys)

```
┌─────────────────────┬──────────┐
│                     │ surfaces │
├─────────────────────┼──────────┤
│ user_message        │ 23       │
├─────────────────────┼──────────┤
│ bot_message         │ 20       │
├─────────────────────┼──────────┤
│ relevant_chunks     │ 6        │
├─────────────────────┼──────────┤
│ model               │ 3        │
├─────────────────────┼──────────┤
│ relevant_chunks_sep │ 1        │
├─────────────────────┼──────────┤
│ _last_bot_prompt    │ 1        │
├─────────────────────┼──────────┤
│ bot_thinking        │ 1        │
└─────────────────────┴──────────┘
```

| Direction | surface | `user_message` | `bot_message` | `relevant_chunks` | `model` | `relevant_chunks_sep` | `_last_bot_prompt` | `bot_thinking` |
|---|---|---|---|---|---|---|---|---|
| input | autoalign check input | ✓ |  |  |  |  |  |  |
| input | content safety check input | ✓ |  |  | ✓ |  |  |  |
| input | crowdstrike aidr guard input | ✓ | ✓ |  |  |  |  |  |
| input | fiddler user safety | ✓ |  |  |  |  |  |  |
| input | gcpnlp moderation | ✓ |  |  |  |  |  |  |
| input | gcpnlp moderation detailed | ✓ |  |  |  |  |  |  |
| input | guardrailsai check input | ✓ |  |  |  |  |  |  |
| input | hf classifier check input | ✓ |  |  |  |  |  |  |
| input | jailbreak detection heuristics | ✓ |  |  |  |  |  |  |
| input | jailbreak detection model | ✓ |  |  |  |  |  |  |
| input | llama guard check input | ✓ |  |  |  |  |  |  |
| input | pangea ai guard input | ✓ | ✓ |  |  |  |  |  |
| input | self check input | ✓ |  |  |  |  |  |  |
| input | topic safety check input | ✓ |  |  | ✓ |  |  |  |
| output | alignscore check facts |  | ✓ | ✓ |  |  |  |  |
| output | autoalign check output |  | ✓ |  |  |  |  |  |
| output | autoalign factcheck output | ✓ | ✓ |  |  |  |  |  |
| output | autoalign groundedness output |  | ✓ |  |  | ✓ |  |  |
| output | cleanlab trustworthiness | ✓ | ✓ |  |  |  |  |  |
| output | content safety check output | ✓ | ✓ |  | ✓ |  |  |  |
| output | crowdstrike aidr guard output | ✓ | ✓ |  |  |  |  |  |
| output | fiddler bot faithfulness |  | ✓ | ✓ |  |  |  |  |
| output | fiddler bot safety |  | ✓ |  |  |  |  |  |
| output | guardrailsai check output |  | ✓ |  |  |  |  |  |
| output | hf classifier check output |  | ✓ |  |  |  |  |  |
| output | llama guard check output | ✓ | ✓ |  |  |  |  |  |
| output | pangea ai guard output | ✓ | ✓ |  |  |  |  |  |
| output | patronus api check output | ✓ | ✓ | ✓ |  |  |  |  |
| output | patronus lynx check output hallucination | ✓ | ✓ | ✓ |  |  |  |  |
| output | self check facts |  | ✓ | ✓ |  |  |  |  |
| output | self check hallucination |  | ✓ |  |  |  | ✓ |  |
| output | self check output | ✓ | ✓ |  |  |  |  | ✓ |
| retrieval | hf classifier check retrieval |  |  | ✓ |  |  |  |  |
| | **surfaces** | 23 | 20 | 6 | 3 | 1 | 1 | 1 |

Pouyanpi:
## Description

Make the inputs and model dependencies of shared library actions explicit in rail manifests, then have IORails compile those declarations instead of maintaining rail-specific knowledge. HTTP remains generic runtime infrastructure: `RailsManager` owns one pooled client, and signature filtering injects it only into actions that accept `http_client`.

This is intentionally one PR rather than a stack. Its four implementation commits are self-contained review units, and the sequence below is the recommended review order.

### Review by commit

1. `refactor(library): declare action context inputs`
   Declares user, bot, and retrieval inputs across shared library actions. Explicit parameters serve manifest-driven runtimes while the existing `context` path remains as an LLMRails fallback.
2. `feat(manifests): declare model resource bindings`
   Marks fixed and configurable model bindings explicitly so compilation validates model requirements without inferring them from a parameter named `model_name`.
3. `refactor(iorails): share managed HTTP client`
   Removes per-surface HTTP classification and per-rail clients. `RailsManager` owns and closes one pooled client shared by compiled rails; manifests do not gain a `requires_http_client` allocation flag.
4. `refactor(iorails): consume manifest-bound action inputs`
   Replaces hard-coded retrieval capability lists and whole-context injection. IORails derives required inputs from each surface and passes only declared bindings plus accepted runtime dependencies.

### LLMRails compatibility

LLMRails does not consume the new execution bindings, and its dispatch path is unchanged. Updated actions accept explicit inputs for IORails but continue to fall back to the existing `context` argument used by LLMRails. Model resource markers are manifest metadata only. LLMRails can also continue registering an `http_client`, or use the existing per-call fallback when none is registered.

Cross-engine tests execute the same model-backed and HTTP-backed actions through both engines and assert equivalent results. This addresses the non-jailbreak portions of #2279 without requiring an LLMRails managed-client lifecycle.

### Out of scope

The local and NIM jailbreak backends still need separate requirements and execution contracts. That work remains tracked in #2285.

## Related Issue(s)

- Addresses #2279

## AI Assistance

- [ ] No AI tools were used.
- [x] AI tools were used; a human reviewed and can explain every change (tool: Codex).

## Checklist

- [x] I've added tests where applicable.

## Summary by CodeRabbit

- **New Features**
  - Guardrails now pass user, bot, and retrieval content explicitly to safety and validation checks.
  - Added clearer model-binding support for configurable guardrail models.
  - Expanded support for message and evidence inputs across built-in integrations.

- **Bug Fixes**
  - Improved handling of missing optional context values.
  - Required unsupported inputs are now identified earlier with clearer validation feedback.
  - API-backed guardrails now share a single connection pool for more consistent resource management.

### DIFF
--- a/nemoguardrails/guardrails/compiled_rail.py
+++ b/nemoguardrails/guardrails/compiled_rail.py
@@ -190,6 +190,18 @@ class CompiledRail:
         self._deps = deps
         self._accepted = accepted
 
+    def with_runtime_dependencies(self, deps: RailDependencies) -> "CompiledRail":
+        """Return the same execution plan with its runtime collaborators finalized."""
+        return CompiledRail(
+            flow=self.flow,
+            surface=self.surface,
+            action=self._action,
+            bound=self._bound,
+            context_bound=self._context_bound,
+            deps=deps,
+            accepted=self._accepted,
+        )
+
     @property
     def surface_name(self) -> str:
         """The manifest surface name, without any ``$param=`` suffix."""
--- a/nemoguardrails/guardrails/compiled_rail.py
+++ b/nemoguardrails/guardrails/compiled_rail.py
@@ -48,6 +48,7 @@ from nemoguardrails.manifests import (
 if TYPE_CHECKING:
     from opentelemetry.trace import Tracer
 
+    from nemoguardrails.http import HTTPClient
     from nemoguardrails.logging.explain import LLMCallInfo
     from nemoguardrails.manifests import RailCatalog
 
@@ -75,6 +76,7 @@ class RailDependencies:
     llm_task_manager: Any
     config: Any
     model_caches: Optional[Mapping[str, Any]] = None
+    http_client: Optional["HTTPClient"] = None
     tracer: Optional["Tracer"] = None
 
 
@@ -175,7 +177,6 @@ class CompiledRail:
         context_bound: tuple[_ContextParameter, ...],
         deps: RailDependencies,
         accepted: frozenset[str],
-        http_client: Any = None,
     ) -> None:
         """Store the frozen execution plan. Build through :func:`compile_rail`.
 
@@ -189,7 +190,6 @@ class CompiledRail:
         self._context_bound = context_bound
         self._deps = deps
         self._accepted = accepted
-        self._http_client = http_client
 
     @property
     def surface_name(self) -> str:
@@ -257,21 +257,11 @@ class CompiledRail:
             "llm": self._deps.llms.get("main"),
             "llm_task_manager": self._deps.llm_task_manager,
             "config": self._deps.config,
-            "http_client": self._http_client,
+            "http_client": self._deps.http_client,
             "model_caches": self._deps.model_caches,
             "context": context,
             "events": messages_to_events(messages),
         }
-
-    async def close(self) -> None:
-        """Close the rail's HTTP client, if it owns one.
-
-        Failures propagate: releasing one client is the whole job here, and the caller
-        closing a whole set of rails is the only layer that can decide a leak is
-        survivable. Repeat calls are safe, as a closed client's ``close()`` is a no-op.
-        """
-        if self._http_client is not None:
-            await self._http_client.close()
 
 
 def _accepted_parameters(action: Callable[..., Any]) -> frozenset[str]:
@@ -595,7 +585,6 @@ def compile_rail(
     direction: RailDirection,
     deps: RailDependencies,
     *,
-    http_client: Any = None,
     catalog: Optional["RailCatalog"] = None,
 ) -> CompiledRail:
     """Compile one configured flow string into an executable rail.
@@ -641,5 +630,4 @@ def compile_rail(
         context_bound=context_bound,
         deps=deps,
         accepted=accepted,
-        http_client=http_client,
     )
--- a/nemoguardrails/guardrails/compiled_rail.py
+++ b/nemoguardrails/guardrails/compiled_rail.py
@@ -333,13 +333,6 @@ def _context_parameters(surface: RailSurface, flow: str) -> tuple[_ContextParame
     same as injecting the whole ``context`` dict: ``user_message`` reaches an action that
     calls the parameter ``text`` or ``user_prompt``.
     """
-    unsupported = _unsupported_context_keys(surface)
-    if unsupported:
-        raise RailCompilationError(
-            f"{flow!r} needs context variable(s) {', '.join(repr(key) for key in unsupported)}, "
-            f"which IORails does not supply; it has {list(_CONTEXT_KEYS)}"
-        )
-
     bound: list[_ContextParameter] = []
     for binding in surface.bindings:
         if binding.kind != "context":
--- a/nemoguardrails/guardrails/compiled_rail.py
+++ b/nemoguardrails/guardrails/compiled_rail.py
@@ -99,7 +99,7 @@ _SYSTEM_MESSAGE_EVENT = "SystemMessage"
 def _history_before_current_turn(messages: LLMMessages) -> LLMMessages:
     """The turns preceding the one being checked.
 
-    Actions append the checked turn themselves, from ``context["user_message"]``, and always
+    Actions append the checked turn themselves, from their bound ``user_message``, and always
     at the end. So the history stops short of it: emitting it here would hand the model the
     same turn twice, and emitting what follows it — an assistant reply in a ``check()``
     transcript, say — would place a later turn ahead of it and reorder the conversation.
@@ -154,8 +154,7 @@ class _ContextParameter:
     key: str
 
 
-# The conversation variables a context binding may name. Also the keys of the ``context``
-# dict injected wholesale into actions that declare it, so the two cannot drift.
+# The conversation variables IORails can supply to explicit context bindings.
 _CONTEXT_KEYS = ("user_message", "bot_message")
 
 
@@ -239,18 +238,14 @@ class CompiledRail:
     def _call_kwargs(self, messages: LLMMessages, bot_response: Optional[str]) -> dict[str, Any]:
         """Assemble the action's arguments from its declared parameters and the manifest."""
         context = _request_context(messages, bot_response)
-        kwargs = {
-            name: value
-            for name, value in self._request_dependencies(messages, context).items()
-            if name in self._accepted
-        }
+        kwargs = {name: value for name, value in self._request_dependencies(messages).items() if name in self._accepted}
         for bound in self._bound:
             kwargs[bound.action_param] = bound.value
         for context_bound in self._context_bound:
             kwargs[context_bound.action_param] = context[context_bound.key]
         return kwargs
 
-    def _request_dependencies(self, messages: LLMMessages, context: Mapping[str, str]) -> dict[str, Any]:
+    def _request_dependencies(self, messages: LLMMessages) -> dict[str, Any]:
         """Every value injectable by parameter name; the caller filters against the signature."""
         return {
             "llms": self._deps.llms,
@@ -259,7 +254,6 @@ class CompiledRail:
             "config": self._deps.config,
             "http_client": self._deps.http_client,
             "model_caches": self._deps.model_caches,
-            "context": context,
             "events": messages_to_events(messages),
         }
 
@@ -339,18 +333,20 @@ def _context_parameters(surface: RailSurface, flow: str) -> tuple[_ContextParame
     same as injecting the whole ``context`` dict: ``user_message`` reaches an action that
     calls the parameter ``text`` or ``user_prompt``.
     """
+    unsupported = _unsupported_context_keys(surface)
+    if unsupported:
+        raise RailCompilationError(
+            f"{flow!r} needs context variable(s) {', '.join(repr(key) for key in unsupported)}, "
+            f"which IORails does not supply; it has {list(_CONTEXT_KEYS)}"
+        )
+
     bound: list[_ContextParameter] = []
     for binding in surface.bindings:
         if binding.kind != "context":
             continue
         key = _binding_source_key(binding, flow)
-        if key not in _CONTEXT_KEYS:
-            # Refused rather than passed an empty value: the rail would otherwise return a
-            # verdict computed over evidence IORails never supplied.
-            raise RailCompilationError(
-                f"{flow!r} binds {binding.action_param!r} to context variable {key!r}, "
-                f"which IORails does not supply; it has {list(_CONTEXT_KEYS)}"
-            )
+        if key not in _CONTEXT_KEYS and not binding.required:
+            continue
         bound.append(_ContextParameter(binding.action_param, key))
     return tuple(bound)
 
@@ -375,27 +371,25 @@ def _unapplicable_transform_reason(surface: RailSurface) -> Optional[str]:
     return f"rewrites {surface.transform_target.value!r}, which a {surface.direction.value} rail cannot apply here"
 
 
-# Surfaces whose actions read retrieval evidence out of the request context: ``relevant_chunks``,
-# ``relevant_chunks_sep``, or the Colang-internal ``_last_bot_prompt``. Keyed by direction as well
-# as name, because one rail can surface in both directions.
-_RETRIEVAL_CONTEXT_SURFACES: frozenset[tuple[RailDirection, str]] = frozenset(
-    {
-        (RailDirection.OUTPUT, "alignscore check facts"),
-        (RailDirection.OUTPUT, "autoalign groundedness output"),
-        (RailDirection.OUTPUT, "fiddler bot faithfulness"),
-        (RailDirection.OUTPUT, "patronus api check output"),
-        (RailDirection.OUTPUT, "patronus lynx check output hallucination"),
-        (RailDirection.OUTPUT, "self check facts"),
-        (RailDirection.OUTPUT, "self check hallucination"),
+def _unsupported_context_keys(surface: RailSurface) -> tuple[str, ...]:
+    """Return declared context variables that IORails cannot construct."""
+    keys = {
+        binding.key
+        for binding in surface.bindings
+        if binding.kind == "context"
+        and binding.required
+        and binding.key is not None
+        and binding.key not in _CONTEXT_KEYS
     }
-)
+    return tuple(sorted(keys))
 
 
-def _retrieval_context_reason(surface: RailSurface) -> Optional[str]:
-    """Report a surface needing retrieval evidence IORails has no source for."""
-    if (surface.direction, surface.name) not in _RETRIEVAL_CONTEXT_SURFACES:
+def _unsupported_context_reason(surface: RailSurface) -> Optional[str]:
+    """Report context requirements outside IORails' request contract."""
+    keys = _unsupported_context_keys(surface)
+    if not keys:
         return None
-    return "needs retrieval evidence, which manifest-driven execution does not supply yet"
+    return f"needs context variable(s) {', '.join(repr(key) for key in keys)}, which IORails does not supply"
 
 
 def _unsupported_rail_reason(surface: RailSurface) -> Optional[str]:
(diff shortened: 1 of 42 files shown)


Pouyanpi:
@greptileai review

nemoguardrails/library/factchecking/align_score/rail.py
```diff
@@ -50,7 +51,15 @@
...
         surfaces=(
-            RailSurface(name="alignscore check facts", direction=RailDirection.OUTPUT, action=ALIGNSCORE_CHECK_FACTS),
+            RailSurface(
+                name="alignscore check facts",
+                direction=RailDirection.OUTPUT,
+                action=ALIGNSCORE_CHECK_FACTS,
+                bindings=(
+                    Binding.context("relevant_chunks", "relevant_chunks"),
+                    Binding.context("bot_message", "bot_message"),
+                ),
```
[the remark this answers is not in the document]
Pouyanpi: this is resolved in the latest commit. Do you confirm?

nemoguardrails/library/jailbreak_detection/actions.py
```diff
@@ -68,7 +69,8 @@ async def jailbreak_detection_heuristics(
     lp_threshold = jailbreak_config.length_per_perplexity_threshold
     ps_ppl_threshold = jailbreak_config.prefix_suffix_perplexity_threshold
 
-    prompt = context.get("user_message")
+    context = context or {}
+    prompt = user_message if user_message is not None else context.get("user_message")
```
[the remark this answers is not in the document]
Pouyanpi: this is resolved in the latest commit. Do you confirm?

nemoguardrails/manifests/manifest.py
```diff
@@ -223,6 +227,22 @@ def literal(cls, action_param: str, value: Any) -> "Binding":
...
+            action_param=action_param,
+            key=name,
+            required=required,
+            resource="model",
+        )
+
+    @classmethod
+    def model(cls, action_param: str, model_type: str) -> "Binding":
+        """Bind an action parameter to a fixed model type."""
+        return cls(kind="literal", action_param=action_param, value=model_type, resource="model")
```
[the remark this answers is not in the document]
Pouyanpi: this is resolved in the latest commit. Do you confirm?

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -311,6 +234,7 @@ def __init__(
         self.tool_call_flows: list[str] = list(tool_call_flows or [])
         self.tool_result_flows: list[str] = list(tool_result_flows or [])
 
+        self._http_client = create_http_client()
```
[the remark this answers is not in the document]
Pouyanpi: @greptileai this is resolved in the latest commit. Do you confirm?

nemoguardrails/library/factchecking/align_score/rail.py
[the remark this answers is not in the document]
Pouyanpi: I forgot to push, just pushed now. what do you think?

nemoguardrails/library/jailbreak_detection/actions.py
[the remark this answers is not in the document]
Pouyanpi: I forgot to push, just pushed now. what do you think?

nemoguardrails/manifests/manifest.py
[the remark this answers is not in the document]
Pouyanpi: I forgot to push, just pushed now. what do you think?

nemoguardrails/guardrails/rails_manager.py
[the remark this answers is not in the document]
Pouyanpi: @greptileai I forgot to push, just pushed now. what do you think?

nemoguardrails/library/factchecking/align_score/rail.py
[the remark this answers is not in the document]
Pouyanpi: what are you looking at? have you checked the latest commits?

nemoguardrails/library/jailbreak_detection/actions.py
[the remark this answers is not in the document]
Pouyanpi: how about now?

nemoguardrails/library/factchecking/align_score/rail.py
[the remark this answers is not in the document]
Pouyanpi: next time when the commit shas haven't changed don't review the code

nemoguardrails/guardrails/rails_manager.py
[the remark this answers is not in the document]
Pouyanpi: @greptileai you reviewed sooner. how about now?

tests/rails/llm/test_builtin_rail_manifests.py
```diff
@@ -175,20 +175,118 @@ def _is_action_decorator(decorator: ast.expr) -> bool:
...
+    }
+
+    assert declared == expected
+
+
+def test_retrieval_dependent_surfaces_declare_context_inputs():
+    surfaces = default_rail_catalog().surfaces()
+    expected = {
+        (RailDirection.OUTPUT, "alignscore check facts"): (
+            Binding.context("relevant_chunks", "relevant_chunks", required=False),
```
christinaexyou: should a user be allowed to enable `self_check_hallucination`, `alignscore_check_facts`, and `call_fiddler_faithfulness` ? iiuc these rails are no-op are no-op bc `relevant_chunks` and `_last_bot_prompt` are not keys in `_CONTEXT_KEYS` so they would not produce a meaningful verdict/fallback

christinaexyou: i would either add them to the unsupported set or surface a warning/error

christinaexyou:
thanks @Pouyanpi - there's just one comment that needs addressing but o.w lgtm !

nemoguardrails/library/hallucination/actions.py
```diff
@@ -46,14 +46,17 @@ async def self_check_hallucination(
...
 ) -> RailOutcome:
     """Checks if the last bot response is a hallucination by checking multiple completions for self-consistency.
 
     :return: A blocking outcome if hallucination is detected, otherwise an allowing outcome.
     """
-    bot_response = context.get("bot_message")
-    last_bot_prompt_string = context.get("_last_bot_prompt")
+    context = context or {}
+    bot_response = bot_message if bot_message is not None else context.get("bot_message")
+    last_bot_prompt_string = last_bot_prompt if last_bot_prompt is not None else context.get("_last_bot_prompt")
```
tgasser-nv, line 59: There's no way for IORails to add a `_last_bot_prompt` key to the context dict without adding some custom logic to handle this. Colang runtimes get this for free. Recommend adding this rail to a blocklist in `rails_manager.py` and routing through to IORails' `can_handle()` so this gets run on LLMRails

nemoguardrails/guardrails/compiled_rail.py
```diff
@@ -383,27 +371,25 @@ def _unapplicable_transform_reason(surface: RailSurface) -> Optional[str]:
     return f"rewrites {surface.transform_target.value!r}, which a {surface.direction.value} rail cannot apply here"
 
 
-# Surfaces whose actions read retrieval evidence out of the request context: ``relevant_chunks``,
-# ``relevant_chunks_sep``, or the Colang-internal ``_last_bot_prompt``. Keyed by direction as well
-# as name, because one rail can surface in both directions.
-_RETRIEVAL_CONTEXT_SURFACES: frozenset[tuple[RailDirection, str]] = frozenset(
```
tgasser-nv, line 389: I don't understand how these retrieval rails will ever work with IORails. They all need one or more of `relevant_chunks`, `relevant_chunks_sep`, `_last_bot_prompt` or `bot_thinking`. These are tightly coupled to the Colang engine.

The 4th commit in the series adds these are `required=False` which then falls back to the `None`-guards in the action. For example `self check hallucination` will never block when called through IORails.

nemoguardrails/library/hallucination/actions.py
```diff
@@ -46,14 +46,17 @@ async def self_check_hallucination(
     context: Optional[dict] = None,
     use_llm_checking: bool = True,
     config: Optional[RailsConfig] = None,
+    bot_message: Optional[str] = None,
+    last_bot_prompt: Optional[str] = None,
```
tgasser-nv, line 50: When run under IORails, this rail will never block. Tracing it through: 

1. `last_bot_prompt` is None, since this is a Colang-specific event.
2. `context` won't contain `last_bot_prompt` key since IOrails doesn't maintain a Colang `context`.
3. So line 61 won't be taken.
4. Drops through to line 127 and unconditionally returns `_hallucination_outcome(False)` which converts to a RailOutcome which always allows

nemoguardrails/library/factchecking/align_score/rail.py
```diff
@@ -50,7 +51,15 @@
         flows=RailFlows(flow_names=("alignscore check facts",)),
         actions=RailActions(refs=(ALIGNSCORE_CHECK_FACTS, ALIGNSCORE_REQUEST)),
         surfaces=(
-            RailSurface(name="alignscore check facts", direction=RailDirection.OUTPUT, action=ALIGNSCORE_CHECK_FACTS),
+            RailSurface(
+                name="alignscore check facts",
+                direction=RailDirection.OUTPUT,
+                action=ALIGNSCORE_CHECK_FACTS,
+                bindings=(
+                    Binding.context("relevant_chunks", "relevant_chunks", required=False),
```
tgasser-nv: This binding **is** required. The rail has nothing to check the bot_message against if the `relevant_chunks` aren't provided. It returns 1.0 without calling the service.

nemoguardrails/library/fiddler/rail.py
```diff
@@ -61,9 +62,27 @@
...
+                action=FIDDLER_BOT,
+                bindings=(Binding.context("bot_message", "bot_message"),),
+            ),
+            RailSurface(
+                name="fiddler bot faithfulness",
+                direction=RailDirection.OUTPUT,
+                action=FIDDLER_FAITHFULNESS,
+                bindings=(
+                    Binding.context("bot_message", "bot_message"),
+                    Binding.context("relevant_chunks", "relevant_chunks", required=False),
```
tgasser-nv: These relevant chunks are required by the rail (required=True). There's no guard in the actions.py to check before POSTing to the service

nemoguardrails/library/hallucination/rail.py
```diff
@@ -47,7 +48,13 @@
         actions=RailActions(refs=(SELF_CHECK_HALLUCINATION,)),
         surfaces=(
             RailSurface(
-                name="self check hallucination", direction=RailDirection.OUTPUT, action=SELF_CHECK_HALLUCINATION
+                name="self check hallucination",
+                direction=RailDirection.OUTPUT,
+                action=SELF_CHECK_HALLUCINATION,
+                bindings=(
+                    Binding.context("bot_message", "bot_message"),
+                    Binding.context("last_bot_prompt", "_last_bot_prompt", required=False),
```
tgasser-nv: This binding is required. If it isn't provided the rail always allows flags it as not a halucination in IOrails but works correctly in LLMRails.

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -342,6 +258,11 @@ def __init__(
         self._tool_call_actions = self._build_tool_actions(self.tool_call_flows, ToolCallRailAction)
         self._tool_result_actions = self._build_tool_actions(self.tool_result_flows, ToolResultRailAction)
 
+        self._http_client = create_http_client()
```
tgasser-nv, line 261: This change moves from rail-specific to a general HTTP client used by all rails and mutates dependencies after rails are compiled. Recommend creating the single HTTP client and passing this into `compile_rail()` rather than mutating after the fact.

The single HTTP Client is shared between every vendor and has default settings of [100 max_connections and 20 max_keepalive_connections](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/nemoguardrails/http/transport.py#L87). If we're using multiple vendor rails, they could potentially starve others from making requests if they consume the pool. Not blocking for this PR, but we need an HTTP client per-vendor and limits on the max_connections and max_keepalive_connections for each.

nemoguardrails/guardrails/compiled_rail.py
```diff
@@ -347,18 +333,20 @@ def _context_parameters(surface: RailSurface, flow: str) -> tuple[_ContextParame
     same as injecting the whole ``context`` dict: ``user_message`` reaches an action that
     calls the parameter ``text`` or ``user_prompt``.
     """
+    unsupported = _unsupported_context_keys(surface)
```
tgasser-nv: This duplicates code in `compile_rail()` at line 593-4

tgasser-nv (commented):
Thanks for adding this, it looks good overall. Just a few comments:

### 1. Unsupportable IORails rails

The following rails won't work correctly in IORails as-is. They rely on `binding.Context`s which are listed as optional (required=False) when they're actually mandatory. IORails doesn't provide these bindings since they're Colang-specific.

1. library/factchecking/align_score/rail.py
2. library/fiddler/rail.py
3. library/hallucination/rail.py

These need to added to the manual blocklist so we fall back to LLMRails and execute these rails correctly.

### 2. Patch coverage

Could you add some more tests to get the 95.92% coverage up to 100%?

nemoguardrails/guardrails/rails_manager.py
Pouyanpi: I addressed the private dependency mutation in 386620ecb8, but slightly differently from creating the client before compilation.

creating it first would regress the existing lifecycle guarantee: if rail or tool validation fails during synchronous construction, we would have allocated an asynchronously closed resource with no cleanup path. the manager now completes validation, creates the shared client, and replaces each validated `CompiledRail` with a finailzed copy carrying the runtime deos. there is no post-compilation mutation.

I did not add per-vendor pools in this PR. the starvation concern is valid as a future resource-isolation consideration: the current client separates connections by origin, but its limits are aggregate. per vendor pools need an explicit service identity and resource configuration contract, together with evidence for appropriate limits, so I think that should be designed and measured separately.

nemoguardrails/guardrails/compiled_rail.py
Pouyanpi: fixed in 6bfae8f951. removed the duplicate check from context-parameter planning so `compile_rail()` now relies on `unsupported_surface_reason()`.

nemoguardrails/library/hallucination/actions.py
Pouyanpi: fixed in 714a1f816c. `_last_bot_prompt` is now required, so the generic capability gate falls back to LLMRails when IORails cannot supply it. I used the manifest requirement instead of adding a rail-name blocklist.

nemoguardrails/guardrails/compiled_rail.py
Pouyanpi: agreed. those three bindings were incorrectly marked optional. Fixed in 714a1f816c; unsupported required context now makes `IORails.can_handle()` fall back to LLMRails through the manifest capability gate.

nemoguardrails/library/hallucination/actions.py
Pouyanpi: Correct 👍🏻  fixed in 714a1f816c by making the originating prompt required.

nemoguardrails/library/fiddler/rail.py
Pouyanpi: You're right, I agree. It's fixed in 714a1f816c and `relevant_chunks` is now a required manifest binding, so IORails cannot run AlignScore without evidence.

nemoguardrails/library/hallucination/rail.py
Pouyanpi: Sure it is fixed in 714a1f816c

tests/rails/llm/test_builtin_rail_manifests.py
Pouyanpi: Good catch Christina 👍🏻  it's fiixed in 714a1f816c: those evidence and prompt bindings are now required. I used the declared requirements rather than adding a separate unsupported-rail list.

tgasser-nv (approved):
Looks good, thanks for the updates

PR #2300, opened 2026-08-18
feat(llm): add public api_key property for in-place credential rotation

### LINKED ISSUE #2299: feature: public `api_key` property for in-place LLM credential rotation
### Did you check the docs?

Applications that build one long-lived `LLMRails`/`OpenAIChatModel` instance at process startup (to avoid per-request construction overhead) and authenticate against their LLM provider with a short-lived, rotating bearer token (OAuth client-credentials, AWS STS-style tokens, internal gateway tokens, etc.) have no supported way to update that credential in place.

Today the only two options are:
- Reach into private attributes two levels deep (`rails.llm._client._api_key`), which is unsupported and can silently break on any minor/patch upgrade with no type or import error — just silent 401s from the provider.
- Rebuild the model via `DefaultFramework.create_model()` and swap it in with `LLMRails.update_llm()`. This looks like the supported path, but `DefaultFramework._clients` caches `OpenAICompatibleClient` instances keyed by `(base_url, api_key, ...)` and never evicts entries, so every token rotation leaks a new `httpx.AsyncClient` connection pool for the life of the process.

Short-lived, periodically-rotated bearer tokens are a common integration pattern for enterprise LLM gateways, not something specific to one deployment, so this gap affects any application holding a long-lived rails/model instance.

Add a small, purely additive `api_key` property (with setter) so callers have
a public, documented way to rotate the credential on an already-built model:

- `BaseClient.api_key` (`nemoguardrails/llm/clients/base.py`): getter/setter over the existing `_api_key` attribute. Safe to call between requests, since `_build_headers()` already re-reads `_api_key` fresh on every call.
- `OpenAIChatModel.api_key` (`nemoguardrails/llm/models/openai_chat.py`): delegates to the underlying client's `api_key`.
- `InstrumentedLLMModel.api_key` (`nemoguardrails/llm/models/instrumented.py`): `LLMRails` wraps the model in this decorator whenever tracing or metrics are enabled, so the tracing/metrics wrapper needs to forward `api_key` too, otherwise assignment silently creates a stray attribute on the decorator and never reaches the real client.
- Document on the `LLMModel` protocol (`nemoguardrails/types.py`) that `api_key` is a bearer-token-specific extension, not a required protocol member, so callers know to `hasattr`-check before assigning.

With this, credential rotation becomes:

```python
if hasattr(rails.llm, "api_key"):
    rails.llm.api_key = new_token
```

No existing method signature, default, or behavior changes; this only adds new
public surface.

- Keep using the private-attribute workaround (`rails.llm._client._api_key`) - rejected, unsupported and can silently break across versions.
- Rebuild the model and call `LLMRails.update_llm()` on every rotation - rejected, leaks an `OpenAICompatibleClient` + `httpx.AsyncClient` connection pool per rotation via `DefaultFramework._clients`, since that cache is keyed by `api_key` and entries are never evicted.
- Bound/evict `DefaultFramework._clients` instead of adding a setter - would fix the leak but not the missing-public-accessor gap, and is a larger, separate change to shared framework caching behavior.

### Additional context

- A draft implementation (property + setter across `BaseClient`, `OpenAIChatModel`, `InstrumentedLLMModel`, plus the `LLMModel` docstring note) and unit tests already exist on a [local branch](https://github.com/waitasecant/Guardrails/tree/feat/llm-client), ready to open as a PR once this issue is triaged and assigned.
- The `DefaultFramework._clients` unbounded-cache issue is a related but distinct bug; flagging it here for awareness but proposing it as a separate follow-up rather than folding it into this change.

waitasecant:
## Description

Applications that build one long-lived `LLMRails`/`OpenAIChatModel` instance at process startup (a common pattern for services that don't want per-request construction overhead) and authenticate against their LLM provider with a short-lived, rotating bearer token (OAuth client-credentials, AWS STS-style tokens, internal gateway tokens, etc.) currently have no supported way to update that credential in place.

Today the only two options are:
- Reach into private attributes two levels deep (`rails.llm._client._api_key`), which is unsupported and can silently break on any minor/patch upgrade with no type or import error — just silent 401s from the provider.
- Rebuild the model via `DefaultFramework.create_model()` and swap it in with `LLMRails.update_llm()`. This "looks" like the supported path, but `DefaultFramework._clients` caches `OpenAICompatibleClient` instances keyed by `(base_url, api_key, ...)` and never evicts entries, so every token rotation leaks a new `httpx.AsyncClient` connection pool for the life of the process.

This PR adds a small, purely additive `api_key` property (with setter) so callers have a public, documented way to rotate the credential on an already-built model without either workaround.

## Related Issue(s)

- Fixes #2299 

## Verification

Ran pytest and pre-commit checks successfully.

## AI Assistance

- [ ] No AI tools were used.
- [x] AI tools were used; a human reviewed and can explain every change (tool: GitHub Copilot / Claude).

## Checklist

- [x] This PR links to a triaged issue assigned to me.
- [x] I've noted any verification beyond CI and any checks I couldn't run.

### DIFF
--- a/nemoguardrails/llm/clients/base.py
+++ b/nemoguardrails/llm/clients/base.py
@@ -137,6 +137,23 @@ class BaseClient:
     def provider_url(self) -> Optional[str]:
         return None
 
+    @property
+    def api_key(self) -> Optional[str]:
+        """The bearer token/API key used for the Authorization header."""
+        return self._api_key
+
+    @api_key.setter
+    def api_key(self, value: Optional[str]) -> None:
+        """Update the API key in place.
+
+        Safe to call between requests on an in-flight client: headers are
+        rebuilt fresh per call in ``_build_headers()``, so this only affects
+        requests started after the assignment. Intended for callers that hold
+        one long-lived client/model across many short-lived credential
+        rotations (e.g. OAuth client-credentials tokens).
+        """
+        self._api_key = value
+
     def _error_context(self) -> ErrorContext:
         return ErrorContext(
             model_name=None,
--- a/nemoguardrails/llm/models/instrumented.py
+++ b/nemoguardrails/llm/models/instrumented.py
@@ -94,6 +94,20 @@ class InstrumentedLLMModel:
         return self._model.provider_url
 
     @property
+    def api_key(self) -> Optional[str]:
+        """The wrapped model's bearer token/API key, if it exposes one.
+
+        Raises ``AttributeError`` for wrapped models without one, so
+        ``hasattr(rails.llm, "api_key")`` still reports support correctly
+        through this decorator.
+        """
+        return getattr(self._model, "api_key")
+
+    @api_key.setter
+    def api_key(self, value: Optional[str]) -> None:
+        setattr(self._model, "api_key", value)
+
+    @property
     def wrapped_model(self) -> LLMModel:
         """Return the underlying model for direct access or re-instrumentation."""
         return self._model
--- a/nemoguardrails/llm/models/openai_chat.py
+++ b/nemoguardrails/llm/models/openai_chat.py
@@ -73,6 +73,15 @@ class OpenAIChatModel:
     def provider_url(self) -> Optional[str]:
         return self._client.provider_url
 
+    @property
+    def api_key(self) -> Optional[str]:
+        """The bearer token/API key used by the underlying HTTP client."""
+        return self._client.api_key
+
+    @api_key.setter
+    def api_key(self, value: Optional[str]) -> None:
+        self._client.api_key = value
+
     def _enrich(self, exc: LLMClientError) -> LLMClientError:
         exc.provider_name = self._provider_name
         exc.model_name = self._model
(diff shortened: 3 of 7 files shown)


waitasecant:
Hi @Pouyanpi @tgasser-nv @miyoungc
 
could you look into this PR put out an update
 
Friendly ping on this PR 🙂
 
This is a small, additive change that exposes a public `api_key` property to support in-place credential rotation without relying on private attributes or recreating/caching new clients on every token refresh.
 
The PR has been open for about 3 weeks and appears to still be awaiting triage/review. When you have a chance, could you please take a look and/or approve the pending workflows so CI can run?
 
Thanks for your time!

waitasecant:
Hi @Pouyanpi @tgasser-nv @miyoungc
 
Following up on this PR. It is currently still marked needs triage, has no assigned reviewer, and the required checks appear to be waiting for approval/status reporting.
 
Could someone please:
 
Confirm whether the proposed API direction is acceptable?
Assign a reviewer (if different)?
Approve/unblock the pending workflows so CI can run?
 
If there are concerns with exposing api_key as a public property, I'm happy to revise the implementation accordingly.
 
Thanks.

PR #2312, opened 2026-08-19
refactor(rails): rename tool rail fields

Pouyanpi:
## Description

Rename rails.`tool_output` and rails.`tool_input` to rails.`tool_call` and rails.`tool_result`, retaining deprecated compatibility aliases.

Update runtime terminology, telemetry, tests, and documentation.

## AI Assistance

- [ ] No AI tools were used.
- [ ] AI tools were used; a human reviewed and can explain every change (tool: ___).

## Checklist

- [ ] I've read the [CONTRIBUTING](https://github.com/NVIDIA-NeMo/Guardrails/blob/develop/CONTRIBUTING.md) guidelines.
- [ ] This PR links to a triaged issue assigned to me.
- [ ] My PR title follows the project commit convention.
- [ ] I've updated the documentation if applicable.
- [ ] I've added tests if applicable.
- [ ] I've noted any verification beyond CI and any checks I couldn't run.
- [ ] I did not update generated changelog files manually.
- [ ] I addressed all CodeRabbit, Greptile, and other review comments, or replied with why no change is needed.
- [ ] @mentions of the person or team responsible for reviewing proposed changes.

## Summary by CodeRabbit

* **New Features**
  * Added canonical `tool_call` and `tool_result` rail configuration for tool-call and tool-result validation.
  * Added support for configuring validation flows and parallel processing for both rail types.
* **Bug Fixes**
  * Improved rail processing, streaming events, logging, and telemetry to consistently report tool calls and results.
* **Documentation**
  * Updated configuration, guardrail, and telemetry references with the new terminology.
  * Deprecated names remain supported with migration guidance and warnings.

PR #2377, opened 2026-09-10
feat: add per-tool rail scoping for tool calls and results

tanushriya910:
## Description

Adds per-tool rail scoping for the IORails engine. A guardrail check can now be scoped to a specific tool's arguments (`rails.tool_output.per_tool`) or result content (`rails.tool_input.per_tool`), keyed by tool name.
This PR includes Phase 1 that wires the regex rail through this path as the first rail using it. Intentionally done to keep the PR size reviewable and get early alignment on the per-tool rail architecture before extending it further.

### New Updates based on the feedback:
1. ToolCall and ToolResult dataclasses are used to pass all the relevant tool context to the rail actions and the actions is responsible for extracting the relevant information from the structured data.
2. As a consequence, the rail action itself performs the filtering/scoping of the tool call arguments using $argument passed as the surface param now
3. Renaming the newly introduced per tool rail surface names and rail direction to follow tool_input/tool_output naming to be consistent
4. Added @tool_output_validation, which validates a tool call's arguments against the tool's declared JSON schema (and blocks calls to undeclared tools) before the per-tool rail action runs, enforced catalog-wide via a new characterization test.
5. Per-tool flows are now validated against the manifest catalog inside IORails.unsupported_reason, so a nonexistent/misdirected per-tool flow degrades gracefully instead of crashing IORails.__init__ with a RailCompilationError. since LLMRails never reads per-tool config at all, Guardrails now raises instead of silently falling back to LLMRails when per-tool rails are configured but IORails can't handle them

### What Changed

The dispatch mechanism reuses IORails' existing manifest-driven `CompiledRail` pipeline rather than adding a second, parallel one. Two new manifest directions (`RailDirection.TOOL_CALL`/`TOOL_RESULT`) and two regex surfaces (`"regex check tool call"`/`"regex check tool result"`) compile through the same `RailManifest`/`compile_rail` path as every other built-in rail.

Getting per-tool-call data (the tool name, and the call's arguments or result content) into a compiled rail's action without changing the signature of `CompiledRail.execute()` or any of the call chain threading through it (`_invoke`, `_call_kwargs`, and every existing call site) required one new piece of shared plumbing: a `tool_context_var` contextvar, set fresh around each `execute()` call in the fan-out loop and merged into the same `_request_context()` dict that already supplies `user_message`/`bot_message` to every other rail.

`RailsManager.are_tool_calls_safe`/`are_tool_results_safe` fan out across every tool call/result in a request, one action invocation per (tool call, configured flow) pair that blocks on the first violation and merges generation-log records from both the pre-existing global check and the new per-tool checks. `$argument=<name>` on a tool-call flow scopes the serialized context to one named argument instead of the full arguments dict, so a check can inspect a user-supplied field without seeing unrelated call metadata.

### Next Steps

This PR (Phase 1) intentionally keeps scope narrow to stay reviewable. Deferred/out-of-scope items, tracked for follow-up:

1. **Naming: `tool_output`/`tool_input` vs `tool_call`/`tool_result`.** The new manifest directions (`RailDirection.TOOL_CALL`/`TOOL_RESULT`) and surface names (`"regex check tool call"`/`"regex check tool result"`) use the new `tool_call`/`tool_result` vocabulary. `RegexDetection.tool_output`/`.tool_input` (the regex per-tool pattern config) deliberately stayed on the old `tool_output`/`tool_input` vocabulary to match the existing public `rails.tool_output`/`rails.tool_input` config sections they sit next to in the same YAML file. A broader rename of the public config vocabulary from `tool_output`/`tool_input` to `tool_call`/`tool_result` is a separate, larger effort and out of scope here.
2. **LLM-backed per-tool checks.** This PR wires the regex rail through the new per-tool dispatch path as a proof of the fan-out architecture. Per-tool checks backed by an LLM are a follow-up that reuses the same `_run_per_tool_rail` plumbing.
3. **Load-time validation gaps.** Two related fail-open cases exist today: (a) an invalid/misdirected flow name in `per_tool`'s values isn't validated against the catalog at config-load time and it only surfaces later as a compile failure (IORails); (b) a tool name configured in `per_tool` with no matching pattern group under `regex_detection.tool_output`/`.tool_input` fails open (no check runs), matching `detect_regex_pattern`'s existing behavior for an unconfigured source. Neither is about validating tool names themselves, since those are arbitrary per-request strings with no static enumeration in `RailsConfig`. A `RailsConfig`-level load-time validator covering (a), and a cross-check between `per_tool` and `regex_detection` covering (b), are both follow-ups.
4. **Streaming.** Per-tool checks for both directions already run under streaming, since `RailsManager.are_tool_calls_safe`/`are_tool_results_safe` are the same shared methods `stream_async` calls. But the streaming case hasn't been validated e2e yet and needs more unit test coverage as well for per tool rails.
5. **`/v1/checks` endpoint.** Per-tool checks aren't yet wired into `IORails._do_check`/the checks endpoint (`RailType.TOOL_OUTPUT`/`TOOL_INPUT` don't exist yet in `rails/llm/options.py`).
6. **Other deterministic per-tool checks.** This PR wires only the regex rail through the new per-tool dispatch path, as the smallest slice proving the fan-out architecture. The RFC and its review comments called for deterministic checks generally, not just regex. Other deterministic library rails (e.g. sensitive data detection) could adopt the same `TOOL_CALL`/`TOOL_RESULT` surface + `tool_context_var` pattern established here.

## Related Issue(s)

* NGUARD-908

### Pre-commit

```
$ uv run pre-commit run --all-files
ruff format...............................................................Passed
Insert license in comments................................................Passed
zizmor.....................................................................Passed
ty..........................................................................Passed
```

### Unit-test

```
$ make test
...
7246 passed, 211 skipped in 24.15s
```

### Integration test (real subprocess server + mock tool-calling LLM)

Unlike the rest of the test suite, `tests/guardrails/test_per_tool_regex_rails_server.py` spawns a real `nemoguardrails server` process and a mock tool-calling LLM server over actual HTTP, driving per-tool-call and per-tool-result regex blocking end-to-end (multiple tool calls in one response, cross-tool pattern isolation, block-on-any-match):

```
$ make test TEST="tests/guardrails/test_per_tool_regex_rails_server.py" ARGS="-v"
...
tests/guardrails/test_per_tool_regex_rails_server.py::TestPerToolCallRegexAgainstRealServer::test_matching_pattern_blocked PASSED
tests/guardrails/test_per_tool_regex_rails_server.py::TestPerToolCallRegexAgainstRealServer::test_non_matching_pattern_passes PASSED
tests/guardrails/test_per_tool_regex_rails_server.py::TestPerToolCallRegexAgainstRealServer::test_multiple_tool_calls_all_safe_pass PASSED
tests/guardrails/test_per_tool_regex_rails_server.py::TestPerToolCallRegexAgainstRealServer::test_multiple_tool_calls_second_one_blocks PASSED
tests/guardrails/test_per_tool_regex_rails_server.py::TestPerToolCallRegexAgainstRealServer::test_multiple_tool_calls_first_one_blocks PASSED
tests/guardrails/test_per_tool_regex_rails_server.py::TestPerToolCallRegexAgainstRealServer::test_second_tools_pattern_does_not_cross_match_first_tools_content PASSED
tests/guardrails/test_per_tool_regex_rails_server.py::TestPerToolCallRegexAgainstRealServer::test_second_tool_blocks_on_its_own_pattern PASSED
tests/guardrails/test_per_tool_regex_rails_server.py::TestPerToolResultRegexAgainstRealServer::test_matching_pattern_blocked PASSED
tests/guardrails/test_per_tool_regex_rails_server.py::TestPerToolResultRegexAgainstRealServer::test_non_matching_pattern_passes PASSED

9 passed in 6.35s
```

Integration test files used: https://github.com/NVIDIA-NeMo/Guardrails/commit/8a600e227cc6e8acfe88570150425aa8ddfccd2f

## AI Assistance

- [ ] No AI tools were used.
- [ ] AI tools were used; a human reviewed and can explain every change (tool: ___).

## Checklist

- [ ] This PR links to a triaged issue assigned to me.
- [ ] My PR title follows the project commit convention.
- [ ] I've noted any verification beyond CI and any checks I couldn't run.
- [ ] I did not update generated changelog files manually.

## Summary by CodeRabbit

* **New Features**
  * Added per-tool validation for tool calls and tool results using configurable regular-expression patterns.
  * Added support for tool-specific flows, argument scoping, tool-name tracking, and validation context.
  * Tool-call and tool-result rails are supported through IORails.

* **Documentation**
  * Updated rail support documentation with tool-call and tool-result coverage, limitations, and configuration guidance.

* **Tests**
  * Added comprehensive coverage for blocking, streaming, configuration, pattern matching, and tool-result validation.

### DIFF
--- a/docs/reference/rail-engine-support.mdx
+++ b/docs/reference/rail-engine-support.mdx
@@ -13,9 +13,9 @@ The NVIDIA NeMo Guardrails library ships 33 built-in rails under `nemoguardrails
 Each rail declares its behavior in a [rail manifest](/reference/rail-manifests), and each manifest exposes one or more *surfaces*.
 A surface is the flow name you list under `rails.input.flows`, `rails.output.flows`, or `rails.retrieval.flows` in `config.yml`.
 
-The 33 manifests declare 78 surfaces in total.
-`LLMRails` runs all 78 through its Colang runtime, in both Colang 1.0 and Colang 2.x.
-`IORails` runs 59 of them by compiling the manifest directly, without Colang.
+The 33 manifests declare 80 surfaces in total.
+`LLMRails` runs 78 of them through its Colang runtime, in both Colang 1.0 and Colang 2.x.
+`IORails` runs 61 of them by compiling the manifest directly, without Colang.
 
 This page lists every surface and the engine that can run it.
 For capability areas that are not tied to a specific rail, such as streaming, observability, and the generation API, refer to [Engine Feature Support](/reference/engine-feature-support).
@@ -25,7 +25,7 @@ For capability areas that are not tied to a specific rail, such as streaming, ob
 Legend: ✓ supported · ✗ not supported.
 
 The **Flow** column is the exact string you put in `config.yml`.
-The **Notes** column records either the conversation value the rail rewrites or the reason `IORails` cannot run it.
+The **Notes** column records either the conversation value the rail rewrites or the reason `IORails` or `LLMRails` cannot run it.
 
 A ✓ in the `IORails` column means the surface is *servable*: nothing in its manifest contract puts it out of reach of the engine.
 Whether a particular configuration actually runs it also depends on that configuration, because `IORails` validates models and optional dependencies at startup.
@@ -40,11 +40,15 @@ The following table summarizes surface support by direction:
 | Input | 32 | 32 | 31 |
 | Output | 35 | 35 | 28 |
 | Retrieval | 11 | 11 | 0 |
-| **Total** | **78** | **78** | **59** |
+| Tool_call | 1 | 0 | 1 |
+| Tool_result | 1 | 0 | 1 |
+| **Total** | **80** | **78** | **61** |
 
 `IORails` does not accept a `rails.retrieval` section at all, so every retrieval surface is an `LLMRails` capability.
 A configuration that declares one routes to `LLMRails` by default.
 Refer to [What Happens to an Unsupported Configuration](#what-happens-to-an-unsupported-configuration).
+
+Per-tool checks (the tool-call and tool-result directions) are an `IORails`-only capability by design, so every tool-call and tool-result surface is `IORails` only.
 
 ## Support Matrix
 
@@ -147,6 +151,25 @@ Retrieval surfaces run against the knowledge-base chunks retrieved for a turn.
 | `regex check retrieval` | Regex Detection | ✓ | ✗ | Rewrites `relevant_chunks` |
 | `detect sensitive data on retrieval` | Sensitive Data Detection | ✓ | ✗ | Reads `relevant_chunks` |
 | `mask sensitive data on retrieval` | Sensitive Data Detection | ✓ | ✗ | Rewrites `relevant_chunks` |
+
+### Tool Call Surfaces
+
+Tool call surfaces run against one tool call at a time, configured under `rails.tool_output.per_tool` and keyed by tool name.
+Unlike the [Tool Surfaces](#tool-surfaces) below, these are manifest surfaces like any input, output, or retrieval rail.
+Per-tool checks are an `IORails`-only capability by design:
+
+| Flow | Rail | LLMRails | IORails | Notes |
+|---|---|:---:|:---:|---|
+| `regex check tool call` | Regex Detection | ✗ | ✓ | IORails-only capability by design |
+
+### Tool Result Surfaces
+
+Tool result surfaces run against one tool result at a time, configured under `rails.tool_input.per_tool` and keyed by tool name.
+Per-tool checks are an `IORails`-only capability by design:
+
+| Flow | Rail | LLMRails | IORails | Notes |
+|---|---|:---:|:---:|---|
+| `regex check tool result` | Regex Detection | ✗ | ✓ | IORails-only capability by design |
 
 ### Tool Surfaces
 
--- a/nemoguardrails/guardrails/iorails.py
+++ b/nemoguardrails/guardrails/iorails.py
@@ -741,6 +741,8 @@ class IORails(BaseGuardrails):
             output_parallel=config.rails.output.parallel or False,
             tool_call_flows=config.rails.tool_output.flows,
             tool_result_flows=config.rails.tool_input.flows,
+            per_tool_call_flows=config.rails.tool_output.per_tool,
+            per_tool_result_flows=config.rails.tool_input.per_tool,
             tracer=self._tracer,
             content_capture_enabled=self._content_capture_enabled,
         )
--- a/nemoguardrails/guardrails/rails_manager.py
+++ b/nemoguardrails/guardrails/rails_manager.py
@@ -16,6 +16,7 @@
 """Rails manager for IORails: runs input/output checks through compiled, manifest-driven rails."""
 
 import asyncio
+import json
 import logging
 import warnings
 from collections.abc import Coroutine, Mapping, Sequence
@@ -25,7 +26,7 @@ from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union
 from nemoguardrails.actions.rail_outcome import RailOutcome, TransformTarget
 from nemoguardrails.guardrails.actions.tool_call_action import ToolCallRailAction
 from nemoguardrails.guardrails.actions.tool_result_action import ToolResultRailAction
-from nemoguardrails.guardrails.compiled_rail import CompiledRail, RailDependencies, compile_rail
+from nemoguardrails.guardrails.compiled_rail import CompiledRail, RailDependencies, compile_rail, tool_context_var
 from nemoguardrails.guardrails.engine_registry import EngineRegistry
 from nemoguardrails.guardrails.guardrails_types import (
     RailCallRecord,
@@ -38,11 +39,11 @@ from nemoguardrails.guardrails.guardrails_types import (
 )
 from nemoguardrails.guardrails.telemetry import mark_rail_stop, rail_span, set_rail_content
 from nemoguardrails.guardrails.tool_rail_action import ToolRailAction
-from nemoguardrails.guardrails.tool_schema import ToolExchange, Toolset
+from nemoguardrails.guardrails.tool_schema import ToolExchange, ToolResult, Toolset
 from nemoguardrails.http.runtime import create_http_client
 from nemoguardrails.llm.taskmanager import LLMTaskManager
 from nemoguardrails.manifests import RailDirection as SurfaceDirection
-from nemoguardrails.rails.llm.config import _get_flow_model, _get_flow_name
+from nemoguardrails.rails.llm.config import _get_flow_argument, _get_flow_model, _get_flow_name
 from nemoguardrails.types import ToolCall, UsageInfo
 
 if TYPE_CHECKING:
@@ -150,7 +151,9 @@ def _result_after_rewrites(
     return RailResult(RailOutcome.transform([(_REWRITABLE_TARGET[direction], final_text)]), records=records)
 
 
-def _model_free_record(flow: str, rail_type: str, result: RailResult) -> RailCallRecord:
+def _model_free_record(
+    flow: str, rail_type: str, result: RailResult, tool_name: Optional[str] = None
+) -> RailCallRecord:
     """Build the per-rail GenerationLog record for a rail that reached no model."""
     base_name = _get_flow_name(flow) or flow
     model = _get_flow_model(flow)
@@ -163,6 +166,7 @@ def _model_free_record(flow: str, rail_type: str, result: RailResult) -> RailCal
         action_name=action_name,
         return_value=result.return_value,
         task=f"{action_name} $model={model}" if model else action_name,
+        tool_name=tool_name,
     )
 
 
@@ -188,10 +192,14 @@ def _merge_llm_call_info(record: RailCallRecord, call: "LLMCallInfo") -> RailCal
 
 
 def _rail_call_record(
-    flow: str, rail_type: str, result: RailResult, calls: Sequence["LLMCallInfo"] = ()
+    flow: str,
+    rail_type: str,
+    result: RailResult,
+    calls: Sequence["LLMCallInfo"] = (),
+    tool_name: Optional[str] = None,
 ) -> RailCallRecord:
     """Build the per-rail GenerationLog record from a rail's result and the calls it made."""
-    record = _model_free_record(flow, rail_type, result)
+    record = _model_free_record(flow, rail_type, result, tool_name)
     if not calls:
         return record
     if len(calls) > 1:
@@ -214,6 +222,8 @@ class RailsManager:
         output_parallel: bool = False,
         tool_call_flows: Optional[list[str]] = None,
         tool_result_flows: Optional[list[str]] = None,
+        per_tool_call_flows: Optional[dict[str, list[str]]] = None,
+        per_tool_result_flows: Optional[dict[str, list[str]]] = None,
         tracer: Optional["Tracer"] = None,
         content_capture_enabled: bool = False,
     ) -> None:
@@ -233,6 +243,8 @@ class RailsManager:
 
         self.tool_call_flows: list[str] = list(tool_call_flows or [])
         self.tool_result_flows: list[str] = list(tool_result_flows or [])
+        self.per_tool_call_flows: dict[str, list[str]] = dict(per_tool_call_flows or {})
+        self.per_tool_result_flows: dict[str, list[str]] = dict(per_tool_result_flows or {})
 
         deps = self._rail_dependencies()
         # Keyed by direction as well as flow: compilation is direction-specific, so a surface
@@ -258,9 +270,22 @@ class RailsManager:
         self._tool_call_actions = self._build_tool_actions(self.tool_call_flows, ToolCallRailAction)
         self._tool_result_actions = self._build_tool_actions(self.tool_result_flows, ToolResultRailAction)
 
+        self._per_tool_rails: dict[tuple[SurfaceDirection, str], CompiledRail] = {}
+        per_tool_configured = (
+            (SurfaceDirection.TOOL_CALL, self.per_tool_call_flows),
+            (SurfaceDirection.TOOL_RESULT, self.per_tool_result_flows),
+        )
+        for direction, per_tool in per_tool_configured:
+            unique_flows = {flow for flows in per_tool.values() for flow in flows}
+            for flow in unique_flows:
+                self._per_tool_rails[(direction, flow)] = compile_rail(flow, direction, deps)
+
         self._http_client = create_http_client()
         runtime_deps = replace(deps, http_client=self._http_client)
         self._rails = {key: rail.with_runtime_dependencies(runtime_deps) for key, rail in self._rails.items()}
+        self._per_tool_rails = {
+            key: rail.with_runtime_dependencies(runtime_deps) for key, rail in self._per_tool_rails.items()
+        }
 
         log.info(
             "RailsManager initialized: input_flows=%s, output_flows=%s, tool_call_flows=%s, "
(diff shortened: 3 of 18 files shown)


nemoguardrails/guardrails/rails_manager.py
```diff
@@ -435,6 +506,38 @@ def _enabled_flows(configured: list[str], enabled: Union[bool, list[str]]) -> li
...
 
+    @staticmethod
+    def _resolve_tool_result_name(exchange: ToolExchange, tool_result: ToolResult) -> str:
+        """The tool name a result belongs to: its own name, or its matching prior call's."""
+        if tool_result.name:
+            return tool_result.name
+        matching_calls = [call for call in exchange.calls if call.id == tool_result.call_id]
+        if len(matching_calls) == 1:
+            return matching_calls[0].function.name
+        return ""
```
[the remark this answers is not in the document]
tanushriya910: resolved in the new commit

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -435,6 +506,38 @@ def _enabled_flows(configured: list[str], enabled: Union[bool, list[str]]) -> li
         requested = {_get_flow_name(name) or name for name in enabled}
         return [flow for flow in configured if (_get_flow_name(flow) or flow) in requested]
 
+    @staticmethod
+    def _resolve_tool_result_name(exchange: ToolExchange, tool_result: ToolResult) -> str:
+        """The tool name a result belongs to: its own name, or its matching prior call's."""
+        if tool_result.name:
+            return tool_result.name
```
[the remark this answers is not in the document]
tanushriya910: resolved in the new commit

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -402,7 +448,7 @@ async def are_tool_results_safe(
         not flagged as ambiguous duplicates.
         """
         active = self._enabled_flows(list(self._tool_result_actions), enabled)
-        if not active:
+        if not active and not self.per_tool_result_flows:
```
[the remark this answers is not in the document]
tanushriya910: resolved in the new commit

nemoguardrails/guardrails/iorails.py
```diff
@@ -741,6 +741,8 @@ def __init__(self, config: RailsConfig, *, _report_usage: bool = True) -> None:
             output_parallel=config.rails.output.parallel or False,
             tool_call_flows=config.rails.tool_output.flows,
             tool_result_flows=config.rails.tool_input.flows,
+            per_tool_call_flows=config.rails.tool_output.per_tool,
+            per_tool_result_flows=config.rails.tool_input.per_tool,
```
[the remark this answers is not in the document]
tanushriya910: Looked into this, but I don't think the suggested fix is safe as-is. unsupported_reason's fallback-to-LLMRails mechanism only makes sense when the fallback engine can serve the config in some form. Per-tool checks (per_tool) have no LLMRails equivalent at all, rails.tool_output.per_tool/rails.tool_input.per_tool are IORails-only config fields that LLMRails never reads.

tests/test_runtime_flow_gate_equivalence.py
```diff
@@ -2955,7 +2964,9 @@ def test_every_enabled_surface_has_an_iorails_case():
     enabled = {
         (direction.value, name)
         for (direction, name) in default_rail_catalog().surfaces()
-        if direction is not RailDirection.RETRIEVAL and unservable_reason(name, direction) is None
+        if direction is not RailDirection.RETRIEVAL
+        and direction not in _LLMRAILS_UNSUPPORTED_DIRECTIONS
+        and unservable_reason(name, direction) is None
```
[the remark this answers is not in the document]
tanushriya910: this is resolved in the new commit

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -435,6 +509,42 @@ def _enabled_flows(configured: list[str], enabled: Union[bool, list[str]]) -> li
...
+
+        ``tool_result.name`` is caller-supplied and untrusted for policy selection: a client
+        could claim a different (or no) tool to dodge that tool's configured checks. Only a
+        call id that resolves to exactly one prior call identifies the tool; an unresolved
+        or ambiguous call id resolves to "", never to the supplied name.
+        """
+        matching_calls = [call for call in exchange.calls if call.id == tool_result.call_id]
+        if len(matching_calls) == 1:
+            return matching_calls[0].function.name
+        return ""
```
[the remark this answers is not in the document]
tanushriya910: resolved in the new commit

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -412,8 +461,40 @@ async def are_tool_results_safe(
...
+                per_tool_rails[f"{index}:{flow}"] = self._run_per_tool_rail(
+                    SurfaceDirection.TOOL_RESULT,
+                    flow,
+                    tool_name,
+                    self._build_tool_result_context(tool_name, tool_result),
+                )
+        if not per_tool_rails:
+            return global_result
+
+        per_tool_result = await self._run_tool_rails_sequential(per_tool_rails, RailDirection.INPUT)
```
tgasser-nv, line 503: Running tool-call rails sequentially makes sense for this first PR when we're doing CPU-bound regexes and tool call ID checks. Once we're calling models and become IO-bound we may want to run in parallel

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -435,6 +516,42 @@ def _enabled_flows(configured: list[str], enabled: Union[bool, list[str]]) -> li
...
+        ``$argument=<name>`` on *flow* scopes ``tool_call`` to that one argument, so a
+        check can inspect a user-supplied field without seeing unrelated metadata that
+        could false-positive. Absent, the full arguments dict is serialized, unchanged
+        from before this scoping existed.
+        """
+        arguments = tool_call.function.arguments
+        argument_name = _get_flow_argument(flow)
+        if argument_name:
+            arguments = {argument_name: arguments.get(argument_name)}
+        return {
```
tgasser-nv: nit: Could you create a dataclass for this (maybe ToolCallContext?). The `tool_call` field seems a bit misleading, aren't these the arguments for the current flow?

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -504,6 +621,32 @@ async def _run_tool_result_rail(self, flow: str, exchanges: list[ToolExchange])
                 )
             return result
 
+    async def _run_per_tool_rail(
```
tgasser-nv, line 612: As we discussed in the meeting, would prefer to pass the `ToolCall` dataclass to the tool rail, rather than recreating a context out of the Python contextvar. Having the ToolCall propagate through signatures from are_tool_calls_safe -> _run_per_tool_rail makes it clear what's being passed down

tgasser-nv (commented):
Looks good! Just needs some updates:

1. Could you pass the `ToolCall` class down into the action and declare this in the input signature of the rail? At the moment it comes via a Python contextvar which is then unpacked into a dict with the same shape as `ToolCallFunction`. The per-tool-call rail knows about a single tool type. This tool call can have its schema validated 
2. Can you add a check to the IORails __init__() to check whether the per-tool-call rails in the RailsConfig are found in the Manifest catalog inside `IORails.unsupported_reason` ? If we have a per-tool-call rail that references a non-existent rail this will fail at inference time otherwise.
3. Nit: In some places we're using tool call / result, and in others tool output / input. This is partly my fault as I was reviewing #2312 from Pouyan and get pulled into emergency bugfixes. Recommend keeping tool output/input in all locations for consistency and then doing the rename after the per-tool-call stack is merged.
4. 
Thanks!

nemoguardrails/guardrails/rails_manager.py
tanushriya910: yep, I'll follow-up with parallel run option when I'm adding the LLM-backed checks

nemoguardrails/guardrails/rails_manager.py
tanushriya910: yep, will update this to use the existing ToolCall and ToolResult dataclasses in the next commit

nemoguardrails/guardrails/rails_manager.py
```diff
@@ -412,8 +468,39 @@ async def are_tool_results_safe(
...
+        for index, (exchange, tool_result) in enumerate(result_items):
+            matched_call = self._resolve_tool_call_for_result(exchange, tool_result)
+            if matched_call is None:
+                # Fail closed: reaching this loop at all means active or enabled_per_tool_flows
+                # is non-empty, so some tool-result check is enabled but this result's identity
+                # can't be verified, and there is no way to know which policy should apply.
+                return RailResult.block(
+                    reason="tool result cannot be linked to exactly one prior call; per-tool policy cannot be verified",
+                    records=global_result.records,
+                )
```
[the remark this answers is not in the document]
tanushriya910: fixed in the new commit

nemoguardrails/guardrails/rails_manager.py
tanushriya910: Addressed in the latest commits, ToolCall dataclass is used to pass the tool call to the rail actions and the action extracts the required information from the tool call there. `$argument` is now passed as a surface param to the tool output regex rail so the action can itself do the required filtering/scoping on the tool call arguments.

nemoguardrails/guardrails/iorails.py
```diff
@@ -741,6 +753,8 @@ def __init__(self, config: RailsConfig, *, _report_usage: bool = True) -> None:
...
```
tanushriya910, line 757: using tool_call/tool_result naming here to maintain consistency with global tool flows naming on L754-755

nemoguardrails/guardrails/iorails.py
```diff
@@ -658,6 +658,18 @@ def unsupported_reason(cls, config: RailsConfig, llm: Optional[LLMModel] = None)
...
+        # way rather than let a bad one raise in RailsManager at construction.
+        per_tool_checks = (
+            (config.rails.tool_output.per_tool, SurfaceDirection.TOOL_OUTPUT),
+            (config.rails.tool_input.per_tool, SurfaceDirection.TOOL_INPUT),
+        )
+        for per_tool, direction in per_tool_checks:
+            flows = [flow for tool_flows in per_tool.values() for flow in tool_flows]
+            reason = cls._unservable_rails_reason(flows, direction, deps)
+            if reason is not None:
+                return reason
```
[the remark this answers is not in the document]
tanushriya910: resolved in new commit

nemoguardrails/guardrails/tool_schema.py
```diff
@@ -143,6 +153,24 @@ def _schema_accepts_no_arguments(schema: dict) -> bool:
...
+    ``patternProperties``: presence alone counts, without matching the name against it.
+
+    TODO: match *argument_name* against each ``patternProperties`` regex for a tighter check.
+    """
+    if argument_name in schema.get("properties", {}):
+        return False
+    if schema.get("additionalProperties") is not False:
+        return False
+    if schema.get("patternProperties"):
+        return False
```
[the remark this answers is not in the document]
tanushriya910: was a TODO but ended up implementing as part of this PR itself.

nemoguardrails/guardrails/tool_schema.py
```diff
@@ -177,3 +204,53 @@ def validate_arguments(tool: Tool, arguments: dict) -> str | None:
...
+    """
+
+    @functools.wraps(func)
+    async def wrapper(*args: Any, **kwargs: Any) -> RailOutcome:
+        tool_call: ToolCall = kwargs["tool_call"]
+        tool_definition: Tool | None = kwargs["tool_definition"]
+        name = tool_call.function.name or tool_call.type
+        if tool_definition is None:
+            return RailOutcome.block(reason=f"tool call '{name}' is not an allowed tool")
+        reason = validate_arguments(tool_definition, tool_call.function.arguments)
```
tgasser-nv, line 226: Could you widen this check to also pick up on missing arguments? For example

```
if argument_name is None or argument_name not in arguments:
         return arguments
return {argument_name: arguments[argument_name]}
```

Would catch the case where extra arguments are provided and need to be scanned

docs/reference/rail-engine-support.mdx
```diff
@@ -148,6 +152,25 @@ Retrieval surfaces run against the knowledge-base chunks retrieved for a turn.
...
 
+### Tool Output Surfaces
+
+Tool output surfaces run against one tool call at a time, configured under `rails.tool_output.per_tool` and keyed by tool name.
+Unlike the [Tool Surfaces](#tool-surfaces) below, these are manifest surfaces like any input, output, or retrieval rail.
+Per-tool checks are an `IORails`-only capability by design:
+
+| Flow | Rail | LLMRails | IORails | Notes |
+|---|---|:---:|:---:|---|
+| `regex check tool output` | Regex Detection | ✗ | ✓ | IORails-only capability by design |
```
tgasser-nv, line 163: Add a worked per_tool example here: a config showing rails.config.regex_detection.tool_output.<tool>.patterns alongside rails.tool_output.per_tool.<tool>, and document $argument= option.

tests/rails/llm/test_builtin_rail_manifests.py
```diff
@@ -172,6 +172,24 @@ def _is_action_decorator(decorator: ast.expr) -> bool:
...
+    Otherwise a rail could ship a tool-call check that never validates the call's
+    arguments against the declared tool schema, letting a malformed or undeclared
+    tool call reach the check unvalidated.
+    """
+    surfaces = default_rail_catalog().surfaces(RailDirection.TOOL_OUTPUT)
+    violations = []
+
+    for (_, surface_name), surface in surfaces.items():
+        action = resolve_import_ref(surface.action)
+        if not getattr(action, "_has_tool_output_validation", False):
```
tgasser-nv, line 188: Can you also check `Binding.context("tool_call", "tool_call")` and `Binding.context("tool_definition", "tool_definition"),`?

tgasser-nv (approved):
Thanks just a few small items to fix before merge: 
 
* Could you add a section in the docs explaining that every tool result has to have a `tool_call_id` that matches a previous turn, so previous turns can't be truncated.
* Can you add a doc line to say the `$argument` only applies to tool calls (not results)?
*

docs/reference/rail-engine-support.mdx
tanushriya910: addressed in the new commit under regex.md doc.

nemoguardrails/guardrails/tool_schema.py
tanushriya910: addressed in the new commit

nemoguardrails/guardrails/tool_schema.py
```diff
@@ -177,3 +204,61 @@ def validate_arguments(tool: Tool, arguments: dict) -> str | None:
...
+    to leave a field out. This leans toward more scrutiny, not less: a check that scoped
+    to that name specifically to *avoid* an unrelated field's content may see it anyway on
+    a call where the named field is absent. A config typo or a name the schema could never
+    produce is caught earlier, at compile time, by ``tool_output_validation``.
+
+    TODO: only a single argument name is supported today; add delimiter-separated
+    multi-argument support (e.g. `$argument=a,b`) as a follow-up.
+    """
+    if argument_name is None or argument_name not in arguments:
+        return arguments
```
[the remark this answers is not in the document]
tanushriya910: this was done on purpose to address comments and err on the side of more security, but will follow up on this later to ensure this is fine

tests/rails/llm/test_builtin_rail_manifests.py
tanushriya910: addressed in the new commit

